PR #18, opened 2016-03-28
Better rubocop

cbeer:
Depends on #17

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,11 @@
 inherit_gem:
   dlss_cops: "config/dlss_baseline.yml"
 
+# Turn on RSpec cops
+require: rubocop-rspec
+
 AllCops:
+  DisplayCopNames: true
   TargetRubyVersion: 2.2
   Exclude:
     - 'bin/**/*'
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,102 @@
...
+
+# Turn on RSpec cops
+require: rubocop-rspec
+
+AllCops:
+  DisplayCopNames: true
+  TargetRubyVersion: 2.2
+  Exclude:
+    - 'bin/**/*'
+    - 'db/**/*'
```
mjgiarlo: Do we want to ignore `db` and `config` entirely?

mjgiarlo:
One rubocop failure w/r/t `spec/spec_helper.rb`, so currently :red_circle: 

I am fine with these changes once the build is green. :+1:

.rubocop.yml
cbeer: `db`, I'd argue certainly. Migrations are the only user-maintained thing in there, and most of them are from upstream.

`config`, I could do either way. :shrug:

mjgiarlo: OK, :+1: to leaving `db` and :-1: to `config`, then.

PR #32, opened 2016-03-29
Add apartment gem and toggle between tenant environments based on the current hostname

cbeer:
There's a couple things to iron out sooner or later:

- is Account <-> CNAME 1:1 or 1:n? (assumed 1:1 here)
- for development, test, and non-multitenant environments, it might be nice to skip apartment. I've left the option open in the `before_action require_active_account!`, but didn't do anything with it.
- any sort of authn/authz for the accounts routes.


spec/controllers/accounts_controller_spec.rb
```diff
@@ -68,6 +68,10 @@
 
   describe "POST #create" do
     context "with valid params" do
+      before do
+        expect(Apartment::Tenant).to receive(:create).with('x')
```
mjgiarlo: Can you say more about why this expectation is here?

mjgiarlo: nm!

PR #33, opened 2016-03-29
Remove shared dlss_cops gem by importing applied rules into the application

cbeer:


mjgiarlo:
The rationale here is that HyBox is a Hydra community gem and we want to follow Hydra community styles, not entangle DLSS's concerns with Hydra's. :+1: to that.

PR #34, opened 2016-03-29
Disable multitenancy for non-production environments

cbeer:
This may make it easier to develop and test against the application.

### DIFF
--- a/spec/requests/home_spec.rb
+++ b/spec/requests/home_spec.rb
@@ -6,11 +6,22 @@ RSpec.describe 'Home page', type: :request do
   end
 
   context 'without a current tenant' do
+    before do
+      allow(Settings).to receive(:multitenant).and_return(true)
+    end
+
     describe 'GET /' do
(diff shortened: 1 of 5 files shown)


spec/requests/home_spec.rb
```diff
@@ -6,6 +6,10 @@
   end
 
   context 'without a current tenant' do
+    before do
+      allow(ApplicatioController).to receive(:multitenant?).and_return(true)
```
mjgiarlo: `s/ApplicatioController/ApplicationController/`

Also, you're using `#multitenant?` here and `#multitenant` (sans `?`) above on line 19 of the `ApplicationController`.

cbeer: Thanks. Pretty weird that nothing caught that.

mjgiarlo: Travis did!

mjgiarlo:
:+1:, if Travis is :green_heart:

PR #41, opened 2016-04-06
Merge curation_concerns and sufia helpers into application_helper for easier overriding

cbeer:


mjgiarlo:
@cbeer Where'd the deleted helpers come from? Were they generated by CC and Sufia?

I like the consolidation here. I have only a vague concern that Sufia and CC might (?) care that those helpers are at a certain path, e.g., in an upgrade generator at some point. This is extremely speculative, and I wouldn't hold the PR up over it, but I thought it worth tossing out there.

cbeer:
Yes, I assume they were generated by those upstream gems. We'll have to keep an eye out if an upgrade generator tries to change them (hopefully unlikely, but..).

One alternative would be to update our ApplicationHelper to include both, but removing unnecessary complication seemed like a better path..

cbeer:
Loss of coverage is just from throwing away covered code.

mjgiarlo:
I'm cool with it. I don't believe we've ever modified an application's Sufia helper, so... :shipit:

PR #42, opened 2016-04-06
Use locally generated sufia css

cbeer:


cbeer:
This CSS is from upstream; we haven't done anything to customize hound's CSS behavior, and I'm not inclined to that here.

mjgiarlo:
@cbeer Rebase, pls.

mjgiarlo:
This PR means that if Sufia makes CSS changes in the future, we won't automatically pick that up (because it's not generated into our app)? I believe this change is consistent with the work you've been doing -- just confirming.

Thoughts on this, @jcoyne @atz?

cbeer:
Right; if Sufia makes changes to the upstream `sufia.scss` file that it generates into our app, we won't automatically pick it up. This file just sets up some dependencies, so it should be relatively static.

If Sufia makes changes within `app/assets/sufia/sufia.scss`, we'll always get it as normal.

mjgiarlo:
:ok_hand:

PR #43, opened 2016-04-07
Add application settings for changing the application name

aeschylus:
demo


spec/controllers/sites_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+# is no simpler way to get a handle on the object needed for the example.
+# Message expectations are only used when there is no simpler way to specify
+# that an instance is receiving a specific message.
+
+RSpec.describe SitesController, type: :controller do
+  # This should return the minimal set of attributes required to create a valid
+  # Site. As you add validations to Site, be sure to
+  # adjust the attributes here as well.
+  let(:valid_attributes) do
+    skip("Add a hash of attributes valid for your model")
```
cbeer: Can we fill `valid_attributes` and `invalid_attributes` out? That'll get the scaffolding tests up and running.

app/controllers/application_controller.rb
```diff
@@ -33,4 +34,8 @@ def require_active_account!
     def multitenant?
       Settings.multitenant
     end
+
+    def set_site_i18n_keys
+      I18n.backend.store_translations(:en, sufia: { product_name: Site.instance.application_name })
```
cbeer: This is hideous; hopefully with https://github.com/projecthydra/sufia/pull/1783, we can now override the `application_name` helper.

app/controllers/sites_controller.rb
```diff
@@ -0,0 +1,30 @@
+class SitesController < ApplicationController
+  before_action :set_site
```
cbeer: We should update this controller to use `load_and_authorize_resource` (from cancan) instead of `set_site`.

cbeer:
It looks like this PR needs to be rebased too.

app/controllers/sites_controller.rb
jcoyne: Can you make this a ticket or fix now?

app/models/site.rb
```diff
@@ -0,0 +1,7 @@
+class Site < ActiveRecord::Base
+  validates :application_name, presence: true, allow_nil: true
```
jcoyne: can you explain what this validator does? How is it simultaneously validating presence and allowing nil?

cbeer: My intent was to require users to set an application title, but allow it to be unset initially.

app/controllers/sites_controller.rb
cbeer: Fixed here.

app/models/site.rb
jcoyne: I see, so empty string is disallowed?

cbeer: Exactly. Both cases are incidentally tested in the models and controller tests.

app/controllers/sites_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  load_and_authorize_resource
```
jcoyne: The load will never be run, because set_site will first assign `@site`, is that intended?

cbeer: I figured there was value in consistency; cancancan will short-circuit when it's already loaded, so it isn't harmful.

jcoyne: OK

PR #48, opened 2016-04-08
Add rolify gem to assign users roles for the application

cbeer:


app/models/ability.rb
```diff
@@ -10,12 +10,19 @@ class Ability
   def custom_permissions
     # Limits deleting objects to a the admin user
     #
-    can [:manage], :all if admin?
+    can [:manage], :all if current_user.has_role? :admin
```
mjgiarlo: I'm guessing you want to revert this change since you made an `#admin?` method below.

app/models/user.rb
```diff
@@ -15,10 +16,18 @@ class User < ActiveRecord::Base
...
   # user class to get a user-displayable login/identifier for
   # the account.
   def to_s
     email
   end
+
+  private
+
+    def add_default_roles
+      add_role :admin unless self.class.any?
```
mjgiarlo: What's the effect of `unless self.class.any?` here?

mjgiarlo:
:+1: 

:eyes: Travis

app/models/ability.rb
cbeer: Nope. I hate `Ability#admin?` and intend to remove it upstream.

mjgiarlo: Oh, is that private method below only to satisfy rolify? I saw the duplication in code and balked.

cbeer: No, to satisfy.. sufia? curation_concerms?

mjgiarlo: OK. np.

PR #52, opened 2016-04-11
Add association between a Site and an Account

cbeer:


spec/models/account_spec.rb
```diff
@@ -13,9 +13,18 @@
   describe '#save_and_create_tenant' do
     subject { described_class.new(tenant: 'x', cname: 'x') }
 
+    after(:each) do
+      Apartment::Tenant.drop('x')
+    end
+
     it 'creates a new apartment tenant' do
-      expect(Apartment::Tenant).to receive(:create).with('x')
       subject.save_and_create_tenant
```
mjgiarlo: This test doesn't appear to be testing anything now, except that `subject.save_and_create_tenant` doesn't error out.

mjgiarlo:
Needs rebase.

spec/models/account_spec.rb
cbeer: Good catch. I'll just keep the expectation, but chain `and_call_original`.

PR #75, opened 2016-04-12
[WIP] Add institution name and full name

aeschylus:


spec/features/site_spec.rb
```diff
@@ -24,5 +24,33 @@
...
+        click_on 'Update Site'
+
+        visit sufia.static_path action: 'agreement'
+        expect(page).to have_content('Custom Inst Name (Custom Inst Name) requires')
+      end
+    end
+
+    describe 'institution name full' do
+      before do
+        Site.update(application_name_full: 'fullname')
```
cbeer: Typo/copy-pasta?

config/routes.rb
```diff
@@ -16,8 +16,6 @@
   devise_for :users
   Hydra::BatchEdit.add_routes(self)
 
-  curation_concerns_collections
```
cbeer: Where did these routes go?

spec/controllers/sites_controller_spec.rb
```diff
@@ -27,7 +27,10 @@
   # Site. As you add validations to Site, be sure to
   # adjust the attributes here as well.
   let(:valid_attributes) do
-    { application_name: "Custom Name" }
+    {
+      application_name: "Custom Name",
+      institution_name: "Custom Inst Name"
```
cbeer: It seems like we should also test `institution_name_full` here, yeah?

db/migrate/20160412223500_add_institution_name_full_to_site.rb
```diff
@@ -0,0 +1,5 @@
+class AddInstitutionNameFullToSite < ActiveRecord::Migration
+  def change
+    add_column :sites, :institution_name_full, :string
```
cbeer: Before merging, we'll probably want to collapse these two migrations for our future sanity.

spec/features/site_spec.rb
```diff
@@ -24,5 +24,37 @@
...
+          institution_name_full: 'fullname'
+        )
+      end
+      it 'updates the institution name in the agreement text' do
+        visit edit_site_path
+        fill_in 'Institution name', with: 'Custom Inst Name'
+        click_on 'Update Site'
+
+        visit sufia.static_path action: 'agreement'
+        expect(page).to have_content('fullname (Custom Inst Name) requires')
```
cbeer: Aren't these expectations backwards? I think the agreement text reads `institution_name (instution_name_full)`?

config/routes.rb
aeschylus: Not Sure. They may have gotten lost when merging from master. I can reinstate them.

db/migrate/20160412223500_add_institution_name_full_to_site.rb
aeschylus: Okay. Would I just edit the text of the migration to combine the two `add_column` statements?

spec/features/site_spec.rb
aeschylus: That's what I assumed going in, but the parenthetical seems to be the abbreviation, which also makes sense in its own way. The test also bears that out ("at least on my machine!").

aeschylus: They seem to fail because the form submission doesn't work in the test. Only the `before` update values ever show up.

mjgiarlo:
Merged in #77. Thanks, @aeschylus!

PR #81, opened 2016-04-14
Rescue from loading rubocop when we don't need it

eefahy:


lib/tasks/rubocop.rake
```diff
@@ -1,7 +1,10 @@
...
-RuboCop::RakeTask.new(:rubocop) do |task|
-  task.requires << 'rubocop-rspec'
-  task.fail_on_error = true
+  desc 'Run style checker'
+  RuboCop::RakeTask.new(:rubocop) do |task|
+    task.requires << 'rubocop-rspec'
+    task.fail_on_error = true
+  end
+rescue LoadError
 end
```
jcoyne: It would be good to log a message here. "Unable to find rubocop" or similar.

cbeer: 👎  I'd say, not being there at all is a good enough indication (and otherwise you'll get this message when you run any rake task without the gem available) .

PR #82, opened 2016-04-15
Content blocks should be editable in settings UI

### LINKED ISSUE #74: Make homepage content blocks editable from settings page.

mjgiarlo:
Fixes #74


app/helpers/sufia_helper.rb
```diff
@@ -15,4 +15,22 @@ def institution_name
   def institution_name_full
     Site.institution_name_full || super
   end
+
+  def displayable_content_block(content_block, class_string = '', id_string = '')
+    return if content_block.value.blank?
+    raw("<div id='#{id_string}' class='#{class_string}'>#{content_block.value}</div>")
```
cbeer: Maybe:

```ruby
content_tag :div, content_block.value,  id: id_string, class: class_string
```

cbeer: Instead of passing the class and id, how about we pass `options: {}` and pass it through to the content_tag?

app/helpers/sufia_helper.rb
```diff
@@ -15,4 +15,22 @@ def institution_name
...
+  end
+
+  def my_editable_content_block(content_block)
+    capture do
+      concat form_for([sufia, content_block]) { |f|
```
cbeer: Should we use `do;end;` with this multi-line block?

app/views/sites/_form.html.erb
```diff
@@ -1,30 +1,39 @@
-<%= form_for(@site, url: site_path) do |f| %>
-  <% if @site.errors.any? %>
-    <div id="error_explanation">
-      <h2><%= pluralize(@site.errors.count, "error") %> prohibited this site from being saved:</h2>
+<div class="tab-content">
```
cbeer: I wonder if the tab panel belongs up in `sites/edit` where we're setting up the tabs? I think it'd be less confusing, and let us reuse this partial in non-tab contexts if we wanted.

.rubocop.yml
```diff
@@ -24,6 +24,9 @@ Metrics/LineLength:
   Exclude:
     - 'app/controllers/catalog_controller.rb'
 
+Metrics/AbcSize:
+  Exclude:
+    - 'app/helpers/content_block_helper.rb'
```
cbeer: Instead of excluding this, maybe we can figure out how to simplify those methods?

app/helpers/content_block_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+  include Sufia::ContentBlockHelperBehavior
+
+  def displayable_content_block(content_block, **options)
+    return if content_block.value.blank?
+    content_tag :div, content_block.value, options
+  end
+
+  def editable_content_block(content_block)
+    capture do
+      concat form_for([sufia, content_block]) { |f|
```
cbeer: Maybe we should extract the `form_for` into its own helper? That'd also help deal with the block precedence confusion.

.rubocop.yml
mjgiarlo: It's the `editable_content_block` method in particular, which has a score of 16 (the threshold is 15). Extracting `form_for` into its own method, like you suggested, may do the trick. I'll have a go at this later.

app/helpers/content_block_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+        concat f.label :value, content_block.name.titleize
+        concat f.text_area :value, class: "tinymce", rows: 20, cols: 120
+        concat f.submit 'Save', class: "btn btn-primary"
+      }
+      concat tinymce_assets
```
cbeer: It looks like this causes duplication of the tinymce assets:

screen shot 2016-04-22 at 8 31 18 am

It looks like there's some ways to make it play nice with the asset pipeline that we ought to look at:

https://github.com/spohlenz/tinymce-rails#asset-compilation

app/views/sites/_settings_tab.html.erb
```diff
@@ -0,0 +1,32 @@
+        <div role="tabpanel" class="tab-pane active" id="labels">
```
cbeer: When we did tabbed forms in spotlight, we kept the tab panel markup in `sites/edit` to make the connection between the tablist and the panel is easier to see (with the additional benefit of making this form portable to non-tab contexts, should we have any).

Another thing we did was wrap the whole tab business in the form, so a single submit button will submit the data for all tabs (otherwise you end up in a weird situation where you can only fill out one tab at a time).

app/helpers/content_block_helper.rb
mjgiarlo: OK, my current impl does away with the duplication but does not explore playing nicer with the asset pipeline. Will push what I have soon.

app/helpers/content_block_helper.rb
mjgiarlo: I nuked this entire method.

.rubocop.yml
mjgiarlo: Killed this method.

app/views/sites/_settings_tab.html.erb
mjgiarlo: Done.

mjgiarlo:
I'm pretty happy with this relative to my last push, thanks in large part to @cbeer's prior review. The one bit I'm questioning (:nose:) is: https://github.com/projecthydra-labs/hybox/pull/82/files#diff-36599456e90f51a6a36d6c1a97e53da1R19

app/models/site.rb
```diff
@@ -7,10 +7,48 @@ class Site < ActiveRecord::Base
...
+    delegate :account, :application_name, :institution_name,
+             :institution_name_full, :reload, :update, :announcement_text,
+             :marketing_text, :featured_researcher, to: :instance
 
     def instance
       first || create
     end
   end
+
+  def announcement_text
```
jcoyne: IMHO these methods do not belong on the site model.  You should extract this stuff into a Form object which is responsible for saving the site and any content blocks.

app/models/site.rb
```diff
@@ -7,10 +7,48 @@ class Site < ActiveRecord::Base
...
+    content_block_by_name(ContentBlock::RESEARCHER).value
+  end
+
+  def featured_researcher=(value)
+    save_content_block_named(ContentBlock::RESEARCHER, value)
+  end
+
+  private
+
+    def save_content_block_named(name, value)
```
jcoyne: How are you differentiating these on a per site basis?  It looks to me like there's only a single ContentBlock (per name) shared for all sites. Is that right?

app/models/site.rb
mjgiarlo: I'd like to defer creating a new form object, if at all possible.

app/models/site.rb
cbeer: `apartment` isolates the database for each tenant/site/account.

app/models/site.rb
cbeer: We're getting `ContentBlock` from sufia, I imagine? Instead of adding our own accessors here, would it make sense to set up the expected AR associations to that model?

app/models/site.rb
jcoyne: :ok_hand:

app/models/site.rb
mjgiarlo: @cbeer Correct, that model is provided by Sufia. Do you have something like the following in mind?

```ruby
class Site < ActiveRecord::Base
  # ...
  belongs_to :account
  has_many :content_blocks
  # ...
  class << self
    delegate :account, ..., :content_blocks, to: :instance
    delegate :announcement_text, :marketing_text, :featured_researcher, to: :content_blocks
    # ...
  end
end
```

If so, I'll need to create some convenience methods in Sufia (`#announcement_text`, `#marketing_text`, `#featured_researcher`) first.

PR #90, opened 2016-04-18
Extend ActiveJob::Base to support apartment tenants

### LINKED ISSUE #55: Investigate impact of solr multitenancy on background workers
Apartment has `activejob` support:

https://github.com/influitive/apartment-activejob

We need to extend that to also switch Fedora + Solr connections

cbeer:
Fixes #55 

Using the general pattern from https://github.com/influitive/apartment-activejob, extend ActiveJob::Base to inject apartment and our own multitenancy requirements.

### DIFF
--- a/config/initializers/apartment_activejob.rb
+++ b/config/initializers/apartment_activejob.rb
@@ -1,0 +1,6 @@
+require 'active_job'
+require 'active_job_tenant'
+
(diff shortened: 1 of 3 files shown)


cbeer:
Something this crazy deserves good test coverage.

PR #105, opened 2016-04-21
Create a new solr collection when creating an account

### LINKED ISSUE #59: When setting up a new site, create a new solr collection for the tenant and register it in the Site configuration
Or, if we're feeling clever, we should create a UUID for each `Site`, and use that identifier to connect it to a solr collection.

cbeer:
Fixes #59

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -26,7 +26,7 @@ class AccountsController < ApplicationController
   # POST /accounts.json
   def create
     respond_to do |format|
-      if @account.save_and_create_tenant
+      if CreateAccount.new(@account).save
         format.html { redirect_to @account, notice: 'Account was successfully created.' }
         format.json { render :show, status: :created, location: @account }
       else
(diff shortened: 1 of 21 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -26,7 +26,7 @@ def edit
   # POST /accounts.json
   def create
     respond_to do |format|
-      if @account.save_and_create_tenant
+      if CreateAccount.new(@account).save
```
mjgiarlo: I have a bias towards noun-y service class names, i.e. `AccountCreator` vs. `CreateAccount`, but meh, I wouldn't hold up this PR over that.

app/services/solr_config_uploader.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def zookeeper_path_for_file(file)
+      zookeeper_path(File.basename(file.path))
+    end
+
+    def zookeeper_path(*path)
+      "/#{([collection_path] + path).compact.join('/')}"
+    end
+
+    def salient_files(config_dir)
```
mjgiarlo: What is a salient file?

PR #118, opened 2016-04-28
Add IIIF presentation manifests

jcoyne:
Subtasks:
- [x] Pull in `DisplayImage` from book-concerns
- [x] FileSetPresenter must respond to `display_image` (and return `DisplayImage`)
- [x] Pull in `ImageBuilder` from book-concerns
- [x] Pull in `ResourceBuilder` from book-concerns
- [x] Move `WorkShowPresenter#work_presenters` to curation_concerns (https://github.com/projecthydra-labs/curation_concerns/pull/755)
- [x] Stop checking if the presenter responds to `work_presenters`
- [x] Figure out how to display a Manifest with a single Canvas with a single Image (need different URIs for all those things?)
- [x] Un-skip the skipped specs.
- [x] Pull in `IIIFEndpoint` from book-concerns
- [x] Fully qualified URIs in `@id`s
- [x] Add `@type":"dctypes:Image"`
- [x] correct URLs for service and image.


spec/routing/curation_concerns_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe CurationConcerns::GenericWorksController do
+  it "routes to manifest" do
+    expect(get: '/concern/generic_works/1234/manifest')
+      .to route_to(controller: 'curation_concerns/generic_works', action: 'manifest', id: '1234')
```
jcoyne: This is contrary to rubocop defaults. https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L384 @cbeer what should we do?

cbeer: I wonder if the defaults changed over different rubocop versions. I'd say ignore it, unless you want to reaffirm the default in our own file?

.solr_wrapper
```diff
@@ -2,10 +2,10 @@
 # port: 8983
 collection:
   dir: solr/config/
-  name: hydra-development
+  # name: hydra-development
```
mjgiarlo: Is this commented out because it's already the default value, or what?

config/initializers/riiif.rb
```diff
@@ -1,8 +1,16 @@
 Riiif::Image.file_resolver = Riiif::HTTPFileResolver.new
-Riiif::Image.info_service = lambda do |_id, _file|
-  # resp = get_solr_response_for_doc_id id
-  # doc = resp.first['response']['docs'].first
-  { height: '', width: '' } # doc['height_isi'], width: doc['width_isi'] }
+Riiif::Image.info_service = lambda do |id, _file|
+  # id will look like a path to a pcdm:file
+  # (e.g. rv042t299%2Ffiles%2F6d71677a-4f80-42f1-ae58-ed1063fd79c7)
+  # but we just want the id for the FileSet it's attached to.
+  puts "rawid #{id}"
```
mjgiarlo: Is it still useful to keep the `puts`es in this file?

.solr_wrapper
jcoyne: I can remove this.

config/initializers/riiif.rb
jcoyne: Probably not necessary.

PR #120, opened 2016-04-28
Allow config to be set through environment variables

cbeer:


jcoyne:
Does it make sense to change the `Settings` const to something more descriptive for environment variables?

e.g.: ENV['HyboxSettings.solr.configset_source_path'] instead of ENV['Settings.solr.configset_source_path']?

PR #121, opened 2016-04-28
Ship successful deploy artifacts to S3

cbeer:


.travis.yml
```diff
@@ -18,6 +18,18 @@ jdk:
...
   - bundle exec rake db:create
 
+before_deploy:
+  - mkdir build; zip -r build/hybox.zip .
+
 services:
   - postgresql
   - redis-server
+
+deploy:
```
jcoyne: No need to do `skip_cleanup` anymore? Is this documentation out of date? https://docs.travis-ci.com/user/deployment/s3

.travis.yml
```diff
@@ -18,6 +18,18 @@ jdk:
 before_script:
   - bundle exec rake db:create
 
+before_deploy:
+  - mkdir build; zip -r build/hybox.zip .
```
jcoyne: It looks like this creates too long of a log. Can we pass the `-q` (quiet) flag to `zip`?

PR #131, opened 2016-04-29
Start sidekiq for all the queues

jcoyne:


mjgiarlo:
I don't suppose there's a config file we can lean on for this?

mjgiarlo:
Merging, and I think a follow-on PR might include one of these: https://github.com/mperham/sidekiq/wiki/Advanced-Options#the-sidekiq-configuration-file

PR #139, opened 2016-05-06
Canonicalize the tenant cname

cbeer:


spec/models/account_spec.rb
```diff
@@ -66,4 +66,13 @@
       expect(ActiveFedora.fedora.host).to eq 'http://127.0.0.1:8986/rest'
     end
   end
+
+  describe '#canonicalize_hostname' do
```
jcoyne: Wrong method name

app/models/account.rb
```diff
@@ -36,4 +38,15 @@ def reset!
...
   end
+
+  def default_cname
+    return unless name
+
+    default_host = Settings.multitenancy.default_host
+    default_host % { tenant: name.parameterize }
+  end
+
+  def canonicalize_cname
```
jcoyne: A comment about what is happening and why would be helpful to me. Removing a dot at the end? Should you use \Z instead of $ ?

app/models/account.rb
```diff
@@ -36,4 +38,15 @@ def reset!
     solr_endpoint.reset! if solr_endpoint
     fcrepo_endpoint.reset! if fcrepo_endpoint
   end
+
+  def default_cname
+    return unless name
+
+    default_host = Settings.multitenancy.default_host
+    default_host % { tenant: name.parameterize }
```
mjgiarlo: Oh hey, I forgot String had a `%` method. :+1:

app/models/account.rb
```diff
@@ -36,4 +38,15 @@ def reset!
...
+  end
+
+  def canonicalize_cname
+    self.cname &&= self.cname.sub(/\.$/, '')
```
mjgiarlo: What's the `&&` doing here?

cbeer: `&&=` is (essentially) short-hand for "x = ... unless x.nil?"

mjgiarlo: The inverse of `||=`, yeah. I hadn't seen anyone use this before. Clever.

PR #141, opened 2016-05-10
Activity streams should be multitenant-aware. Fixes #110.

### LINKED ISSUE #110: Activity streams should be multi-tenant aware
The user activity stream shows cross-tenant activity for users that happen to share the same database id.  Possibly look into https://github.com/resque/redis-namespace?

We could also consider refactoring Sufia's activity streams to not use Redis, maybe:

- https://github.com/chaps-io/public_activity
- https://github.com/ledermann/unread

mjgiarlo:


config/initializers/apartment.rb
```diff
@@ -11,8 +11,7 @@
 
   # Add any models that you do not want to be multi-tenanted, but remain in the global (public) namespace.
   # A typical example would be a Customer or Tenant model that stores each Tenant's information.
-  #
-  config.excluded_models = %w{ Account Endpoint SolrEndpoint }
+  config.excluded_models = %w{ Account Endpoint SolrEndpoint RedisEndpoint }
```
cbeer: I'm suspicious that `FcrepoEndpoint` isn't in this list. It's very possible, with STI, we don't need to list out all the subclasses. On the other hand, maybe it's a bug we just haven't noticed yet.

app/models/redis_endpoint.rb
```diff
@@ -0,0 +1,17 @@
+class RedisEndpoint < Endpoint
+  store :options, accessors: [:namespace]
+
+  def self.default_options
+    { namespace: Rails.env }
```
cbeer: What's `Rails.env` in this context?

app/jobs/create_redis_namespace_job.rb
```diff
@@ -0,0 +1,7 @@
+class CreateRedisNamespaceJob < ActiveJob::Base
+  def perform(account)
+    name = account.cname
```
cbeer: You'll probably want `account.tenant` here (or some derivative thereof.. not sure what naming restrictions we have here). The cname is user-changeable, and the tenant is our internal UUID for the account.

app/models/redis_endpoint.rb
```diff
@@ -0,0 +1,19 @@
+class RedisEndpoint < Endpoint
+  store :options, accessors: [:namespace]
+
+  def switch!
+    Sufia.config.redis_namespace = options[:namespace]
+  end
+
+  def reset!
+    redis_instance.keys.each do |namespaced_key|
```
cbeer: What does this method do? Our other endpoints reset the connection to the non-account-namespaced form, but it seems like this might do more?

config/initializers/simple_form.rb
```diff
@@ -117,7 +117,7 @@
   # in this configuration, which is recommended due to some quirks from different browsers.
   # To stop SimpleForm from generating the novalidate option, enabling the HTML5 validations,
   # change this configuration to true.
-  config.browser_validations = false
+  config.browser_validations = true
```
cbeer: Do you want to pull this change out into its own commit, maybe?

app/models/redis_endpoint.rb
```diff
@@ -0,0 +1,18 @@
+class RedisEndpoint < Endpoint
+  store :options, accessors: [:namespace]
+
+  def switch!
+    Sufia.config.redis_namespace = options[:namespace]
+  end
+
+  # Reset the Redis namespace back to the default value
+  def reset!
+    Sufia.config.redis_namespace = 'hybox'
```
cbeer: Can we use the Sufia config setting here, instead of hard-coding it?

PR #147, opened 2016-05-11
Add .ebextension to launch the sidekiq process on the webapp boxes

cbeer:
Hopefully this is only temporary, until we migrate to SQS, worker environments, and shared storage..


cbeer:
Incldues #146

PR #157, opened 2016-05-17
Move the actor into the correct namespace

jcoyne:


jcoyne:
This is already in the update-dependencies pr.

PR #159, opened 2016-05-17
Apply upstream migrations from sufia

cbeer:


jcoyne:
I suspect this also requires you to update the version of sufia in the Gemfile.lock

PR #160, opened 2016-05-17
Move actor to the correct namespace

jcoyne:
Tracking change in curation_concerns 1.0.0.beta2
https://github.com/projecthydra-labs/curation_concerns/commit/502138bf4c55002797b6a6849fb4f4f7b6a2bf66

### DIFF
--- a/app/actors/curation_concerns/actors/generic_work_actor.rb
+++ b/app/actors/curation_concerns/actors/generic_work_actor.rb
@@ -1,6 +1,8 @@
 # Generated via
 #  `rails generate curation_concerns:work GenericWork`
-module CurationConcerns::Actors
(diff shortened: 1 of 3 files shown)


app/actors/curation_concerns/actors/generic_work_actor.rb
```diff
@@ -1,6 +1,6 @@
 # Generated via
 #  `rails generate curation_concerns:work GenericWork`
-module CurationConcerns
+module CurationConcerns::Actors
```
cbeer: Any reason we disabled `Style/ClassAndModuleChildren` instead of addressing the issue?

jcoyne: Because that was code generated by CurationConcerns.

PR #166, opened 2016-05-18
Add Account.use_account!

jcoyne:
This allows us to use a single command instead of having to switch both
the account and the Apartment::Tenant separately

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -37,6 +37,17 @@ class Account < ActiveRecord::Base
     cname
   end
 
+  # Change the current active Account.
+  # Required until https://github.com/influitive/apartment/pull/307 is merged
+  # @param [String] cname the host name for the account
+  # @example
+  #   Account.use_account!('whatever.somewhere.io')
+  def self.use_account!(cname)
+    account = find_by_cname(cname)
+    Apartment::Tenant.switch!(account.tenant)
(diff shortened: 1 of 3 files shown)


app/models/account.rb
```diff
@@ -37,6 +37,17 @@ def self.canonical_cname(cname)
     cname
   end
 
+  # Change the current active Account.
+  # Required until https://github.com/influitive/apartment/pull/307 is merged
+  # @param [String] cname the host name for the account
+  # @example
+  #   Account.use_account!('whatever.somewhere.io')
+  def self.use_account!(cname)
```
cbeer: I'm surprised this is a class method, with responsibility for both finding the account record and switching the connections. If you're happy with it in consuming code, though, :shipit:.

jcoyne: I'd be happy to make an instance method for the switching stuff, but I didn't want to be confusing since there's already a `switch!` method there. What would I call it `switch_all_the_things!`?

mjgiarlo: I prefer the class method approach here. It says, "I don't care what instance you currently have; just use this named account." I agree about the potential for confusion with `switch!` if defined as an instance method.

mjgiarlo:
Merging since @cbeer sounded OK with this change.

PR #175, opened 2016-05-19
Configure Nginx to accept post bodies up to 2gb

jcoyne:


.ebextensions/nginx.config
```diff
@@ -0,0 +1,8 @@
+# From http://stackoverflow.com/a/18951706
+files:
+    "/etc/nginx/conf.d/proxy.conf" :
```
cbeer: maybe rename `proxy.conf` to.. `client_max_body_size.conf`? `upload.conf`?

jcoyne: Okay, I'm not familiar with configuring nginx. I was just copying and pasting as you suggested.

PR #176, opened 2016-05-19
[WIP] Add the client_max_body_size to proxy.conf

jcoyne:


jcoyne:
Truly unnecessary given the nginx.conf on eb

PR #204, opened 2016-05-25
Config ~> 1.2 supports configuring settings via environment variables

cbeer:


Gemfile.lock
```diff
@@ -786,7 +786,7 @@ DEPENDENCIES
   byebug
   capybara
   coffee-rails (~> 4.1.0)
-  config
+  config (~> 1.2)
```
jcoyne: The version of config that was installed hasn't been changed. It's good that we're being more specific with what we want though.

config/settings.yml
```diff
@@ -1,23 +1,23 @@
 multitenancy:
   enabled: false
-  default_host: "<%= ENV.fetch('SETTINGS__MULTITENANCY__DEFAULT_HOST', '%{tenant}.dev') %>"
+  default_host: "%{tenant}.dev"
```
jcoyne: Would it be helpful to write something either as a comment in this file, or in the readme explaining how use env variables to override the settings here?

cbeer: 👌

mjgiarlo: I'd like to see something in the readme about this.

cbeer: I disagree; I'm not sure there's much value in polluting our readme with implementation details about upstream gems (although a sentence saying "there's a bunch of settings in `config/settings.yml` seems reasonable, but I'd argue unrelated to this PR...).

PR #209, opened 2016-05-26
Only redirect to the splash page on the root uri.

jcoyne:
Other uris without account should be 404.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,6 +22,7 @@ class ApplicationController < ActionController::Base
   before_action :set_account_specific_connections!
 
   rescue_from Apartment::TenantNotFound do
+    raise ActionController::RoutingError, 'Not Found' unless base_host?
     redirect_to splash_path
   end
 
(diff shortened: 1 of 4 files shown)


jcoyne:
This will require the AWS repo to set `SETTINGS__MULTITENANCY__HOST = 'demo.hydrainabox.org'`

app/controllers/application_controller.rb
```diff
@@ -22,7 +22,14 @@ class ApplicationController < ActionController::Base
   before_action :set_account_specific_connections!
 
   rescue_from Apartment::TenantNotFound do
-    redirect_to splash_path
+    # If you do a return here you'll get a LocalJumpError
+    # rubocop:disable Style/GuardClause
+    if base_host?
+      redirect_to splash_path if base_host?
```
mjgiarlo: No need for the`if` here.

app/controllers/application_controller.rb
```diff
@@ -22,7 +22,14 @@ class ApplicationController < ActionController::Base
   before_action :set_account_specific_connections!
 
   rescue_from Apartment::TenantNotFound do
-    redirect_to splash_path
+    # If you do a return here you'll get a LocalJumpError
```
mjgiarlo: Instead of a return here, could you do:

```ruby
raise ActionController::RoutingError, 'Not Found' unless base_host?
redirect_to splash_path
```

:question:

mjgiarlo:
Looks good to me. @cbeer ?

PR #230, opened 2016-05-31
Add UniversalViewer

jcoyne:


mjgiarlo:
:red_circle:

cbeer:
Yikes. I wonder if there's value in pulling out the assets into a `universalviewer-rails` gem?

jcoyne:
@cbeer Part of the difficulty of that would be that UV dynamically loads code, so I don't think it can be compiled in the asset pipeline.  This is something that @tpendragon ran into in Plum.

mjgiarlo:
Egads. Do we still need hound in the mix on hybox?

tpendragon:
You'd have to do it in an engine, which would be a lot of breaking up UV. If you can do it though, I'd totally use it!

mjgiarlo:
We discussed this PR in today's HyBox/Sufia standup and decided to merge it so that we can include the UV in our demos in Dublin. This is ripe for extraction to a new engine, and it could possibly be taken up this summer.

PR #234, opened 2016-06-01
Add a flot graph for collections.

jcoyne:


app/views/admin/stats/index.html.erb
```diff
@@ -16,7 +16,15 @@
     <!-- Tab panes -->
     <div class="tab-content">
       <div role="tabpanel" class="tab-pane active" id="collections">
-        derp
```
mjgiarlo: :laughing:

PR #241, opened 2016-06-03
Show graph for works created over time

jcoyne:


app/assets/javascripts/statistics.es6
```diff
@@ -0,0 +1,43 @@
+export class Statistics { 
+  constructor(collection_data, work_data) {
```
cbeer: I wonder if this might be cleaner as, say, `Statistics(data, tab_content)`? I was initially surprised at how it was wired into the view (expecting the reference to the DOM element would be explicit there..).

jcoyne: I'm not sure I understand. Can you say more about that?  What exactly are you thinking `tab_content` is?

cbeer: I was thinking `#work` or `#collection`, but maybe passing the dom reference would be even better?

jcoyne: Statistics is for how the tabs interact with the graph. It is scoped to the whole tab container not just works or collections.

jcoyne:
@cbeer I've renamed the class `StatisticsTabManager`, hopefully that more clearly communicates the purpose.

cbeer:
I think my comment still applies. Maybe I'll try the refactor I have in mind and we can discuss?

mjgiarlo:
@cbeer OK if we merge this PR (assuming :green_apple:) and then consider your refactor later? I'm eager to get this in soon.

jcoyne:
Yes, please do that.

cbeer:
👌 ; hopefully it'll be quick.

mjgiarlo:
Does @cbeer's work supersede this PR or does this one need to be reworked?

jcoyne:
@mjgiarlo: @cbeer's work is merged into this PR.

PR #243, opened 2016-06-03
Whitelist devise and peek controllers so superadmins can log in witho…

cbeer:
…ut being in a tenant context


jcoyne:
Is there a way to test this code?

cbeer:
Not in any useful way I can think of, no.

PR #244, opened 2016-06-03
Enable Google Analytics

jcoyne:


config/initializers/sufia.rb
```diff
@@ -15,10 +15,10 @@
   # Enable displaying usage statistics in the UI
   # Defaults to FALSE
   # Requires a Google Analytics id and OAuth2 keyfile.  See README for more info
-  config.analytics = false
+  config.analytics = true
 
   # Specify a Google Analytics tracking ID to gather usage statistics
-  # config.google_analytics_id = 'UA-99999999-1'
+  config.google_analytics_id = 'UA-71683084-1'
```
cbeer: Do you want to hard-code this, or instead make it an environment variable we can set as part of the stack?

jcoyne: I don't care. Do you?

cbeer: I do. Not everyone running our hybox application wants to report to our GA account.

PR #245, opened 2016-06-03
Adds honeybadger gem. API key to be set in AWS

eefahy:
Normally `config/honeybadger.yml` would have the api key in it but we will be using an environment variable in our AWS stack for that. I kept the config file in place though as we will surely use it to configure the honeybadger service.


cbeer:
Can you use `Settings` to pass the api key in?

PR #253, opened 2016-06-07
Updates to splash page

ggeisler:
These changes are mostly for OR purposes; I think we'll want to do more thorough thinking about what we want on the splash page and how it might be different in the local install versus hosted contexts, which could affect this page quite a bit.

Note that the "Features" text blocks in the lower part of the page will probably be updated when @hannahfrost has a chance to review. It might seem premature, but the idea with this section of the page is to show the general direction of the product and fill out the page a bit more, so it looks more like a real product splash page.

Longer-term, I think we'll want to do more thoughtful organization of the SCSS files and use more variables; I just threw all the new styling in the hybox.scss file for now, knowing it might be temporary.

Here's what the new page looks like:

index_splash____hydra-in-a-box


app/assets/stylesheets/hybox.scss
```diff
@@ -1,26 +1,141 @@
-$brand-primary: #629dd1 !default;
-$jumbotron-heading-font-size: 2.5em;
-$jumbotron-heading-color: #fff;
-$jumbotron-color: #d8d8d8;
 
-.jumbotron {
-  background-image: url("http://hydrainabox.projecthydra.org/images/dark_tint.png"), url("http://hydrainabox.projecthydra.org/images/fractal.jpg");
+#masthead,
+#footer {
```
jcoyne: As of https://github.com/projecthydra/sufia/pull/2166 this selector no longer exists.

app/views/_footer.html.erb
```diff
@@ -0,0 +1,5 @@
+<footer id="footer" class="navbar navbar-inverse">
```
jcoyne: as of https://github.com/projecthydra/sufia/pull/2166 this should be in `app/views/shared/_footer.html.erb`

app/assets/stylesheets/hybox.scss
```diff
@@ -1,26 +1,141 @@
...
+
+  a,
+  p {
+    color: #ccc;
+    font-weight: 500;
+  }
+}
+
+#masthead {
+  #logo {
```
cbeer: This styling makes me think the upstream masthead might be using the wrong markup.

app/assets/stylesheets/hybox.scss
```diff
@@ -1,26 +1,141 @@
...
-.jumbotron {
-  background-image: url("http://hydrainabox.projecthydra.org/images/dark_tint.png"), url("http://hydrainabox.projecthydra.org/images/fractal.jpg");
+#masthead,
+#footer {
+  background-color: #3c3c3c;
+  border: 0;
+
+  a,
+  p {
+    color: #ccc;
```
cbeer: I'd be much more comfortable if we made these global style changes using variables (but I'm fine with punting on splash-page-specific styling for now..)

app/assets/stylesheets/hybox.scss
```diff
@@ -1,26 +1,141 @@
...
+    text-align: center;
+
+    strong {
+      font-weight: 500;
+    }
+  }
+}
+
+.copyright {
+  display: block;
```
cbeer: Is this just `.center-block`?

app/assets/stylesheets/hybox.scss
```diff
@@ -1,26 +1,141 @@
...
     text-transform: uppercase;
   }
+
+  p {
+    color: #444;
+    font-size: 1.6em;
+    margin-bottom: 20px;
+  }
+
+  .btn-sign-up {
```
cbeer: Can we use `button-variant` (from https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/mixins/_buttons.scss#L6) instead of using the compiled styles directly?

ggeisler:
Updates to fix CSS load order issue, which enabled me to refactor the SCSS to use variables, etc.

These updates include moving the `footer` partial to match what was done in projecthydra/sufia#2166, so the footer won't look as intended until HyBox is using the Sufia code updated with projecthydra/sufia#2166.

mjgiarlo:
Could you rebase and squash? I'm seeing a lot of unrelated commits in this PR. Thanks, @ggeisler

mjgiarlo:
Rebased on master in #255.

PR #290, opened 2016-08-18
Update to rails 5.0

jcoyne:


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '4.2.6'
+gem 'rails', '5.0.0'
```
cbeer: 5.0.0.1?

mjgiarlo:
@jcoyne I poked at this for an hour and spent most of the time in `byebug` trying to figure out where the errors in Travis are coming from. The errors are:

* an infinite loop due to `representative_presenter`
* a `NoMethodError` trying to call `#destroy` on `nil`
* a failed validation relating to the depositor not having edit access when saving a collection

I can't imagine any of these are due to Rails 5.

PR #331, opened 2016-09-01
Don't show the admin layout to non-admins. Fixes #330

### LINKED ISSUE #330: Sign up page isn't using sufia layout
https://demo.hydrainabox.org/accounts/new

screen shot 2016-09-01 at 3 50 50 pm

jcoyne:
Fixes #330


app/controllers/accounts_controller.rb
```diff
@@ -78,6 +78,14 @@ def destroy
 
   private
 
+    def decide_layout
+      if can? :read, :admin_dashboard
```
cbeer: Hm, would we ever want to show the `accounts#new` action using the admin template? Or maybe this is where the admin workflow and normal end-user workflow ought to diverge.

mjgiarlo: > Hm, would we ever want to show the `accounts#new` action using the admin template? 

Not that I can think up.

jcoyne: Yes. There's a "Create a new account" link on the Accounts index page (for admins)

jcoyne: See the blue button: 
screen shot 2016-09-01 at 5 06 44 pm

cbeer:
Rebase?

jcoyne:
Rebased.

PR #353, opened 2016-09-14
Move about page editor from the about page to the content block edito…

### LINKED ISSUE #115: Make about page content block editable from settings page
Building on the work done in #74, make the about page content block editable. This could be done in sufia. (remove the in-place editing blocks)

dlpierce:
…rs screen.

Fixes #115

### DIFF
--- a/app/controllers/content_blocks_controller.rb
+++ b/app/controllers/content_blocks_controller.rb
@@ -14,7 +14,7 @@ class ContentBlocksController < ApplicationController
   # PATCH/PUT /sites/1.json
   def update
     respond_to do |format|
-      if @site.update(site_params)
+      if @site.update_content(site_params)
         format.html { redirect_to edit_site_content_blocks_path, notice: 'Site was successfully updated.' }
       else
         format.html { render :edit }
@@ -32,6 +32,7 @@ class ContentBlocksController < ApplicationController
     def site_params
       params.require(:site).permit(:announcement_text,
                                    :marketing_text,
-                                   :featured_researcher)
+                                   :featured_researcher,
+                                   :about_page)
     end
 end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -11,11 +11,27 @@ class Site < ActiveRecord::Base
            :announcement_text=, :marketing_text=, :featured_researcher=,
            to: :content_blocks
 
+  def content_by_name(name)
+    content_blocks.find_by_name(name) || content_blocks.create(name: name)
+  end
+
+  def set_content_by_name(name, value)
+    content_by_name(name).update(value: value)
+  end
+
+  def update_content(params)
+    params.each do |n, v|
+      set_content_by_name(n, v)
+    end
+  end
+
   class << self
     delegate :account, :application_name, :institution_name,
              :institution_name_full, :reload, :update, :announcement_text,
              :marketing_text, :featured_researcher, :announcement_text=,
-             :marketing_text=, :featured_researcher=, to: :instance
+             :marketing_text=, :featured_researcher=, :about_page, :about_page=,
+             :content_by_name, :set_content_by_name,
+             to: :instance
 
     def instance
       first_or_create
--- a/spec/controllers/content_blocks_controller_spec.rb
+++ b/spec/controllers/content_blocks_controller_spec.rb
@@ -27,7 +27,8 @@ RSpec.describe ContentBlocksController, type: :controller do
     {
       announcement_text: 'This is announcement text!',
       marketing_text: 'This is marketing text!',
-      featured_researcher: 'This is featured researcher!'
+      featured_researcher: 'This is featured researcher!',
+      about_page: 'This page is about us!'
     }
   end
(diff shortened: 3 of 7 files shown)


dlpierce:
This depends on projecthydra/sufia#2690

app/controllers/content_blocks_controller.rb
```diff
@@ -32,6 +32,7 @@ def set_site
     def site_params
       params.require(:site).permit(:announcement_text,
                                    :marketing_text,
-                                   :featured_researcher)
+                                   :featured_researcher,
+                                   :about_page)
```
mjgiarlo: Add a spec to `spec/controllers/site_controller_spec.rb` for this change, pls.

app/models/site.rb
```diff
@@ -9,13 +9,15 @@ class Site < ActiveRecord::Base
 
   delegate :announcement_text, :marketing_text, :featured_researcher,
            :announcement_text=, :marketing_text=, :featured_researcher=,
+           :about_page, :about_page=,
```
mjgiarlo: Please add spec coverage for the changes to this file in `spec/features/site_spec.rb`. You can modify the "updates the homepage" feature spec for this.

dlpierce:
I think site_spec.rb is now content_block_spec.rb after the labels and content_block split. The additions to delegate are the methods I  had planned to add to sufia since that is where the other three sets of delegated methods are defined. The assignment methods are what is required to make the content blocks update with a simple @site.update call in the controller.

I could make some content_block utility methods in Site to keep all of this in lerna. Making  sufia's ContentBlock a concern might also help this situation?

mjgiarlo:
@dlpierce Can you rebase this on the latest in master and push again? Looks like there are a whole bunch of extraneous commits in this PR.

app/models/site.rb
```diff
@@ -11,11 +11,27 @@ class Site < ActiveRecord::Base
            :announcement_text=, :marketing_text=, :featured_researcher=,
            to: :content_blocks
 
+  def content_by_name(name)
```
mjgiarlo: @dlpierce Can you write a bit about what the new methods add? We didn't need these for the first few content blocks, is why I ask. Is this a potential or partial refactoring?

dlpierce: @mjgiarlo I'd say it's a partial refactoring. Cloning the existing way of updating ContentBlocks requires a specific method to be defined so it can be passed through the Site via delegate. This is what I originally proposed in a sufia PR.

I could refactor the other content blocks to use this new content_by_name method, if you like?

mjgiarlo: @jcoyne @cbeer ?

cbeer: What does this offer that an ordinary accessor doesn't?

dlpierce: Removes the need to modify the Site and ContentBlock models when adding a new block somewhere on the site. DRYs up the repeated methods in ContentBlock, though the researcher methods have some unique behavior.
If ContentBlock is going to be touched anyway, maybe these new content_by_name methods ought to go there instead.

mjgiarlo: @jcoyne do you have thoughts on the above?

app/models/site.rb
```diff
@@ -11,11 +11,27 @@ class Site < ActiveRecord::Base
...
+    end
+  end
+
   class << self
     delegate :account, :application_name, :institution_name,
              :institution_name_full, :reload, :update, :announcement_text,
              :marketing_text, :featured_researcher, :announcement_text=,
-             :marketing_text=, :featured_researcher=, to: :instance
+             :marketing_text=, :featured_researcher=, :about_page, :about_page=,
+             :content_by_name, :set_content_by_name,
```
jcoyne: Is `set_content_by_name` or `content_by_name` ever called on the class?

app/models/site.rb
```diff
@@ -11,11 +11,27 @@ class Site < ActiveRecord::Base
            :announcement_text=, :marketing_text=, :featured_researcher=,
            to: :content_blocks
 
+  def content_by_name(name)
+    content_blocks.find_by_name(name) || content_blocks.create(name: name)
```
jcoyne: I'd prefere `content_blocks.find_or_create_by(name: name)`

spec/controllers/content_blocks_controller_spec.rb
```diff
@@ -81,12 +83,17 @@
           FactoryGirl.create(:content_block, name: 'featured_researcher')
         end
 
+        let!(:about) do
```
jcoyne: If there's no need to give this object a name, use a `before` block instead.

mjgiarlo: @dlpierce is just repeating the pattern already in this file.

jcoyne: Yes, but it doesn't conform to the new patterns: https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/let_setup.rb

PR #366, opened 2016-09-19
Configure to use https for url generation

jcoyne:


cbeer:
Hm, tricky.  It's true we're using https for our AWS deployment, but it isn't necessarily true everywhere (like, for instance, the docker-based deploy we were talking about last week). Can we put this behind a `config` setting, maybe?

jcoyne:
@cbeer better?

cbeer:
Works for me, thanks.

PR #395, opened 2016-10-04
Move the about page editor to the content block editors screen.

### LINKED ISSUE #115: Make about page content block editable from settings page

dlpierce:
Fixes #115

### DIFF
--- a/app/models/content_block.rb
+++ b/app/models/content_block.rb
@@ -1,0 +1,4 @@
+class ContentBlock < ActiveRecord::Base
+  include Sufia::ContentBlockBehavior
+  include Lerna::ContentBlockBehavior
+end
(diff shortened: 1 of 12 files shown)


app/models/content_block.rb
```diff
@@ -0,0 +1,4 @@
+class ContentBlock < ActiveRecord::Base
+  include Sufia::ContentBlockBehavior
+  include Lerna::ContentBlockBehavior
```
mjgiarlo: I'm on the fence as to whether we need a new module in Lerna for this, but I wouldn't hold up merging this PR because of it.

PR #396, opened 2016-10-04
Fix migration to match what exists in Sufia

jcoyne:
See https://github.com/projecthydra/sufia/blob/master/db/migrate/20160328222236_change_trophy_generic_file_id_to_work_id.rb


db/migrate/20160328222239_change_proxy_deposit_request_generic_file_id_to_work_id.sufia.rb
```diff
@@ -1,5 +1,5 @@
-class ChangeProxyDepositRequestGenericFileIdToGenericWorkId < ActiveRecord::Migration
+class ChangeProxyDepositRequestGenericFileIdToWorkId < ActiveRecord::Migration
   def change
-    rename_column :proxy_deposit_requests, :generic_file_id, :generic_work_id if ProxyDepositRequest.column_names.include?('generic_file_id')
+    rename_column :proxy_deposit_requests, :generic_file_id, :generic_id if ProxyDepositRequest.column_names.include?('generic_file_id')
```
mjgiarlo: `:generic_id`, @jcoyne?

PR #408, opened 2016-10-25
[WIP] Actioncable notifications

jcoyne:


mjgiarlo:
oooooooooooooh

config/cable.yml
```diff
@@ -1,5 +1,7 @@
 development:
-  adapter: async
+  # adapter: async
+  adapter: redis
+  url: redis://localhost:6379/1
```
mjgiarlo: Could we use ENV var jazz in this new file so we're cool in AWSland?

app/channels/notifications_channel.rb
```diff
@@ -0,0 +1,5 @@
+class NotificationsChannel < ApplicationCable::Channel
+  def subscribed
+    stream_from "notifications:1"
```
mjgiarlo: Is the `1` here, and down in the Redis URL below, a Redis "database" to keep these keys cordoned off from other ones?

atz:
@jcoyne still WIP?  should we rebase or kill this?

jcoyne:
Yep, it's still a thing. Needs rebase and tests

mjgiarlo:
I hope this can be superseded by https://github.com/samvera/hyrax/pull/1624

Though we will likely need to do some work in Hyku to make sure the above impl works as expected in a multi-tenant instance.

mjgiarlo:
Closing per above comment.

PR #410, opened 2016-10-25
Fix redis configuration when there are no sentinels configured

cbeer:


jcoyne:
Why is this important? What's the bug?

cbeer:
Without this, you can't start the rails application in development mode.

jcoyne:
Hmm, that hasn't been my experience. I've been running it in development mode all week. What are you seeing?

jcoyne:
I see, this just changed.

PR #411, opened 2016-10-25
Update the admin sidebar to use standard bootstrap components

cbeer:
Not great right now, but at least it's a start.

screen shot 2016-10-25 at 4 04 35 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,12 @@
 module ApplicationHelper
   include ::SufiaHelper
   include ContentBlockHelper
+
+  def nav_link(href)
+    content_tag(:li, class: ('active' if current_page?(href))) do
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,12 @@
 module ApplicationHelper
   include ::SufiaHelper
   include ContentBlockHelper
+
+  def nav_link(href)
```
mjgiarlo: Good addition! Would like to see this extracted into Sufia, but that can happen later.

cbeer: Agreed. I didn't do it in sufia both because it only has 3 items right now and such a change probably should happen down in curation_concerns anyway.

mjgiarlo: :ok:

PR #470, opened 2016-11-10
Revert "Defer registering the apartment callbacks until the database …

jcoyne:
…is initialized"

This reverts commit c87bc16d3eaba5f5214668e8f14584aa388973f9.

This may be causing the current problems connecting to Redis on AWS


mjgiarlo:
Until we have a better solution, we're trying this approach. We can unrevert this in the future if we need it.

PR #523, opened 2016-12-07
Provide iiif manifests for all work types

### LINKED ISSUE #501: My image work should display to me even if user collection is private.
I have an image work (marked public) in a user collection (marked private). 

After navigating to my image work from my dashboard, I would expect to see the image files presented in the image viewer, below the work title and above all the other information, but the space is empty. 
screenshot 2016-11-22 17 24 23

After I change the visibility on the user collection (now marked public), the image viewer displays to me.
screenshot 2016-11-22 17 31 44

jcoyne:
Previously manifests were only created for GenericWorks (not for
Images).

Fixes #501

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -11,7 +11,7 @@ require: rubocop-rspec
 RSpec/SubjectStub:
   Exclude:
     - 'spec/models/solr_endpoint_spec.rb'
-      
+
 # Offense count: 10
 RSpec/LeadingSubject:
   Exclude:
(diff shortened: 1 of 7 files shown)


mjgiarlo:
@jcoyne related: when should we extract our IIIF image/preso impl to Hyrax?

jcoyne:
@mjgiarlo whenever the product owner says that is their highest priority.

mjgiarlo:
@jcoyne good answer. You pass the test!

PR #561, opened 2016-12-19
Prevent using TurboLinks on links to views that use TinyMCE

mjgiarlo:


app/views/content_blocks/_form.html.erb
```diff
@@ -1,16 +1,16 @@
 <div class="panel panel-default tabs">
   <ul class="nav nav-tabs" role="tablist">
     <li class="active">
-      <a href="#about_page" role="tab" data-toggle="tab"><%= t('.tabs.about_page') %></a>
+      <a href="#about_page" role="tab" data-toggle="tab" data-no-turbolink="true"><%= t('.tabs.about_page') %></a>
```
jcoyne: As these are anchor tags, no request is made to the server. Is the turbolinks attribute required in this case?

mjgiarlo: I don't think so, no. Good catch -- changed.

mjgiarlo:
@jcoyne ready for another review.

Issue #563, opened 2016-12-19
Figure out how to load workflows on AWS


mjgiarlo:
It's helpful for demo and QA purposes.

Loading workflows is currently done via rake task, not in the UI.

jcoyne:
Should this go into the codepipeline templates?

jcoyne:
Actually, we probably need it in the "create a new tenant" process

PR #566, opened 2016-12-20
Load workflows when rendering workflow roles UI.

mjgiarlo:
Refs #563

This is a temporary hack to automatically load workflows in the demo instance. It also:

* Overrides the WorkflowRolesController to specify the admin layout (workaround for projecthydra-labs/hyrax#67)
* Adds breadcrumbs to the WorkflowRolesController override (should also be done in Hyrax)
* Pulls in the workflow section to the admin sidebar from Hyrax (and tweaked to use the Hyku MenuPresenter)
* Adds I18n for the workflow roles header
* Updates the hyku mediated deposit json to pull in the latest from Hyrax

### DIFF
--- a/app/presenters/hyku/menu_presenter.rb
+++ b/app/presenters/hyku/menu_presenter.rb
@@ -17,6 +17,12 @@ module Hyku
     end
 
     # Returns true if the current controller happens to be one of the controllers that deals
+    # with workflow.  This is used to keep the parent section on the sidebar open.
+    def workflows_section?
+      %w(admin workflow_roles).include?(controller_name)
+    end
+
+    # Returns true if the current controller happens to be one of the controllers that deals
     # with roles and permissions.  This is used to keep the parent section on the sidebar open.
     def roles_and_permissions_section?
       # we're using a case here because we need to differentiate UsersControllers
--- a/app/views/hyrax/admin/_sidebar.html.erb
+++ b/app/views/hyrax/admin/_sidebar.html.erb
@@ -67,6 +67,23 @@
       <span class="fa fa-sitemap"></span> <%= t('hyrax.admin.sidebar.admin_sets') %>
     <% end %>
 
+    <li>
+      <a role="button" class="collapsed collapse-toggle" data-toggle="collapse" href="#collapseWorkflows" aria-expanded="false" aria-controls="collapseWorkflows">
+        <span class="fa fa-code-fork"></span>
+        <span><%= t('hyrax.admin.sidebar.workflow') %></span>
+      </a>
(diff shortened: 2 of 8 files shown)


app/controllers/hyrax/admin/workflow_roles_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Hyrax
+  module Admin
+    class WorkflowRolesController < ApplicationController
```
jcoyne: You should add a comment here about why this is in Hyku and not Hyrax

app/controllers/hyrax/admin/workflow_roles_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      before_action :require_permissions
+      layout 'admin'
+
+      def index
+        # TODO: Find a better way to make sure workflows are loaded
```
jcoyne: Do we need to load them per tenant? Can we do it when a tenant is created as we do with admin accounts?

app/controllers/hyrax/admin/workflow_roles_controller.rb
jcoyne: Could you add to an initializer `WorkflowRolesController.before_action :index do ... end`?

app/controllers/hyrax/admin/workflow_roles_controller.rb
mjgiarlo: Maybe, but this PR is about enabling workflows in our current tenants. I have not written the words that will magically close the issue so we should revisit in the new year with a better solution.

app/controllers/hyrax/admin/workflow_roles_controller.rb
mjgiarlo: I could. Do you think that would be better for this temporary hack?

jcoyne: Yes, i think that will make it a lot less permanent.

mjgiarlo: Pushed. Let's see what Travis thinks.

jcoyne: This is better, but I was suggesting getting rid of this controller and adding the additions in an initializer.  That results in less copy/paste code that we have to hunt down later.

mjgiarlo:
@jcoyne Have all the decomposed bits of this been submitted and merged now? It seems close if not done.

jcoyne:
@mjgiarlo I think I have all the bits now.

PR #577, opened 2017-01-04
Remove duplicate "Featured Researcher" title

laritakr:
Removed "Featured Researcher" heading because the title is already either on the content
section or on the tab from the prior partial (in hyrax). 

Also changed to access the text from the hyrax yml file to enable appropriate language function for this content block.

This resolves issue https://github.com/projecthydra-labs/hyku/issues/555


mjgiarlo:
@laritakr Should our override also include the markup from Hyrax's partial so that our override is more closely aligned with what we're getting from upstream? See what's in Hyrax: https://github.com/projecthydra-labs/hyrax/blob/master/app/views/hyrax/homepage/_featured_researcher.html.erb

Cc: @jcoyne @jeremyf @ggeisler

jcoyne:
@mjgiarlo no, this should not align with upstream. This was how @ggeisler did the mockup. We changed the local partial, and left the upstream as it was before.

PR #578, opened 2017-01-05
Automatically load workflows when switching accounts

### LINKED ISSUE #563: Figure out how to load workflows on AWS

jcoyne:
Fixes #563

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -46,10 +46,18 @@ class Account < ActiveRecord::Base
     cname
   end
 
+  # Make all the account specific connections active
   def switch!
     solr_endpoint.switch! if solr_endpoint
     fcrepo_endpoint.switch! if fcrepo_endpoint
     redis_endpoint.switch! if redis_endpoint
+    load_workflows
+  end
+
+  # Ensure that some workflows have been loaded into the database
+  def load_workflows
+    return if Sipity::Workflow.any?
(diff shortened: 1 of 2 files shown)


app/models/account.rb
```diff
@@ -46,10 +46,18 @@ def self.canonical_cname(cname)
...
   def switch!
     solr_endpoint.switch! if solr_endpoint
     fcrepo_endpoint.switch! if fcrepo_endpoint
     redis_endpoint.switch! if redis_endpoint
+    load_workflows
+  end
+
+  # Ensure that some workflows have been loaded into the database
+  def load_workflows
+    return if Sipity::Workflow.any?
```
mjgiarlo: @jcoyne Why do we need this check? If I have, say, 2 workflows available and I run the importer *n* times, I will still wind up with 2 `Sipity::Workflow`s. Performance related, I guess?

mjgiarlo:
@jcoyne Left one question. Otherwise, happy to merge.

app/models/account.rb
jcoyne: Yes. The importer reimports the workflows when called, so it's very slow.

mjgiarlo: :ok_hand:

PR #607, opened 2017-01-12
Update dry-validation to avoid deprecation errors

jcoyne:


config/workflows/mediated_deposit_workflow.json
```diff
@@ -40,7 +40,7 @@
                     "notifications": [
                         {
                             "notification_type": "email",
-                            "name": "Hyrax::Workflow::CompleteNotification",
+                            "name": "Hyrax::Workflow::DepositedNotification",
```
jeremyf: Why was this changed as part of the update for the dependency?

jcoyne: The hyrax version got updated, and this change was made in hyrax.

PR #617, opened 2017-01-16
Updating rubocop

jeremyf:


.rubocop.yml
```diff
@@ -2,6 +2,11 @@
 require: rubocop-rspec
 inherit_from: .rubocop_todo.yml
 
+# With the rubocop 0.47.0 and rubocop-rspec 1.8.0 the following stderr message was sent:
+#   An error occurred while RSpec/DescribedClass cop was inspecting path/to/file
```
jcoyne: Is this a bug in rubocop?  Is there a ticket open for it?

.rubocop_todo.yml
```diff
@@ -1,18 +1,32 @@
...
+# Include: **/Gemfile, **/gems.rb
+Bundler/OrderedGems:
   Exclude:
-    - 'spec/models/solr_endpoint_spec.rb'
+    - 'Gemfile'
+
+# Offense count: 56
+# Configuration parameters: CountComments, ExcludedMethods.
+Metrics/BlockLength:
+  Max: 220
```
jcoyne: This seems really big, possibly it's an outlier (config block maybe?) Do we want to exclude just a few files and reduce the max?

Gemfile
```diff
@@ -46,8 +46,8 @@ group :development, :test do
   gem 'solr_wrapper', '~> 0.10'
   gem 'fcrepo_wrapper', '~> 0.4'
 
-  gem 'rubocop', '~> 0.43.0'
-  gem 'rubocop-rspec', '~> 1.8.0'
+  gem 'rubocop', '~> 0.43'
```
jcoyne: This seems like it'll break next time rubocop is updated.  I'd prefer we handle the updates on our terms.

.rubocop.yml
jeremyf: I think the issue is that 0.47.0 was released and 1.10.0 of rubocop-rspec was released. So we were running into conflicts in API changes (or some such thing).

PR #646, opened 2017-01-19
Set lucene to match version of solr

### LINKED ISSUE #636: Update TokenFilterFactory
There are lots of messages like this in the solr logs on AWS:
```
o.a.s.s.FieldTypePluginLoader TokenFilterFactory is using deprecated 5.0.0 emulation. You should at some point declare and reindex to at least 6.0, because 5.x emulation is deprecated and will be removed in 7.0
```

Can we stop emulating 5.0 so they go away?

eefahy:
closes #636

### DIFF
--- a/solr/config/solrconfig.xml
+++ b/solr/config/solrconfig.xml
@@ -28,7 +28,7 @@
        that you fully re-index after changing this setting as it can
        affect both how text is indexed and queried.
   -->
-  <luceneMatchVersion>5.0.0</luceneMatchVersion>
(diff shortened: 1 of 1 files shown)


atz:
@eefahy any concern about how to roll out this change, since it requires bouncing solr?

eefahy:
@atz well, at the moment there are zero collections or indices in solr so it wouldn't need a restart.

cbeer:
Yes --  it is generally fine to bump that version, even with data in the index. There may be some edge cases if you care a lot about search relevancy, but :shrug:.

PR #656, opened 2017-01-24
Ignore postgres artifacts

atz:
These are generated when developing according to the steps in README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ capybara-*.html
 /tmp
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/db/postgres
+pg_upgrade_*.log
 /public/system
 /coverage/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,6 +5,8 @@ capybara-*.html
 /tmp
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/db/postgres
```
jcoyne: We don't have a dependency on postgres. How did these get in you code?

jcoyne: Derp, I was thinking hyrax.

jcoyne: How are you getting these files here though? aren't they in you system postgres directory?

atz: Follow the instructions in the readme and they end up there.

PR #714, opened 2017-01-27
Separate RemoveSolrCollectionJob from CleanupAccountJob

### LINKED ISSUE #711: CleanupAccountJob is not idempotent
The first time it runs it can successfully remove the Solr endpoint and fail to remove the fedora endpoint.  Then if the job is rerun, it will fail when it can't find the solr endpoint (that it removed on the previous run).

jcoyne:
This improves the idempotency of CleanupAccountJob

Fixes #711

### DIFF
--- a/app/jobs/remove_solr_collection_job.rb
+++ b/app/jobs/remove_solr_collection_job.rb
@@ -1,0 +1,15 @@
+class RemoveSolrCollectionJob < ActiveJob::Base
+  # @param collection [String] the name of the collection
+  # @param connection_options [Hash] options for connecting to solr.
+  # @option connection_options [String] :url
+  # @option connection_options [String] :url
+  def perform(collection, connection_options)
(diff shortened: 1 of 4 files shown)


app/jobs/remove_solr_collection_job.rb
```diff
@@ -0,0 +1,15 @@
...
+  # @option connection_options [String] :url
+  # @option connection_options [String] :url
+  def perform(collection, connection_options)
+    solr_client(connection_options).get '/solr/admin/collections', params: { action: 'DELETE', name: collection }
+  end
+
+  private
+
+    def solr_client(connection_options)
+      RSolr.connect(connection_options)
```
mjgiarlo: We used to use `Blacklight.default_index.connection` to get the connection information. Can you say why we are moving away from that in this PR?

jcoyne: The Blacklight.default_index is mutable (changed by Account#switch). Since we're moving the solr call to a time after the account record has been deleted, we need to make it immutable.

Issue #705, opened 2017-01-27
Document tenancy architecture


atz:
You know, beyond the one sentence saying "we use apartment with postgres".  

The audience is developers and maintainers.  

Document should answer questions like:
- when does tenant switching get invoked? what is impacted?
- where in the DB does a given user account get stored?
- when do new users automatically get special roles/permissions?
- how do I test tenant-segregation in the interface? in spec?
- how does tenancy relate to DNS/hostname? how do I test that?
- how do I "see" a given tenant environment in `rails console`?

Diagram or flow chart might help also.

atz:
Also, how do we run `db:migrate` against all tenant DB schemas?

jcoyne:
This page? https://github.com/projecthydra-labs/hyku/wiki

mjgiarlo:
@atz 
> how do I "see" a given tenant environment in rails console?
I believe that's the `AccountElevator.switch!` jazz in the README.

atz:
@jcoyne: wiki location is fine, but you're not claiming all those points are covered in that doc.

mjgiarlo:
Certainly not all those points. We need better documentation in this area.

atz:
Related clarifications are needed around the expectations for our various levels of user permissions:
- Anonymous user
- Logged in user
- User with admin role
- User with superadmin role
- Edge case: User w/ superadmin but *not* admin role.  Is anything expected to be restricted?

Separately, we need to define what we expect to expose in the demo site, i.e. who is the audience?  Should everybody be able to create accounts and make themselves superadmins?  etc.

atz:
See related DRAFT document to be finalized: https://docs.google.com/spreadsheets/d/1TSZ8CUuq6puwzyJ1Av0aQ798FOQZgzhkn2jaDYyt118/edit?pli=1#gid=0

mjgiarlo:
@atz How close does the newly revised https://github.com/projecthydra-labs/hyku/wiki#multi-tenant-domain-model (HT to @tdonohue!) get us to closing this ticket?

atz:
Yeah, closing this now.

Issue #649, opened 2017-01-20
Can't create a new repo if my first naming attempt was unsuccessful due to conflict


hannahfrost:
When I am trying to create a new repo tenant, if my first title/short name combination was not acceptable due to a conflict (one or the other already in use), no matter what I enter in either of the two fields on the second attempt causes an error.

mjgiarlo:
Blocked by #647

mjgiarlo:
@atz should #727 have closed this issue?

mjgiarlo:
@atz :point_up: bump!

atz:
@mjgiarlo: Not sure.  Right now the demo site doesn't behave like I'd expect.  The error no longer occurs at repo creation time for superadmin, even during collision.  POST succeeds, forwards to user creation.  In that sense, we might close this.

However, no message informs of collision.  Human friendly name may be overwritten(?) and subsequent users co-mingled unintentionally.  Meanwhile in my local console, if I attempt the same, I get an appropriate error:
```ruby
2.3.1 :002 > x=Account.new(title: "My Title", name: "xname")
 => #<Account id: nil, tenant: nil, cname: nil, created_at: nil, updated_at: nil, solr_endpoint_id: nil, fcrepo_endpoint_id: nil, name: "xname", redis_endpoint_id: nil, title: "My Title"> 
2.3.1 :003 > x.save
   (0.2ms)  BEGIN
  Account Exists (6.0ms)  SELECT  1 AS one FROM "public"."accounts" WHERE "public"."accounts"."tenant" = $1 LIMIT $2  [["tenant", "1e2e65f3-b6f8-4b96-8b91-7840d66f9db7"], ["LIMIT", 1]]
  Account Exists (0.3ms)  SELECT  1 AS one FROM "public"."accounts" WHERE "public"."accounts"."cname" = $1 LIMIT $2  [["cname", "xname.dev"], ["LIMIT", 1]]
  SQL (0.9ms)  INSERT INTO "public"."accounts" ("tenant", "cname", "created_at", "updated_at", "name", "title") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"  [["tenant", "1e2e65f3-b6f8-4b96-8b91-7840d66f9db7"], ["cname", "xname.dev"], ["created_at", 2017-04-10 22:08:15 UTC], ["updated_at", 2017-04-10 22:08:15 UTC], ["name", "xname"], ["title", "My Title"]]
   (0.7ms)  COMMIT
[...]

Issue #648, opened 2017-01-20
resolving a repository name conflict


hannahfrost:
While creating a new repository tenant, if the title or short name that I enter has already been used, in addition to getting a message about the conflict, the form changes. The title field in the form is cleared, but not the short name.  Based on testing, this seems to be case regardless if the title is the conflict or the short name is the conflict.

In either case, it would be would be better to clear out both fields for the user to enter a new title and short name combination.

mjgiarlo:
Blocked by #647

mjgiarlo:
@atz Still in progress, or should we shunt it onto the backlog?

atz:
@mjgiarlo Should just merge this with the other repo creation ticket (#649).  I cannot seem to cause collision in the demo site, oddly.

PR #727, opened 2017-01-31
Accounts forms

### LINKED ISSUE #655: app/views/accounts/_edit_form.html.erb is unneeded or untested
This seems to have been replaced by `edit.html.erb`, corresponding to lines 8-58.  At best the partial is untested (I delete it and all tests pass).  But I also can't find anywhere it is invoked, referenced or rendered.  Could be cruft.  Can anyone vouch for its usefulness or should we just get rid of it?

atz:
This does not entirely resolve all the accounts creation tickets, but it does fix several important pieces.  

While waiting for clarity on #705, we can merge this.

See #725 
See #649 
See #648 
See #647 

Fixes #655


app/models/account.rb
```diff
@@ -1,9 +1,19 @@
 # Customer organization account
 class Account < ActiveRecord::Base
+  # @param [String] piece the tenant piece of the canoncial name
```
mjgiarlo: Typo: `s/canoncial/canonical/`

app/models/account.rb
```diff
@@ -1,9 +1,19 @@
...
+  def self.default_cname(piece)
+    return unless piece
+    default_host = Settings.multitenancy.default_host
+    format(default_host, tenant: piece.parameterize)
+  end
+
   attr_readonly :tenant
-  validates :name, presence: true
+  # name is unused after create, only used by sign_up/new forms
+  validates :name, presence: true, unless: 'cname.present? && cname != default_cname("")'
```
mjgiarlo: Since we have multiple calls to `default_cname` with an empty string passed in, I wonder if we should handle this in that method's definition:

```ruby
def self.default_cname(piece = '')
```

And then call `default_cname` without args in these places?
I'm not sure what it means to call that method with an empty string.

app/views/account_sign_up/new.html.erb
```diff
@@ -8,10 +8,9 @@
         <% if @account.errors.any? %>
           <div id="error_explanation" class="alert alert-danger">
             <%= pluralize(@account.errors.count, "error") %> prohibited this repository from being saved:
-
             <ul>
-            <% @account.errors.full_messages.each do |message| %>
-              <li><%= message %></li>
+            <% @account.errors.messages.each do |key, messages| %>
```
mjgiarlo: Can you say more about this change?

app/views/accounts/_edit_form.html.erb
```diff
@@ -1,49 +0,0 @@
-<%= form_for(@account, class: 'form') do |f| %>
```
mjgiarlo: Was this file dead code? Are we providing users the ability to tweak these endpoints elsewhere?

app/views/account_sign_up/new.html.erb
atz: The underlying ticket complained about identifying the field(s) in question when the response includes validation errors.  This is because the same model validation messages are triggered from different forms/controllers, but the forms do not expose the same fields.  

This step uses greater detail available from `errors` to include the rejected value in the message.  Exposing the `key` would be prerequisite to (for example) mapping the error message for one field (e.g. `cname`) back to a message referring to `Short Name`, where only the latter is exposed on the form.

app/models/account.rb
atz: Yes, that is acceptable.  One of the fields (`cname`) obtains a default value based on another field (`name`) passed through a configurable template string.  Presently, it may have already obtained that value, but erroneously (`name` was empty).  Calling `default_cname('')`, validators know what to exclude as effectively a `nil` value.

app/views/accounts/_edit_form.html.erb
atz: Yes, there is an `edit` form.  The partial is unused.  It was already deleted once but seems to have crept back in via rebase.  See #655.

app/models/account.rb
atz: However, changing the default does introduce some weirdness since the previous instance-level `default_cname` param takes a different default.  I find the explicit args help make it clear what the expected result is since the logic is the same.  We're talking about 2 calls, so I don't think this warrants much attention.

atz:
Fixed the typo.

app/models/account.rb
mjgiarlo: :shipit:

app/views/accounts/_edit_form.html.erb
mjgiarlo: Ah, it was never deleted, but it stopped being used a while back. Cool.

mjgiarlo:
@atz Approved this PR and fixed a merge conflict. Feel free to merge once 💚 . Thanks!

PR #773, opened 2017-02-13
Adds config for log capture via logz.io

bbranan:
Installs Filebeat (https://www.elastic.co/products/beats/filebeat) via Elastic Beanstalk extentions for shipping logs to logz.io.

### DIFF
--- a/.ebextensions/logzio.config
+++ b/.ebextensions/logzio.config
@@ -1,9 +1,28 @@
+packages:
+  yum:
+    gettext: []
+
 files:
+  "/etc/init/filebeat.conf":
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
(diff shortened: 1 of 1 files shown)


atz:
I have no way to evaluate this technically, but looks OK to me.

bbranan:
Upon further testing, it looks like I will need to make a few tweaks to allow all defined logs to be shipped properly. Let's hold off on merging this one until I'm able to make those changes.

bbranan:
@eefahy this is ready to merge

jcoyne:
Should we install https://github.com/dwbutler/logstash-logger ?

bbranan:
@jcoyne We could, though we'd still need filebeat to capture the other logs (nginx, beanstalk, aws, etc), so we'd end up with log capture config in two places rather than one.

PR #778, opened 2017-02-14
Require rails_helper in one place instead of 100

atz:


jcoyne:
Prior to this change I could do `$ ruby -Ispec  spec/actors/hyrax/actors/generic_work_actor_spec.rb` and now I can't.

atz:
@jcoyne: do `rspec spec/actors/hyrax/actors/generic_work_actor_spec.rb`.  I believe that is the conventional invocation.

mjgiarlo:
Agree that I'm not sure we care too much about that invocation, which I've not seen in any of our/Hydra docs. Inclined to merge.

PR #797, opened 2017-02-16
Split AccountsController into tenant and superadmin controllers

cbeer:
I'm looking for early feedback, particularly a better name for this than `multitenant_admin`.


mjgiarlo:
We already have tenants living in apartments; we could double down on this metaphor and call the multitenant stuff "superintendent" or "landlord."

cbeer:
Thanks, good idea. I've gone with "Proprietor"

PR #803, opened 2017-02-17
Add proprietor route for the dashboard pending upstream bug fix

cbeer:
Work-around pending https://github.com/projecthydra-labs/hyrax/issues/376

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,10 @@ Rails.application.routes.draw do
       get '/account/sign_up' => 'account_sign_up#new', as: 'new_sign_up'
       post '/account/sign_up' => 'account_sign_up#create'
       get '/', to: 'splash#index'
+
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -5,6 +5,10 @@
       get '/account/sign_up' => 'account_sign_up#new', as: 'new_sign_up'
       post '/account/sign_up' => 'account_sign_up#create'
       get '/', to: 'splash#index'
+
+      # pending https://github.com/projecthydra-labs/hyrax/issues/376
+      get '/dashboard', to: redirect('/')
```
jcoyne: what if you actually want to go to the dashboard?  won't this prevent that from happening?

cbeer: There is no dashboard for the proprietor app.

PR #809, opened 2017-02-18
Canonicalize the admin host name setting

cbeer:


app/models/account.rb
```diff
@@ -9,6 +9,13 @@ def self.default_cname(piece)
     format(default_host, tenant: piece.parameterize)
   end
 
+  def self.admin_host
```
cbeer: This isn't a great place for this method, but, short of introducing something brand new, seemed like an acceptable place.

PR #824, opened 2017-02-22
Use the default job adapter in development

cbeer:


cbeer:
Yes, I have no problems with async.

PR #884, opened 2017-03-23
Create a default admin set when a tenant is created

mjgiarlo:
After a long discussion with @atz, @cbeer, and @jcoyne, I have decided to start by adding a new job to create a default Admin Set. A follow-up PR might be to sacrifice idempotency to handle account creation in a single job. This would be temporary, to ensure that default AS creation works during the community beta testing period (May 15th through June). Ultimately we should pursue the "UX step" approach outlined by @atz in #884, though we likely don't have time to complete and test that for a month or two.

### DIFF
--- a/app/jobs/create_default_admin_set_job.rb
+++ b/app/jobs/create_default_admin_set_job.rb
@@ -1,0 +1,5 @@
+class CreateDefaultAdminSetJob < ActiveJob::Base
+  def perform
+    AdminSet.find_or_create_default_admin_set_id
+  end
+end
--- a/spec/jobs/create_default_admin_set_job_spec.rb
+++ b/spec/jobs/create_default_admin_set_job_spec.rb
@@ -1,0 +1,8 @@
+RSpec.describe CreateDefaultAdminSetJob do
+  describe '#perform' do
+    it 'creates a new admin set for an account' do
+      expect(AdminSet).to receive(:find_or_create_default_admin_set_id).once
+      described_class.perform_now
+    end
+  end
+end


mjgiarlo:
Cc: @eefahy @bbranan

app/services/create_account.rb
```diff
@@ -41,6 +42,10 @@ def create_redis_namespace
     CreateRedisNamespaceJob.perform_later(account)
   end
 
+  def create_default_admin_set
+    CreateDefaultAdminSetJob.perform_later(account)
```
jcoyne: This has to happen after the Fcrepo & Solr collection is created. Presently we don't have any way of making that assurance.  Perhaps we need an "InitializeClientJob".

mjgiarlo: Doesn't this happen up above in `#create_external_resources` where this call happens the Fcrepo and Solr collections are successfully created?

atz: @jcoyne Are you thinking this job would do the various `create` steps as `.perform_now`s?  Right now we have entirely parallelized asynchronous jobs.  I agree somewhere we would have to sacrifice that parallelization to get logic about prior tasks completion.  Certainly tacking it on here w/o that assurance is incorrect.  

Alternatively, creation of default admin set could be tied to the *next* action of repo creation, namely creation of the first user account.  The logical cost is the user model suddenly having to know a lot about fcrepo, admin sets, and solr.  That seems unwieldy.

Allow me to suggest that we may need an additional stop at a "system status" screen with long-polling for the results of the other async jobs.  Add a status line for "default admin set" (status: needed).  Then when all the rest are green, enable a button to create one.  It can say "Next" and bounce the user to the user creation, for all I care, the point is to give the user feedback and give the task the conditions it needs to succeed.

jcoyne: After talking this over with @cbeer I can't think of a good way to do this.  We need our jobs to be idempotent (i.e. small) so they can fail and be retried, but we also need to have dependency management of jobs. I don't know what the solution is, but I don't think this is the place to add it.

mjgiarlo: @atz @jcoyne @cbeer An admittedly janky workaround that abides by both of the above principles might be a job that intentionally fails until both fcrepo and solr are present, and then fires off. Too janky?

jcoyne: Wow, that's thinking outside of the box.

jcoyne: The problem with this approach is that jobs are only allowed to fail a few times before being sent to the dead letter queue.  If your job hits that number before the solr or fedora jobs are done, your job will never complete it's work.

jcoyne: I think this is going to have to happen with `Operation`s.  Where we make some sort of composite operation for Solr + Fedora.  When both of those jobs have marked their part as done (last success), then the AdminSet job could be enqueued.

mjgiarlo: @jcoyne That sounds like it's worth pursuing.

atz: yeah, @mjgiarlo, intentional failure is not system-friendly.  I agree w/ @jcoyne re: failures and idempotence.  Job failure can trigger events to a separate error queue.  These should be taken seriously, not spammed.  

Our options include:
- **chaining**: defining (versions of) the jobs that know their successor, enqueuing it after their own completion.  Pro: logic.  Con: composability.
- **Tie to first user creation**: described above.  Pro: defers the timing of the problem (note: doesn't actually solve it!).  Con: conflates concerns.
- **UX step** explicitly waiting for prerequisites to complete (also above).  This also gives us space to expand if we add other components that, for whatever reason, might take longer than the couple seconds expected for a single HTTP response.

jcoyne:
The UX step option currently works great at the moment.  Unfortunately you have to know/remember to add the "magic" group "registered" as a depositor.

mjgiarlo:
@jcoyne @atz thx, y'all, for writing up the options. I don't have time to take this one forward anymore, so I'm happy to hand it off.

mjgiarlo:
@atz @cbeer @jcoyne: After kicking this around with Atz yesterday, I have decided to start by adding a new job to create a default Admin Set. A follow-up PR might be to sacrifice idempotency to handle account creation in a single job. This would be temporary, to ensure that default AS creation works during the community beta testing period (May 15th through June). Ultimately we should pursue the "UX step" approach outlined by @atz in #884, though we likely don't have time to complete and test that for a month or two.

mjgiarlo:
@jcoyne @atz re-review?

PR #892, opened 2017-03-27
Make the navbar smaller. Fixes #891

### LINKED ISSUE #891: scale down navbar-header
https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L365

35px should be enough.

jcoyne:
This matches the designs in https://ggeisler.github.io/lerna-dashboard/index.html

### DIFF
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -9,3 +9,5 @@ $jumbotron-button-background-color: #7ab55c;
 
 $features-section-background-color: #f4f4f4;
 $features-section-border-color: #ddd;
+
+$navbar-height: 35px;


mjgiarlo:
Thoughts, @projecthydra-labs/hyrax-ui-ux-advisors ?

justcolin:
Looks good. Nit picky detail: should the height be in `em` instead of `px`?

jcoyne:
@ggeisler would you care to weigh in regarding em vs px?

ggeisler:
I don't really have an opinion other than it seems to me there might be lots of places where we could use `em` instead of `px` so I don't know how we decide to switch from `px` to `em` in a given code update (especially in a case like this, where we are overriding something currently in `px`). Unless we have some guideline that helps developers decide when they should use `em` instead of `px`, this feels like something that might be better addressed independently from the actual feature updates.

For example, if @justcolin was willing to describe what aspects of the application we'd be better off using `em` for, we could do a systematic update in a dedicated PR to address the issue and not have to worry about it on a PR-by-PR basis.

mtribone:
The use of `em` is much better especially when people increase the font size either in their browser settings or by increasing or decreasing the text size using shortcuts. It means that the navbar will increase or decrease based on their choice. More flexible than `pixel` for setting sizes in the UI.

justcolin:
To tag onto what @mtribone said, though most browsers have switched over to more complicated zooming algorithms, some users have text sizes increased in their browser or OS. If `em`s or `rem`s are used the layout will not break for them. When `px`s are used often text overflows, overlaps, and/or gets cropped, while others can lose the ability to have the text scaled to the size they want since `px` are supposed to be an absolute unit.

Many have adopted `em`s and `rem`s (along with `%`) as the default units since it makes controlling some components easier. Since `em`s cascade, a change of size at a components root will cause the whole thing to evenly scale, which often makes it easier to adjust things. (Sometimes it does make it harder if one isn't careful, so I am not sure if it is fully a win there, but that's a discussion for another day.)

All that put together means I tend to prefer using `em`s, `rem`s, `%` and the other relative units by default, and to only use `px` when absolutely nessesary.

A nice article with more thoughts on this: https://css-tricks.com/rems-ems/

justcolin:
As for the strategy to deal with this kind of thing: there was discussion of developer guidelines before, and I looked into doing a crawl through the CSS to address things like this, but fixing things piece by piece as people work on the code seems to me like a more sustainable and agile approach. It gets these accessibility recommendations worked into people's daily practice, makes checking for these kinds things a part of general code review, and is less likely to cause bugs to leak into the system—as big PR's tend to do.

I'd love to hear other's thoughts, but that is probably a discussion to have elsewhere.

cbeer:
Upstream bootstrap 3.x uses `px`(and switches to `rem` in 4.x). I think staying consistent with upstream (in this case, `px`) is probably worthwhile, and switching over to `rem` after they release bootstrap 4.

PR #908, opened 2017-04-04
Reduce height of topbar when viewport is desktop size

ggeisler:
Our previous strategy for reducing the height of the topbar/brandbar adversely affected the main menu navigation bar. This PR scopes changes to the topbar and to desktop viewports, so the height is unaffected on mobile where the taller topbar is helpful for containing the taller expand menu icon.

### Desktop
after

### Mobile

hyku

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -1,4 +1,16 @@
 #masthead {
+  @media (min-width: $screen-sm-min) {
+    &.navbar {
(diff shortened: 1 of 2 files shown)


jcoyne:
this causes a gap to appear on the dashboard page.

screen shot 2017-04-04 at 8 15 04 pm

ggeisler:
I'm pretty sure that gap was there already (it was on my mental list to look at when I had time) but I'll fix it with a new PR.

PR #944, opened 2017-04-11
Temporarily turn off the rescuing of not found exceptions

jcoyne:
So that we can figure out why non-admins can't submit works.
Ref https://github.com/projecthydra-labs/hyrax/issues/753

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -26,5 +26,11 @@ module Hyku
     config.action_dispatch.rescue_responses.merge!(
       "I18n::InvalidLocale" => :not_found
     )
+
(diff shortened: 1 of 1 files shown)


atz:
Apparently we weren't testing for this behavior anywhere?

jcoyne:
This behavior is built int every rails application, so it's tested in rails itself

atz:
Why isn't it enough to do this locally?

jcoyne:
I'm only able to reproduce it on the demo server

atz:
Crap, this means we're going to get tons of Honeybadger notifications, right?

PR #961, opened 2017-04-14
Create superadmin:grant rake task.

### LINKED ISSUE #733: Create a rake task to add superadmin role to user account
This allows us to be sure a deployed instance has at least one user with superadmin access. (At that point, the superadmin should use the new superadmin UI to create/remove tenants, create/remove tenant admins, and so forth. This is covered in another ticket.)

Note that spotlight has such a rake task.

dlpierce:
Create superadmin:grant rake task.

Fixes #733

`rake superadmin:grant[user1@example.org]`
A rake task to grant the superadmin role to the specified users. 
'spec/support/rake.rb' is adapted from hyrax with added support for multiple task arguments.
find_by_user_key is whitelisted because it is actually a defined method of User..

### DIFF
--- a/lib/tasks/grant_superadmin.rake
+++ b/lib/tasks/grant_superadmin.rake
@@ -1,0 +1,14 @@
+namespace :superadmin do
+  desc 'Grant the superadmin role to specified users'
+  task 'grant', [:user_list] => [:environment] do |_cmd, args|
+    raise ArgumentError, 'A list of users is required: `rake superadmin:grant[user1,user2,...]`' unless args.user_list
+    args.to_a.each do |u|
+      user = User.find_by_user_key(u)
+      if user
+        user.add_role(:superadmin)
+      else
+        warn("Could not find user #{u}")
+      end
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,8 @@ Rails:
   Enabled: true
 
 Rails/DynamicFindBy:
+  Whitelist:
+    - find_by_user_key
   Exclude:
     - 'lib/importer/factory/object_factory.rb'
 
@@ -72,6 +74,7 @@ RSpec/DescribeClass:
     - 'spec/features/**/*'
     - 'spec/views/**/*'
     - 'spec/routing/**/*'
+    - 'spec/tasks/**/*'
 
 Rails/FilePath:
   Exclude:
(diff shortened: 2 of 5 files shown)


lib/tasks/grant_superadmin.rake
```diff
@@ -0,0 +1,14 @@
+namespace :superadmin do
+  desc 'Grant the superadmin role to specified users'
+  task 'grant', [:user_list] => [:environment] do |_cmd, args|
+    raise ArgumentError, 'A list of users is required: `rake superadmin:grant[user1,user2,...]`' unless args.user_list
+    args.to_a.each do |u|
+      user = User.find_by_user_key(u)
```
mjgiarlo: If you rework this as

```ruby
user = User.find_by(email: u)
```

I believe you could remove the white list entry to .rubocop.yml above.

lib/tasks/grant_superadmin.rake
```diff
@@ -0,0 +1,14 @@
+namespace :superadmin do
+  desc 'Grant the superadmin role to specified users'
+  task 'grant', [:user_list] => [:environment] do |_cmd, args|
+    raise ArgumentError, 'A list of users is required: `rake superadmin:grant[user1,user2,...]`' unless args.user_list
+    args.to_a.each do |u|
```
mjgiarlo: What does `args` look like here, before you call `to_a` on it? Does rake already do some processing on the comma-separated list of usernames?

lib/tasks/grant_superadmin.rake
jcoyne: I prefer it as is, because it encapsulates that email is the user key. We may change this later when we implement some other authN provider

lib/tasks/grant_superadmin.rake
dlpierce: `args` is actually a Rake::TaskArguments object which behaves a bit oddly when multiple arguments are allowed. Any arguments after the first in this task are not named arguments, so they are considered "extras". TaskArgument's .each ignores extras, but to_a captures everything.

lib/tasks/grant_superadmin.rake
dlpierce: That is exactly why I did it this way.

mjgiarlo: :shipit:

lib/tasks/grant_superadmin.rake
mjgiarlo: Perfect. Thanks, @dlpierce

PR #965, opened 2017-04-15
Job chaining

atz:
For discussion.

Built on top of the existing job.  So you could reasonably compose different chains that used the same jobs in different sequences.

The main problem is that the two most costly jobs that benefit the most from parallelization are solr and fedora creation, but those are the two that we need to have logic about them both being complete.  So we either:
- sacrifice parallelism between solr and fedora creation to ensure logic
- allow the downstream job (e.g. AdminSet creation) to "spin" waiting for
  preconditions to arrive
- use a UX step/page with status polling that accomplishes the same thing

Spinning (check, sleep, check, sleep... eventually timeout) is anti-pattern for queue mgmt.

### DIFF
--- a/app/jobs/create_account_fcrepo_job.rb
+++ b/app/jobs/create_account_fcrepo_job.rb
@@ -1,0 +1,6 @@
+# This job is the same as its parent, but chains to the next step
+class CreateAccountFcrepoJob < CreateFcrepoEndpointJob
+  after_perform do |job|
+    CreateDefaultAdminSetJob.perform_later(job.arguments.first)
+  end
+end
--- a/app/jobs/create_default_admin_set_job.rb
+++ b/app/jobs/create_default_admin_set_job.rb
@@ -1,0 +1,6 @@
+class CreateDefaultAdminSetJob < ActiveJob::Base
+  def perform(account)
+    AccountElevator.switch!(account.tenant)
+    AdminSet.find_or_create_default_admin_set_id
+  end
+end
--- a/spec/jobs/create_account_fcrepo_job_spec.rb
+++ b/spec/jobs/create_account_fcrepo_job_spec.rb
@@ -1,0 +1,14 @@
+RSpec.describe CreateAccountFcrepoJob do
+  let(:account) { FactoryGirl.create(:account) }
+
+  it 'lets parent class handle the work' do
+    expect(CreateFcrepoEndpointJob).to receive(:perform_now).with(account)
+    described_class.perform_now(account)
+  end
+  it 'chains to next job' do
+    expect(CreateDefaultAdminSetJob).to receive(:perform_later).with(account) do |acct|
+      expect(acct.fcrepo_endpoint.base_path).to eq "/#{account.tenant}"
+    end
+    described_class.perform_now(account)
+  end
+end
--- a/spec/jobs/create_default_admin_set_job_spec.rb
+++ b/spec/jobs/create_default_admin_set_job_spec.rb
@@ -1,0 +1,15 @@
+
+RSpec.describe CreateDefaultAdminSetJob do
+  let(:account) { FactoryGirl.create(:account) }
+
+  before do
+    allow(AccountElevator).to receive(:switch!)
+  end
+
+  describe '#perform' do
+    it 'creates a new admin set for an account' do
+      expect(AdminSet).to receive(:find_or_create_default_admin_set_id).once
+      described_class.perform_now(account)
+    end
+  end
+end


app/jobs/create_default_admin_set_job.rb
```diff
@@ -0,0 +1,6 @@
+class CreateDefaultAdminSetJob < ActiveJob::Base
+  def perform(account)
+    AccountElevator.switch!(account.tenant)
```
cbeer: Does the `ActiveJobTenant` (that automatically sets the tenant) not work here? If so, can we update it (or the way we build the job)?

app/jobs/create_account_fcrepo_job.rb
```diff
@@ -0,0 +1,6 @@
+# This job is the same as its parent, but chains to the next step
+class CreateAccountFcrepoJob < CreateFcrepoEndpointJob
```
cbeer: This approach seems completely reasonable (especially if `CreateDefaultAdminSetJob` is idempotent). Is there a reason to subclass the other job?

app/jobs/create_default_admin_set_job.rb
atz: I'm not sure I understand enough to answer that.  That's is a question for @mjgiarlo, who wrote the original code in his PR #884.

app/jobs/create_account_fcrepo_job.rb
atz: So that they share the same `#perform` implementation and reduce maintenance burden.  If we proliferated a number of different job chains, the value in keeping the extensions as light as possible is pronounced.  

The idea is to keep the job itself separate from its position in various sequences.  With this approach, a hypothetical rake task or admin button to `perform_now` could utilize the parent class and avoid chaining.  

If we decide we will never make use of such functionality, then no subclassing is needed.  But for purposes of demonstrating the minimal application of `after_perform`, this is ideal.

app/jobs/create_default_admin_set_job.rb
mjgiarlo: @cbeer It may work here. Would that involve something beyond removing line 3?

app/jobs/create_account_fcrepo_job.rb
mjgiarlo: @cbeer Yes, `CreateDefaultAdminSetJob` should be idempotent because `AdminSet.find_or_create_default_admin_set_id` is: https://github.com/projecthydra-labs/hyrax/blob/master/app/models/concerns/hyrax/admin_set_behavior.rb#L42-L47

mjgiarlo:
@atz You outline options above, and I hear you re: spinning. What we don't yet have a solution for in this PR is, now that we have a job coupling FCR endpoint creation and default AS creation, there is the possibility this job will fire before the Solr collection is created (which I think will cause an error). How do you recommend we proceed? Are there other folks we'd like to invite to this discussion?

atz:
@mjgiarlo This is not intended to be "merge this an it fixes a problem".  This was a demonstration of an approach that, if acceptable, could be further implemented to address the problem.  

You are correct about the sequence problem re: Solr, which would necessitate one of those 3 bullet points.  We can't have everything be peer async jobs and then want guaranteed sequence/preconditions.  The respective approaches would be:
- **sacrifice parallelism**: job chains both a `perform_now` for Solr creation and then the `perform_later` for the subsequent step;
- **spinning checks**: Need to centralize the status checking that currently pings from view template;
- **UX step**: I have started trying to wire in ActionCable to allow the Status page to receive broadcast events from the completion of the various jobs.  Also benefits from modeling system status as above, and needs to avoid crashing on uninitialized components.  

@cbeer responded in conversation that he would support spinning for this case, but only because each job is expected to complete relatively quickly.  When preconditions fail, this would bottleneck a given worker on a queue for a short timeout, say, 30 seconds.  I'm less inclined to follow that pattern because I can see it readily leading to failure.  If the jobs complete relatively quickly, then there sacrificing parallelism for the jointly required jobs is also a minimal cost.  

My own preference is for the UX step.  While it is more work, it also improves the status page and is a pattern that would benefit the stack.

mjgiarlo:
@atz @cbeer I wonder if it might be defensible to put in a quick patch that uses spinning (to "fix" the current behavior in our demo instance, ensuring all tenants have a default Admin Set), and then replace that small patch with the more substantial UX step once AC has been wired into our stack? I'm flexible.

app/jobs/create_account_fcrepo_job.rb
jcoyne: I think what cbeer is saying (and I agree with) is that we only need a single class (`CreateFcrepoEndpointJob`), we don't need two classes with an inheritance relationship as there is no other user of `CreateFcrepoEndpointJob`

atz: Right, I gather from these comments that this approach has been demonstrated adequately and is broadly acceptable.  I would agree that only one version of that class is necessary (right now).

atz:
I think this discussion was useful, but we can continue outside the PR format.

PR #966, opened 2017-04-18
Adds dc term extent to image works.

### LINKED ISSUE #877: add dc:extent element to Image Work metadata
A depositor of an Image Work needs the option to add description about the extent (size, duration, number, etc.) of the work.

Let's add support for this element: http://dublincore.org/documents/dcmi-terms/#terms-extent

It should be optional and repeatable.

andjsmit:
Fixes #877

Adds dc extent field to image works.

### DIFF
--- a/app/forms/hyrax/image_form.rb
+++ b/app/forms/hyrax/image_form.rb
@@ -3,6 +3,6 @@
 module Hyrax
   class ImageForm < Hyrax::Forms::WorkForm
     self.model_class = ::Image
-    self.terms += [:resource_type]
+    self.terms += [:resource_type, :extent]
   end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -4,6 +4,10 @@ class Image < ActiveFedora::Base
   include Hyrax::WorkBehavior
   validates :title, presence: { message: 'Your work must have a title.' }
 
+  property :extent, predicate: ::RDF::Vocab::DC.extent, multiple: true do |index|
+    index.as :stored_searchable
+  end
+
   # This indexer uses IIIF thumbnails:
   self.indexer = WorkIndexer
   self.human_readable_type = 'Image'
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -10,3 +10,9 @@ en:
     footer:
       copyright_html: "<strong>Copyright &copy; 2017 Project Hydra</strong> Licensed under the Apache License, Version 2.0"
       service_html: A service of <a href="http://projecthydra.org/" class="navbar-link" target="_blank">Project Hydra</a>.
+  simple_form:
+    hints:
+      defaults:
+        extent: "The extent (size, duration, number, etc.) of the work."
+      labels:
+        extent: "Extent"
--- a/config/locales/hyrax.es.yml
+++ b/config/locales/hyrax.es.yml
@@ -10,3 +10,9 @@ es:
     footer:
       copyright_html: "<strong>Copyright &copy; 2017 Proyecto Hydra</strong> bajo licencia de Apache, Version 2.0"
       service_html: Un servicio de <a href="http://projecthydra.org/" class="navbar-link" target="_blank">Proyecto Hydra</a>.
+  simple_form:
+    hints:
+      defaults:
+        extent: "La extensión (tamaño, duración, número, etc.) de la obra."
+      labels:
+        extent: "Extensión"


mjgiarlo:
Thanks for working on this, @andjsmit!

Dear @projecthydra-labs/hyrax-metadataists, could I ask for a quick comment on this issue when you get a minute?

ruthtillman:
@mjgiarlo You looking for approval of use of extent or?

cmh2166:
what's involved:

- adding dcterms:extent field to Work > Image resources in Hyrax
- obligation: {0,n} (not required, repeatable)
- indexed, searchable
- form provided def: "The extent (size, duration, number, etc.) of the work."
- expects/stores a string? (Can we confirm this? Because dcterms expects range of dcterms:SizeOrDuration)
- Image resource expects a title (only required element specific to this Work type?)

questions:

- do we diverge from the dcterms:SizeOrDuration expectation here? 
- would extent (once decided) be something we should push back to Basic_Metadata or other? Seems not specific to Images. 

proposed actions:

- I know some Hydra folks aren't fans of this type of semantic flexibility of using dcterms fields that expect non-literal values with literal values in practice. [We're not the only ones to fudge this though](https://dp.la/info/wp-content/uploads/2015/03/MAPv4.pdf). Alternatives:
    - start working on managing non-first class resources, i.e., dcterms:SizeOrDuration resources, or dcterms:MediaTypeOrExtent, or ld4l:Activity, etc. This seems unrealistic at present.
    - use alternative term for extent that's unconstrained: http://www.rdaregistry.info/Elements/u/#P60550
- if not in this PR, create workflow for eventually migrating metadata fields like extent from Work subclasses to Basic Metadata or Generic Work where applicable. I can take on reviewing what this looks like and queuing up PRs/requests as these become apparent. 
- Update expectations / understandings of how we will work within the semantic restraints (or not) of this term in our MAP. I can do this once we agree.

Depending on ease of integrating RDAU into Hyrax MAP, that's my slight preference here. But I'm not opposed to use of dcterms:extent with a literal and just documenting our transgression, which we already do with dcterms:rights. Can we have a day to get input and decide? Would be good to pin down how 'flexible' we want to be as a general rule.

geekscruff:
Given that DPLA and EDM both fudge this element to use a literal, I think we should let this through (for now).  But I don't see any reason it should only exist for Images, it isn't really very image specific.

I do think there should be a broader consensus on 'semantic flexibility', but that can happen outside this PR IMHO.

ruthtillman:
Thanks very much, Christina. A question I initially had re: extent was why this was being added specifically for Image (I saw the Issue it addressed but not an expectation of what function it would serve--is this for something which could perhaps be programmatically generated like the dimensions of the image or the filesize of the image) and then whether it should simply be pushed back into everything which could theoretically have an extent.

I agree re: non-first class objects particularly for dcterms:SizeOrDuration. It's the kind of thing where any utility derived from doing that would as easily be derived from harvesting and working with the literals. Perhaps we could document it that we're storing it as a literal which we expect to conform to SizeOrDuration? 

At least with something like FileFormat, there are a theoretically limited number of them and controlling that through non-first class objects would make more sense. I don't know, maybe this would be a good thing for metadatists to review? We could compile a list of classes for which there _should_ be existing objects like FileFormat "application/pdf" or "text/csv." And then leave the others for people to either choose to implement locally or to map to literals with documented assumption that they conform to class. Or maybe this is adding too much work/too many layers of work.

remerjohnson:
@cmh2166 I see that you referenced `Work > Image` resources and I wanted to ask, is there a doc that specifies all those members of work that aren't the generic Object?

atz:
I'm seeing a lot of discussion, but no intent to block this change, so barring objection, I'll merge tomorrow.

cmh2166:
Seems to be what is mostly agreed upon, with downstream work metadata folks need to do being outlined.

Fwiw, didn't feel like there was a ton of discussion here though.

On Wed, Apr 19, 2017 at 7:48 PM Joe Atzberger

<
mailto:Joe Atzberger <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

I'm seeing a lot of discussion, but no intent to block this change, so barring objection, I'll merge tomorrow.

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/projecthydra-labs/hyku/pull/966#issuecomment-295517302
, or

atz:
For literally 4 lines of code, this was a lot of expertise brought to bear, imo.  A lot of the policy/community level questions are not going to get resolved one way or the other by the inclusion or exclusion of those 4 lines.

cmh2166:
For sure, I get your point. More thinking this is a common and involved metadata/modeling question that has larger threads we should pin down, and it just happened to show up here. Hoping we can catch these kind of minor code shifts with not minor modeling questions attached earlier in the process in the future- it would make all of this easier.

Christina

On Wed, Apr 19, 2017 at 8:14 PM Joe Atzberger

<
mailto:Joe Atzberger <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

For literally 4 lines of code, this was a lot of expertise brought to bear, imo. A lot of the policy/community level questions are not going to get resolved one way or the other by the inclusion or exclusion of those 4 lines.

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/projecthydra-labs/hyku/pull/966#issuecomment-295526456
, or

PR #979, opened 2017-04-21
Apply page header css to groups admin pages.

### LINKED ISSUE #946: Update how page header for Manage Groups is rendered
The HTML around the title of Manage Groups page is not the same as other pages in the Administration section, so the rendered title of the page doesn't match other Admin pages in terms of styling. 

This appears to be because the code for the `h1` in the Manage Groups page is not wrapped by a `<% content_for :page_header do %>` (Hyku) or `<% provide :page_header do %>` (Hyrax). So the rendered `h1` does not get wrapped by a `<div class="row">  <div class="col-xs-12 main-header">`. It should be wrapped by those `divs` when rendered so the appropriate CSS is applied.

(I'd try to make the update myself but am not up-to-speed with the difference between what we're doing with the page headers in Hyrax vs. Hkyu.)

### Current appearance
manage_groups____administration____hyku 2

### Target appearance
(Might be hard to tell the difference from screenshot so best way is to compare code around the `h1` to other Admin pages.)
manage_groups____administration____hyku

dlpierce:
Fixes #946 

Apply page header css to groups admin pages.

Uses content_for to add surrounding divs to the page header so they match the style of other admin pages.

### DIFF
--- a/app/views/admin/groups/edit.html.erb
+++ b/app/views/admin/groups/edit.html.erb
@@ -1,5 +1,5 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.edit'), t('hyku.admin.title')) %>
-<% content_for :page_header do %>
+<% provide :page_header do %>
   <h1>
     <span class="fa fa-users"></span>
     <%= t('hyku.admin.groups.title.edit') %>
--- a/app/views/admin/groups/edit.html.erb
+++ b/app/views/admin/groups/edit.html.erb
@@ -1,7 +1,9 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.edit'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.edit') %>
-</h1>
+<% content_for :page_header do %>
+  <h1>
+    <span class="fa fa-users"></span>
+    <%= t('hyku.admin.groups.title.edit') %>
+  </h1>
+<% end %>
 <%= render 'nav' %>
 <%= render 'form' %>
--- a/app/views/admin/groups/index.html.erb
+++ b/app/views/admin/groups/index.html.erb
@@ -1,8 +1,10 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.index'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.index') %>
-</h1>
+<% content_for :page_header do %>
+  <h1>
+    <span class="fa fa-users"></span>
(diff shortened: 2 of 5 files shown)


app/views/admin/groups/edit.html.erb
```diff
@@ -1,7 +1,9 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.edit'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.edit') %>
-</h1>
+<% content_for :page_header do %>
```
jcoyne: This should be `provide` rather than `content_for`

app/views/admin/groups/index.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.index'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.index') %>
-</h1>
+<% content_for :page_header do %>
```
jcoyne: Use `provide` rather than `content_for`

app/views/admin/groups/new.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.new'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.new') %>
-</h1>
+<% content_for :page_header do %>
```
jcoyne: Use `provide` rather than `content_for`

app/views/admin/groups/remove.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.edit'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.edit') %>
-</h1>
+<% content_for :page_header do %>
```
jcoyne: Use `provide` rather than `content_for`

app/views/admin/groups/users.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :page_title, construct_page_title(t('hyku.admin.groups.title.edit'), t('hyku.admin.title')) %>
-<h1>
-  <span class="fa fa-users"></span>
-  <%= t('hyku.admin.groups.title.edit') %>
-</h1>
+<% content_for :page_header do %>
```
jcoyne: Use `provide` rather than `content_for`

mjgiarlo:
@jcoyne re-review?

PR #989, opened 2017-04-25
Enable multitenancy in test (CI)

atz:


spec/requests/home_spec.rb
```diff
@@ -1,31 +1,29 @@
 RSpec.describe 'Home page', type: :request do
-  context 'without a current tenant', multitenant: true do
-    before do
-      allow(Settings.multitenancy).to receive(:admin_host).and_return('localhost')
-    end
-
+  context 'with multitenancy', multitenant: true do
```
jcoyne: remove `multitenant`?

spec/support/multitenancy_metadata.rb
```diff
@@ -3,16 +3,38 @@
...
-    if example.metadata[:multitenant]
+    if example.metadata[:multitenant] || example.metadata[:singletenant]
       allow(Settings.multitenancy).to receive(:enabled).and_call_original
       Rails.application.reload_routes!
     end
   end
+
+  # There are 3 optional flags available to a test block.  Only ONE will be active
+  # at any given time.  They are (with areas of likely use):
+  #   :multitenant  - general case default, only needs to be explicit for types described below
```
jcoyne: Can we collapse this to a single setting? `tenancy: :single`, `tenancy: :multi`, `tenancy: :fake`?

jcoyne:
Should this all be squashed to a single commit?

atz:
I did a ton of squashing already.  There are distinct changes, in particular chronicling enabling the basic setting and then correcting components along the way.  It *could* be any number of commits, but the fewest I would do would be 5.

spec/support/multitenancy_metadata.rb
atz: If I felt like there were going to be more modes added regularly, I would.  That doesn't seem likely.

atz:
@jcoyne squashed down to 5 logically distinct commits.

mjgiarlo:
@jcoyne re-review?

spec/support/multitenancy_metadata.rb
jcoyne: My concern is that one could set both `:multitenant` and `:singletenant` and the behavior would be undefined.

atz:
@jcoyne No, the result IS defined, by the conditional.  Multitenant would "win".

atz:
Unless anybody wants to concur w/ Justin that reorganizing the flags to be values of single flag is worth blocking this over (I agree w/ him in principle, I just don't think it is worth blocking over), then with due respect, I would proceed on basis of Mike's approval.

mjgiarlo:
@atz :shipit:

PR #992, opened 2017-04-26
Remove redundant action

atz:
`before_action :ensure_admin!` is already part of `AdminController`.


atz:
Looks like it is.  Wasn't part of the logic I was tracing, so I didn't notice it.  Will amend.

PR #1011, opened 2017-04-27
Referential integrity for endpoint IDs

atz:
Only one account should be able to reference a given **[solr|fcrepo|redis]_endpiont**.  It is important to understand that Rails validations never truly provide guarantees of uniqueness in a multi-server application.  It **must** be done at the database level, since at runtime, the database is the only component used in common.  

For more background, see:
https://robots.thoughtbot.com/the-perils-of-uniqueness-validations

Hyku is fortunate, since it is already using postgres in all environments and postgres is perfectly capable of handling referential integrity.

### DIFF
--- a/db/migrate/20170427003546_add_index_to_accounts.rb
+++ b/db/migrate/20170427003546_add_index_to_accounts.rb
@@ -1,0 +1,15 @@
+class AddIndexToAccounts < ActiveRecord::Migration[5.0]
+  def change
+    # Remove first, else error like:
+    # Index name 'index_accounts_on_solr_endpoint_id' on table 'accounts' already exists
+    remove_index :accounts, :solr_endpoint_id
+    remove_index :accounts, :fcrepo_endpoint_id
+    remove_index :accounts, :redis_endpoint_id
+    add_foreign_key :accounts, :endpoints, column: :solr_endpoint_id, on_delete: :nullify
+    add_foreign_key :accounts, :endpoints, column: :fcrepo_endpoint_id, on_delete: :nullify
+    add_foreign_key :accounts, :endpoints, column: :redis_endpoint_id, on_delete: :nullify
+    add_index :accounts, :solr_endpoint_id, unique: true
+    add_index :accounts, :fcrepo_endpoint_id, unique: true
+    add_index :accounts, :redis_endpoint_id, unique: true
+  end
+end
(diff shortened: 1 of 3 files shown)


jcoyne:
Before we merge this, can we first address this issue? https://github.com/projecthydra-labs/hyku/issues/1012#issuecomment-297899600

atz:
I don't see any reason to defer this.  The rails uniqueness validation will _never_ be truly sufficent:
https://robots.thoughtbot.com/the-perils-of-uniqueness-validations

jcoyne:
@atz I just want to make sure we can fix the current code, before inserting new code. I think that will make it easier to validate we have the existing code in the correct state. Rather than having broken code that relies on your new code.

atz:
Your proposed fix is also "new code", but I wholeheartedly support it (in abstract).  However, it does nothing to deal with the fundamental limitations of rails uniqueness validations, which **still should be addressed**.  The behaviors (and failures) are distinct.  I test for the behavior introduced here without removing or altering any existing tests.  Our ability to reason about the overall system behavior is improved by both approaches and there is no reason to prioritize one over the other (in particular when this one is already done).

mjgiarlo:
@atz @jcoyne Is an argument to tackle #1012 in advance of merging this PR that this PR will immediately bork the demo instance?

mjgiarlo:
Either way, I'd like to see both taken on today/ASAP! I don't care much about the order unless this PR breaks the demo instance.

atz:
This is not projected to break demo, since as it turns out, the duplication of endpoint IDs was not the mechanism of failure.  I.e., the corrupt data is not in the fields being further constrained here.  

Both approaches *would* fail on `cname`, where they are more urgently needed.  The DB-level constraint would fail in migration (DB cannot accept new constraint on existing data), the model-level validation would fail inasmuch as it would be unable to update or save one of the accounts that already has invalid (duplicate) `cname`.  Both approaches require cleaning the existing data.

atz:
This initiative was discussed to widespread consensus earlier in the week.  The changes are concise, well-defined and tested. @jcoyne has suggested there is some benefit to handling `cname` before `[solr|fedora|redis]_endpoint_id`.  Technically speaking, there is none.  

My analysis yesterday located the duplicate `cname` values in different Account rows.  That is what I would address next.  Delaying this PR helps that effort not at all.

PR #1039, opened 2017-05-02
Remove unused "title" field from accounts

atz:
Pretty straight forward.

### DIFF
--- a/app/views/account_sign_up/new.html.erb
+++ b/app/views/account_sign_up/new.html.erb
@@ -15,13 +15,6 @@
             </ul>
           </div>
         <% end %>
-        <div class="form-group">
-          <%= f.label :title, 'Title', class: 'col-md-2 control-label required' %>
-          <div class="col-md-10">
-            <%= f.text_field :title, class: 'form-control' %>
-            <span class="help-block">A human-friendly title for the repository. This can be changed later.</span>
-          </div>
-        </div>
 
         <div class="form-group">
           <%= f.label :name, 'Short name', class: 'col-md-2 control-label' %>
--- a/app/views/proprietor/accounts/new.html.erb
+++ b/app/views/proprietor/accounts/new.html.erb
@@ -18,14 +18,6 @@
               </div>
             <% end %>
             <div class="form-group">
-              <%= f.label :title, 'Title', class: 'col-md-2 control-label required' %>
-              <div class="col-md-10">
-                <%= f.text_field :title, class: 'form-control' %>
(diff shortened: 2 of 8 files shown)


app/views/account_sign_up/new.html.erb
```diff
@@ -15,13 +15,6 @@
             </ul>
           </div>
         <% end %>
-        <div class="form-group">
-          <%= f.label :title, 'Title', class: 'col-md-2 control-label required' %>
-          <div class="col-md-10">
-            <%= f.text_field :title, class: 'form-control' %>
-            <span class="help-block">A human-friendly title for the repository. This can be changed later.</span>
```
jcoyne: @ggeisler was this intended for use somewhere?

jcoyne:
Squash?

app/views/account_sign_up/new.html.erb
ggeisler: I don't know. If I had something to do with that code it was a year ago when I was just trying to make the pages to be shown in the OR demo presentable. 

If these pages are being/will be updated, it seems pretty safe to remove it if it doesn't seem necessary since I don't think anything would be used beyond this context of sign in/sign up.

atz: Yeah, the problem is that whatever **intent** never got **implemented**.  Once submitted, `@account.title` is not used.

mjgiarlo: @jcoyne @ggeisler @atz @hannahfrost I wonder if we might use the value of title to pre-populate the "Application Name" label that is available here: /site/labels/edit They *feel* like they're similar enough.

atz: @mjgiarlo Right, that makes sense.  Does that make it part of the Account model though?

atz:
Production table looks like this:
```
select id,cname,name,title from accounts;
 id |                cname                 |      name       | title 
----+--------------------------------------+-----------------+-------
  1 | cubs.demo.hydrainabox.org            | cubs            | 
  2 | hyphy.demo.hydrainabox.org           | hyphy           | 
  3 | iu-repo.demo.hydrainabox.org         | IU Repo         | 
  4 | troubleshooting.demo.hydrainabox.org | troubleshooting | 
(4 rows)
```
Even when we had dozens of accounts, title was never populated.

app/views/account_sign_up/new.html.erb
mjgiarlo: No, not part of the Account model, as I see it.

atz:
@jcoyne squashed

PR #1048, opened 2017-05-03
[WIP] Solr namespace for this hydra app (avoid collision with other hydra apps)

darrenleeweber:
hydra -> hyku

- might be rebased, pending #1052, which extracts commits here without the namespace change.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -16,8 +16,14 @@ action_controller:
 redis:
   default_namespace: hyku
 
+<%
+  # Read the config/solr.yml
+  # - note that it specifies a URL that includes the collection path, which is stripped below.
+  solr_config = YAML.load(ERB.new(IO.read(Rails.root + 'config' + 'solr.yml')).result)[Rails.env].with_indifferent_access
+%>
+
 solr:
-  url: "http://127.0.0.1:8983/solr/"
+  url:  <%= solr_config["url"].gsub(/solr\/.*/, 'solr/') %>
   configset: hyku
   configset_source_path: <%= File.join(Rails.root, 'solr', 'config') %>
   collection_options:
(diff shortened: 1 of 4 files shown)


darrenleeweber:
If coveralls coverage decreased in this PR, there must be some hard-coded strings in the specs somewhere, or some general assumptions about the solr namespace that need to change.  The namespace should be configurable.  Does this PR miss some config settings?

mjgiarlo:
@darrenleeweber I believe some of our underlying components generate config files with `hydra-development` and `hydra-test` in them. I'm not sure why changing these values wouldn't work, though.

Thoughts, @jcoyne?

config/settings.yml
```diff
@@ -16,8 +16,14 @@ action_controller:
...
 
+<%
+  # Read the config/solr.yml
+  # - note that it specifies a URL that includes the collection path, which is stripped below.
+  solr_config = YAML.load(ERB.new(IO.read(Rails.root + 'config' + 'solr.yml')).result)[Rails.env].with_indifferent_access
+%>
+
 solr:
-  url: "http://127.0.0.1:8983/solr/"
+  url:  <%= solr_config["url"].gsub(/solr\/.*/, 'solr/') %>
```
jcoyne: Settings.yml should be the authoritative source, don't change these

darrenleeweber: This concern to be addressed in #1052 and see what happens, this PR is pending that debate and needs to be rebased or whatever.

atz: I agree w/ @jcoyne here.  Also, the purpose of *this* PR is to do the namespacing.  If we can keep it separate from config consolidation, all the better.  If we can't (for testing consistency reasons), then we should prefer for Settings to be the authoritative source.

PR #1051, opened 2017-05-03
Update to rsolr 2.0

jcoyne:


atz:
`"solr" is not registered on Faraday::Adapter` in CI

PR #1052, opened 2017-05-03
[WIP] Solr dedup configs, consolidate into settings

darrenleeweber:
Considate solr configuration values in `config/solr.yml` and load them in other config files (using ERB).  Extracted from #1048 to see if this passes specs without changing the solr namespace.


atz:
I like having one authoritative source of configurations for solr.  Would be better if we could have components stop looking for different files, but this is a good step.

darrenleeweber:
The settings is a good place for app-specific configuration details.  This targets the service config, basically the URL.  I don't really like that `config/solr.yml` includes default paths for the collection etc. but I didn't want to change that here in this PR.  Also, changing settings is much bigger change than the intention of this PR and I propose that is done in a new PR.  That work will have to consider how the config gem merges settings across Rails.env files in the config/settings/*.yml files.  This work gets more into the details of how to represent the components of the Solr connection, etc. and where the responsibility lies for basic service settings vs. application-specific details.  (I'm thinking about the possibility that a system could set a global ENV for the base Solr URL, or mainly the PORT I guess, and other apps can build on that.)

darrenleeweber:
Also, the `config/solr.yml` already uses `ENV['SOLR_URL']`, so the ability to override locally is still possible when using this PR (I think).

darrenleeweber:
The last commit attempts to separate the Solr base URL from the collection details and uses the same collection structure in the `.solr_wrapper`, i.e.
```
collection:
  dir: solr/config/
  name: hydra-development
```
However, that's failing specs, whereas the prior refactor was good to go.

config/solr.yml
```diff
@@ -1,8 +1,28 @@
 # This is a sample config file that points to a solr server for each environment
 # TODO: there's also a Settings.solr.url configuration that is used in multitenant environments
+
+<%
+  namespace = 'hydra'
```
jcoyne: This is not an erb parsed file.

darrenleeweber:
ActiveFedora uses a YAML.safe_load that doesn't like YAML aliases for some reason, e.g.
```ruby
irb(main):001:0> solr_config = YAML.load(ERB.new(IO.read(Rails.root + 'config' + 'solr.yml')).result)[Rails.env].with_indifferent_access
=> {"collection"=>{"dir"=>"solr/config/", "name"=>"hydra", "namespace"=>"hydra-test"}, "url"=>"http://127.0.0.1:8985/solr/hydra-test/"}
irb(main):002:0> solr_config = YAML.safe_load(ERB.new(IO.read(Rails.root + 'config' + 'solr.yml')).result)[Rails.env].with_indifferent_access
Psych::BadAlias: Unknown alias: default
	from /users/dlweber/.rbenv/versions/2.4.1/lib/ruby/2.4.0/psych/visitors/to_ruby.rb:402:in `visit_Psych_Nodes_Alias'
```

darrenleeweber:
Blocked by https://github.com/projecthydra/active_fedora/pull/1240

cbeer:
It's not clear to me what value we're creating here -- do you think you could summarize the goal in as a new ticket (separate from the implementation)? I think I see an attempt at arriving at a single point of "truth" for this configuration, perhaps? In any case, having a better idea of what you're trying to do will help the team give more useful feedback and suggestions.

Right now, I look at this and see straightforward but redundant/duplicative configuration replaced with less-clear and/or more complex code. I'm hoping with more context, I'd be able to offer a more fruitful path.

darrenleeweber:
ActiveFedora skips Settings, it treats solr.yml as the "source of truth".  Replacing fixed strings that are duplicated multiple places with relatively simple code to parse a single YAML file makes a lot of sense to me.  The inital PR was a simple refactor to remove duplication.  The review commentary lacks specific recommendations and alludes to vague preferences for an alternative solution that has no specific implementation consequences for this PR, as yet.  Does anyone want to or intend to propose to remove solr.yml to be replaced by settings?  (The issue was discussed at length in slack commentary and, as yet, has not materialized in a specific issue; but it's clear that various commentary indicates the solr infrastructure is problematic and various attempts to resolve it seem to have proven frustrating.  Rather than waterfall the entire problem, this PR was intended to simply replace duplicate config details.  Given reluctance to move on this, I'm prepared to let it go and move on to other things.  Having said that, I think the current PR together with an ActiveFedora PR are close to achieving some useful consolidation.)

darrenleeweber:
It's not obvious how the ActiveFedora changes are made available to this PR.  A very quick attempt to update AF fails (the latest changes may not be published yet anyway), i.e.
```
$ bundle update active-fedora
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies......
Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    active-fedora (>= 11.1.4) was resolved to 11.1.4, which depends on
      activesupport (< 6, >= 4.2.4)

    active-fedora (>= 11.1.4) was resolved to 11.1.4, which depends on
      active-triples (~> 0.11.0) was resolved to 0.11.0, which depends on
        activesupport (>= 3.0.0)

    active-fedora (>= 11.1.4) was resolved to 11.1.4, which depends on
      activemodel (< 6, >= 4.2) was resolved to 5.1.0, which depends on
        activesupport (= 5.1.0)

    rails (= 5.0.2) was resolved to 5.0.2, which depends on
      activesupport (= 5.0.2)

    active-fedora (>= 11.1.4) was resolved to 11.1.4, which depends on
      solrizer (< 5, >= 3.4) was resolved to 4.0.0, which depends on
        activesupport
```

I'm reluctant to change the Gemfile for AF, to use a git repo instead of a published gem, but curious to see if it helps this PR to pass specs.

darrenleeweber:
https://github.com/projecthydra-labs/hyku/pull/1052/commits/499cc1a0e89e412e4ad147f71cc9f6c276e0a851 is purely an experimental exploration of whether this PR can pass specs using the AF changes to loading solr.yml safely.

darrenleeweber:
I think this actually passing specs now; the build failed for something that seems to have nothing to do with specs.  No clear stacktrace, might be something to do with cleanup or service shutdowns?

Gemfile
```diff
@@ -25,7 +25,7 @@ gem 'jbuilder', '~> 2.0'
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 
-gem 'active-fedora', '>= 11.1.4'
+gem 'active-fedora', git: 'https://github.com/projecthydra/active_fedora'
```
darrenleeweber: EXPERIMENTAL - do not merge this.

darrenleeweber:
This is blocked by:
- reviewer requirements, need clarification or split off an issue to be resolved elsewhere
- ActiveFedora release and bundle update
  - https://github.com/projecthydra/active_fedora/wiki/Release-management-process#activefedora-release-management-step-by-step

atz:
@jcoyne Settings isn't used by several of the components in the stack, so your complaint might be directed at them, but isn't practicable at this level.

cbeer:
I think we're talking past each other here.

@darrenleeweber thanks for that reply. If we're trying to reduce duplication in the stack, what do you think about @jcoyne's suggestion to use the settings files as that source of truth? I think both of us believe the settings files would be a more natural fit and would allow us take advantage of its built-in environment variable parsing, rails environment resolution, etc.

@atz I don't see how other components in the stack factor in here? This is for the hyku application, where we /are/ using Settings. Upstream components shouldn't need to know how they are configured. We should be able to easily drop in settings into the existing configuration files even, e.g.:

```yaml
# ./config/solr.yml
# This is a sample config file that points to a solr server for each environment
# TODO: there's also a Settings.solr.url configuration that is used in multitenant environments
development:
  url: <%= Settings.solr.some_new_setting_name || ENV['SOLR_URL'] || "http://127.0.0.1:#{ENV.fetch('SOLR_DEVELOPMENT_PORT', 8983)}/solr/hydra-development" %>
test:
  url: <%= Settings.solr.some_new_setting_name || ENV['SOLR_URL'] || "http://127.0.0.1:#{ENV.fetch('SOLR_TEST_PORT', 8985)}/solr/hydra-test" %>
production:
  url: <%= Settings.solr.some_new_setting_name || ENV['SOLR_URL'] || "http://your.production.server:8080/bl_solr/core0" %>
```

As I said in my earlier comment, though, this is still somewhat unsatisfying as a general approach, and we may want consider dynamically configuring our dependencies (allowing us to eliminate these  configuration files entirely).

darrenleeweber:
@cbeer - the additional details help to clarify what your thinking and suggesting as an alternative solution, just looking at the Settings injection into solr.yml for the moment (because removing all of these yaml files is beyond the scope of this PR, although work on this PR certainly begs the question(s)).

- I had not thought that Settings could be available in solr.yml, so that's a new thought and turns this PR upside down entirely.  Having said that, it is a proposal that has concrete implementation consequences for changing this PR and it could be done in this PR.
  - In the given code snippet for solr.yml (thanks!), I just wonder whether it needs anything other than a reference to Settings.solr data.  Assuming that is always available (initialized for this file), then it should take care of using `ENV['SOLR_URL']` and otherwise setting a default.  If for some reason the Settings.solr is not available, would that be a bug or expected behavior (depending on the access patterns to solr.yml)?One more question about that; should `ENV['SOLR_URL']` always come first?

cbeer:
I think it's safe to assume `Settings` will always exist. If it doesn't, something has gone incredibly wrong with the rails initialization and failing fast is a good idea.

One reason to keep `ENV['SOLR_URL']` in the mix is to maintain compatibility with our AWS and docker deploys. Putting it first makes sense to me, although you could might consider using `ENV['SOLR_URL']` in the settings file in the first place? 

> because removing all of these yaml files is beyond the scope of this PR

I'd just like to re-emphasize the importance of describing the problem separate from its implementation. Without that, it is difficult (or impossible) for other reviewers to know what you consider in or out of scope.

mjgiarlo:
I'm seeing lots of good discussion here. If we don't wrap up this issue by the end of the day, let's table this so we can concentrate our energy on the issues in Waffle and return to this at a later date. Thanks!

darrenleeweber:
Adopted the reviewer's approach to consolidate into Settings.solr, need to tweak it to get specs passing and adapt to any further review requests on the details in Settings.

darrenleeweber:
OK, so all the CI checks are good with this.  Requires:
- another round of reviews and possibly changes required
  - there might be details to modify in settings.yml
    - the current defaults use erb-code to leverage Rails.env and ENV values
  - possibly details to be added or moved to config/settings/*.yml
- ActiveFedora release and bundle update for that

darrenleeweber:
Moved RAILS_ENV settings into `config/settings/{RAILS_ENV}.yml` to see what that looks like and whether specs are passing.

darrenleeweber:
There is one test failure on travis:
```
 1) SolrEndpoint#connection returns the initialized connection (without an adapter option)
     Failure/Error: af_defaults = ActiveFedora::SolrService.instance.conn.options

       RSolr received :connect with unexpected arguments
         expected: ({"read_timeout"=>120, "open_timeout"=>120, "url"=>"http://example.com/solr/"})
              got: ({:read_timeout=>120, :open_timeout=>120, :url=>"http://127.0.0.1:8985/solr/hydra-test/"})
```
I can try track that down.

darrenleeweber:
Weird, just re-running the PR build fixed it.  Might be something fragile in that.

darrenleeweber:
Rebased on master to get up to date and double check the specs are good.

darrenleeweber:
Leaving commit history for review and intend to squash down if/when this gets accepted.

darrenleeweber:
Started a PR to release a new version of ActiveFedora:
- https://github.com/projecthydra/active_fedora/pull/1242

atz:
@cbeer: thx for the example, which I agree is a way to make Settings authoritative.  The limitation I was referring to was that the files and their various structures are still required by components.  Getting rid of the duplicative files is the broader goal.

So when @jcoyne says `Shouldn't settings be the authorative source?`, I think we are all agreeing that makes sense here.  However following that with `That way we can set env vars to override locally` is an orthogonal idea.  You can use ENV var overrides in any configuration and, as the PRs I put in just days ago demonstrated, you **must** do it in every place that is authoritative or something breaks.

darrenleeweber:
@cbeer - wondering whether to add a `SolrWrapper` init that uses the `Settings` into the `Rakefile`, as in
```
SolrWrapper.default_instance_options = {
    verbose: true,
    cloud: Settings.solr.cloud,
    port: Settings.solr.port,
    version: '5.3.1',  # no value in Settings.solr yet
    instance_dir: 'solr',  # no value in Settings.solr yet
    download_dir: 'tmp'  # no value in Settings.solr yet
}
require 'solr_wrapper/rake_task'
```
I'm guessing this is a very useful idea for development.  Hopefully it has no implications for production (maybe the solr_wrapper rake tasks could (or already do) check the Rails.env and ensure they do nothing on production).

mjgiarlo:
Given the earlier discussion:
> I'm seeing lots of good discussion here. If we don't wrap up this issue by the end of the day, let's table this so we can concentrate our energy on the issues in Waffle and return to this at a later date. ... [@mjgiarlo]

It's a good time to pause work on this so we can come back to it when there aren't so many other pressing issues on our plates. Cool with that?

darrenleeweber:
Rebased on master and updated to active-fedora 11.2

PR #1063, opened 2017-05-06
Update Hyrax and pull in new Blacklight translations from latest Hyrax

mjgiarlo:


config/locales/blacklight.en.yml
```diff
@@ -1,3 +1,45 @@
 en:
   blacklight:
     application_name: 'Hyku'
+    search:
```
jcoyne: Why don't we let Blacklight provide these?

jcoyne: Nm

mjgiarlo:
@jcoyne Isn't Hyku the local app? Where else would you put these in Hyku?

jcoyne:
Sorry, thought this was hyrax.

PR #1064, opened 2017-05-08
User - remove comment reference to account

darrenleeweber:
User code should not refer to `account` in this App, as this is an important distiction in this app.


darrenleeweber:
There's a bug in the specs somewhere - something is fragile - this PR makes no code changes, the `pr` build should not fail when the `push` build is passing.

PR #1067, opened 2017-05-08
Test methods individually

atz:
Because these have bearing on our rails-level detection for unique cname.

See:
https://github.com/projecthydra-labs/hyku/issues/1012#issuecomment-297899330


jcoyne:
@atz It appears that rubocop has a few changes for you.

atz:
Eh, had those fixed but apparently dropped the one-liner out in a rebase triage.  Fixed again.

atz:
@jcoyne does this represent what you intended?

PR #1075, opened 2017-05-08
[EXPERIMENTAL - DO NOT MERGE]  Travis debugging

darrenleeweber:
Travis - use `rake spec --trace` to identify build bugs

Also related to https://github.com/projecthydra-labs/hyku/issues/1085

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ dist: trusty
 rvm:
   - 2.4.1
 
+services:
+  - postgresql
+  - redis-server
+
 # We provide AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY encrypted by travis here
 env:
   global:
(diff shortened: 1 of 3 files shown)


jcoyne:
@darrenleeweber what's the status of this experiment?  If it's complete, can we delete the branch?

darrenleeweber:
The travis builds have been flaky, that's why this branch exists.  The root cause of travis flaky builds is not clear yet and not identified in this experiment.  The first guess is that it had something to do with aws deployments.  As soon as the `echo` were added on this branch, the deployment "failures" seemed to resolve.
I'm not a huge fan of having PRs depend on a successful archive and deposit process to an S3 data store - that seems to be superfluous to the actual specs on this app.  Why should a PR fail when all the specs are good but travis cannot deposit and archive on S3?

jcoyne:
@darrenleeweber so what's the plan now?

darrenleeweber:
I suppose I'd like to morph this PR into something that can run the S3 archives without crashing/failing the specs build.  I'm thinking of wrapping the S3 deposit in a ruby script to add some exception handling and ensure that it always returns OK - maybe sends a notification if someone wants to know when it fails?

darrenleeweber:
Closing this, as discussion elsewhere has indicated we do not want to cache the solr/fcrepo downloads and we do not want to use the `rake --trace` option.  The deploy-status changes have been extracted to another PR #1355, in case they are useful.

PR #1079, opened 2017-05-09
Revert "Fix extra space below top bar"

### LINKED ISSUE #1078: Navbar buttons are mis-styled, should fill entire navbar area.
screen shot 2017-05-09 at 7 47 55 am

cbeer:
Fixes #1078.

@ggeisler you were trying to fix the top navbar height in #945, but the fix broke other navbar styling (see #1078). This commit reverts it, and changes the top navbar to the default 50px, so at least nothing looks broken.

How important is it that the top navbar be 35px high? Is 50px something we can live with for now, or do we need to come up with a better fix?

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -1,4 +1,16 @@
 #masthead {
+  @media (min-width: $screen-sm-min) {
+    &.navbar {
+      min-height: 35px;
+    }
+
+    .navbar-nav > li > a,
+    .navbar-brand {
+      height: 35px;
+      padding-top: 7.5px;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


ggeisler:
@cbeer Yeah, I must've only checked how things looked in the admin section (my attention has been focused on there and not the end-user side yet, so I tend to forget to look at the end-user side).

I don't think it would be hard to fix the main menu separate from the topbar, but I also don't really care too much about 35px versus 50px in the topbar. I think 35px looks better but not enough to worry about. We'll probably go to 35px when we customize Hyku for Stanford, just to be consistent with most of our other applications, but that doesn't mean everyone needs 35px.

PR #1082, opened 2017-05-09
Use fits 0.10.2 (for consistency with AWS deploy)

cbeer:


cbeer:
Obsolete by #1083, if that build passes.

PR #1087, opened 2017-05-10
Switch all or bust

atz:
Prevents half-formed tenants from partially or entirely in the global lobby.  

Also refactors `reset!` to be at each Endpoint's descendants' class level, therefore `@account.reset!` is not dependent on account well-formedness, while `switch!` decidedly is.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -77,9 +77,9 @@ class Account < ActiveRecord::Base
 
   # does not use confirm_endpoints! because we can still nil-ify settings
   def reset!
-    solr_endpoint.reset!
-    fcrepo_endpoint.reset!
-    redis_endpoint.reset!
+    SolrEndpoint.reset!
+    FcrepoEndpoint.reset!
+    RedisEndpoint.reset!
   end
 
   private
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -62,9 +62,10 @@ class Account < ActiveRecord::Base
 
   # Make all the account specific connections active
   def switch!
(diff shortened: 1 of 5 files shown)


app/models/account.rb
```diff
@@ -74,14 +75,22 @@ def switch
...
+    SolrEndpoint.reset!
+    FcrepoEndpoint.reset!
+    RedisEndpoint.reset!
   end
 
   private
 
+    # @raise [RuntimeError] if missing any endpoint
+    def confirm_endpoints!
+      raise "Account #{cname} is missing solr_endpoint, cannot switch!" unless solr_endpoint
```
cbeer: If we're going to go through the trouble of validating these, can we at least create our own exception class?

atz: Considered it, but didn't think it would necessarily add anything, since literally nobody should ever be trying to catch this exception.

atz: @cbeer what is the utility of a novel exception class that should never be caught?  If you want one, might'nt you prefer three?

cbeer: I'm just after something a little more specific than a `RuntimeError`. If nothing else, it'll give us a slightly more informative error in logs and honeybadger, and give us something to rescue to present a user-friendly error in the future. I was thinking something generic like `MissingEndpointException`, but if you feel like it's better to have separate ones for every service, 🤷‍♂️ .

mjgiarlo: That sounds like a valuable addition.

atz: OK, will do.

PR #1090, opened 2017-05-10
Update dependencies

cbeer:


config/environments/production.rb
```diff
@@ -58,6 +58,7 @@
   # Use a real queuing backend for Active Job (and separate queues per environment)
   require 'active_job/queue_adapters/better_active_elastic_job_adapter'
   config.active_job.queue_adapter     = Settings.active_job.queue_adapter
+  config.active_elastic_job.secret_key_base = Rails.application.secrets[:secret_key_base]
```
mjgiarlo: Is this already set in our AWS templates? I assume so.

cbeer: Is what? `secret_key_base`? That's a default Rails-ism.

PR #1104, opened 2017-05-11
Enable multi-tenancy in test for real

atz:
What was the baseline of d45c0bcfa13c4ec32b88ecddd379de809148d4b1 got suppressed by rebasing for the benefit of the reader.

See #989.


atz:
The whole branch never made it to master.  Replaced with #1105

PR #1106, opened 2017-05-12
hyrax controllers are missing a breadcrumbs mixin

### LINKED ISSUE #1096: Edit work page is missing breadcrumbs
As a depositor or administrative user, when I am on the Edit Work page, I would expect to see the breadcrumbs at the top of the main page area, indicating that this page falls within "Administration" like other related pages do, ie., "Home / Administration / Works / Edit"

screenshot 2017-04-20 16 46 39

darrenleeweber:
Fixes #1096 

The issue describes the approach taken here.  Basically, it updates the hyrax controllers with a new mixin for the breadcrumbs.

NOTE:  the breadcrumbs for the `FileSetsController` is a GUESS.  I don't know, yet, whether it should be added and, if so, whether it should be a `BreadcrumbsForCollections` mixin.

### DIFF
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -1,7 +1,6 @@
 module Hyrax
   class FileSetsController < ApplicationController
     include Hyrax::FileSetsControllerBehavior
-    include Hyrax::FileSetsControllerBehavior
     self.show_presenter = Hyku::FileSetPresenter
   end
 end
--- a/app/controllers/hyrax/generic_works_controller.rb
+++ b/app/controllers/hyrax/generic_works_controller.rb
@@ -2,6 +2,7 @@ module Hyrax
   class GenericWorksController < ApplicationController
     # Adds Hyrax behaviors to the controller.
     include Hyrax::WorksControllerBehavior
+    include Hyrax::BreadcrumbsForWorks
 
     self.curation_concern_type = GenericWork
 
--- a/app/controllers/hyrax/images_controller.rb
+++ b/app/controllers/hyrax/images_controller.rb
@@ -2,6 +2,7 @@ module Hyrax
   class ImagesController < ApplicationController
     # Adds Hyrax behaviors to the controller.
     include Hyrax::WorksControllerBehavior
+    include Hyrax::BreadcrumbsForWorks
 
     self.curation_concern_type = Image
 


app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Hyrax
   class FileSetsController < ApplicationController
     include Hyrax::FileSetsControllerBehavior
-    include Hyrax::FileSetsControllerBehavior
+    include Hyrax::BreadcrumbsForCollections
```
jcoyne: `BreadcrumbsForCollections` is for the CollectionsController (https://github.com/projecthydra-labs/hyrax/blob/5c2df85da6181a2ad18e3d23bc951699f8c3508f/app/controllers/hyrax/collections_controller.rb#L4) I don't think it's appropriate here.

darrenleeweber: Removed

darrenleeweber:
Assuming all the specs are in the hyrax engine.  Not clear about duplicating engine specs in an app using the engine.  The use of engine generators may need to be reviewed when the engine itself is moving fast and there are close ties of the app on the engine.  Perhaps all the engine generators need to be re-run every time there is an engine update?  See related [WIP] PRs on this issue.

mjgiarlo:
@darrenleeweber :speech_balloon: 
> Assuming all the specs are in the hyrax engine. Not clear about duplicating engine specs in an app using the engine. 

Yeah, Hyrax itself does test the `BreadcrumbsForWorks` module. The spec that'd be helpful to have included in this PR would ensure Hyku is integrating with Hyrax in a way that matches our expectations.

> The use of engine generators may need to be reviewed when the engine itself is moving fast and there are close ties of the app on the engine. Perhaps all the engine generators need to be re-run every time there is an engine update? See related [WIP] PRs on this issue.

We wouldn't want to re-run *all* engine generators, but the work generator is safe-ish to re-run (thanks to git, we can see what changes it makes, and only accept the ones that don't rollback customizations that we need).

FWIW, we used to create version-specific generators that rolled up all the changes since the past version ([see some examples from sufia](https://github.com/projecthydra/sufia/tree/v6.6.0/lib/generators/sufia/)). They were costly to create but arguably quite useful. That's not a model we could use for Hyku now, though, because we're moving so fast -- *i.e.*, our velocity is such that we can't wait on releases with upgrade paths now, which is why we're tracking Hyrax `master`.

Also, I restarted the failing build above.

mjgiarlo:
If @jcoyne, @atz, or @cbeer feel differently about the spec thing, though, I'm happy to let it go.

mjgiarlo:
Meh, I'm shipping it.

mjgiarlo:
Thank you, @darrenleeweber!

PR #1107, opened 2017-05-12
rails generate hyrax:collection

darrenleeweber:
Double checking the latest hyrax generators - hyrax:collection in this PR.


app/models/collection.rb
```diff
@@ -1,6 +1,7 @@
-# Generated by curation_concerns:models:install
+# Generated by hyrax:models
```
jcoyne: Technically this was generated by curation_concerns (before we moved to hyrax). But I don't feel strongly about this.

jcoyne:
@darrenleeweber why is this marked `[WIP]`?  Is there more to come?

darrenleeweber:
I've marked all the `rails generate` PRs as [WIP], pending advice on whether they are good changes or not.  There's nothing more to add here.

jcoyne:
@darrenleeweber there is no need to mark something WIP unless you are planning to add more to it.  If you don't mark it, we will consider it ready for review.

PR #1108, opened 2017-05-12
rails generate hyrax:models

darrenleeweber:
Double checking the latest hyrax generators - hyrax:models in this PR.


jcoyne:
This commit reverts 508e5028f52281a7c6024a5d03e3cb161f43a5c3 is that your intention?

mjgiarlo:
@jcoyne I believe @darrenleeweber was learning about Hyrax's work generator yesterday, and it perhaps isn't obvious that the FileSet and Collection models are not work types generated via the work generator.

darrenleeweber:
I started to look at `rails generate --help` to identify any hyrax generators that might be creating different results from those generated when hyku was first spun up.  The intention is to surface any changes in the results of hyrax:generators (a carry over from identification of the cause of missing breadcrumbs on works).  These PRs just re-run the generators and may surface two kinds of changes: (a) those that are useful and (b) changes to intentional modifications to the generators.  (For my part, I can't tell the difference and I have no intention of wanting to change the intentional-changes.)  We need to review the generator changes - these PRs are a relatively quick way to formally do that.  I suppose line-by-line commentary on the diffs is most effective?

spec/models/file_set_spec.rb
```diff
@@ -1,6 +1,7 @@
+require 'rails_helper'
+
 RSpec.describe FileSet do
-  describe 'indexer' do
-    subject { described_class.indexer }
-    it { is_expected.to eq FileSetIndexer }
+  it "has tests" do
+    skip "Add your tests here"
```
jcoyne: This overwrites a useful test with one that does nothing

app/models/file_set.rb
```diff
@@ -1,5 +1,4 @@
-# Generated by curation_concerns:models:install
+# Generated by hyrax:models:install
 class FileSet < ActiveFedora::Base
   include ::Hyrax::FileSetBehavior
-  self.indexer = FileSetIndexer
 end
```
darrenleeweber: Is this indexer a custom addition? Should it be removed?  If it's removed here, does it get included somewhere else in the stack?

darrenleeweber:
Test failure now is:
```
  1) FileSet indexer should eq FileSetIndexer

     Failure/Error: it { is_expected.to eq FileSetIndexer }

       expected: FileSetIndexer
            got: Hyrax::FileSetIndexer
```
Does HyKu have it's own FileSetIndexer or can this spec check for the actual indexer from Hyrax?

mjgiarlo:
@darrenleeweber Yes, Hyku has its own FileSetIndexer: https://github.com/projecthydra-labs/hyku/blob/master/app/indexers/file_set_indexer.rb#L2

We overrode what Hyrax provides to inject IIIF behavior here.

app/models/file_set.rb
mjgiarlo: Yes, no, no. :)

darrenleeweber:
I _think_ updates for app/models/file_set.rb are done now.

PR #1118, opened 2017-05-13
Account creation job renovation

atz:
- `bundle update hyrax`
- Remove busted after create callback
- Remove Jobs that have no reason to be async
- Put inline stuff inline, not behind unnecessary one-line jobs
- Use new `Hyrax::ApplicationJob` as the injection point for `ActiveJobTenant`, leaving `ActiveJob::Base` available *without* switching
- Update tests to match


hannahfrost:
I was able to create a new tenet!

mjgiarlo:
Didn't work for me. Trying out a small amendment to this patch.

PR #1130, opened 2017-05-16
Update active-elastic-job to 2.0 and fix tenant jobs

### LINKED ISSUE #1129: Upgrade to active-elastic-job 2.0
This will involve several steps:

1. Add the new environment variable `PROCESS_ACTIVE_ELASTIC_JOBS` to https://github.com/hybox/aws/blob/bd95b91c569ccb931abf33a321dc77a04933969e/templates/worker.yaml
1. Update the Gemfile and bundle install
1. Remove the old `DISABLE_SQS_CONSUMER` variable: https://github.com/hybox/aws/blob/bd95b91c569ccb931abf33a321dc77a04933969e/templates/webapp.yaml#L315-L317

cbeer:
fixes #1129

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -27,10 +27,6 @@ module Hyku
       "I18n::InvalidLocale" => :not_found
     )
 
-    # Rails handles ParseError by default, but it is occurring on SQS and I need more details about why,
-    # so turn it off.
-    ActionDispatch::ExceptionWrapper.rescue_responses.delete("ActionDispatch::ParamsParser::ParseError")
-
     if defined? ActiveElasticJob
       Rails.application.configure do
         config.active_elastic_job.process_jobs = Settings.worker == 'true'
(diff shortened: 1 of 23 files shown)


config/application.rb
```diff
@@ -27,9 +27,13 @@ class Application < Rails::Application
       "I18n::InvalidLocale" => :not_found
     )
 
-    # Rails handles ParseError by default, but it is occurring on SQS and I need more details about why,
-    # so turn it off.
-    ActionDispatch::ExceptionWrapper.rescue_responses.delete("ActionDispatch::ParamsParser::ParseError")
+    if defined? ActiveElasticJob
```
cbeer: Should we wrap this in a `config.before_initialize` or something?

jcoyne: http://guides.rubyonrails.org/initialization.html#config-application-rb the application.rb comes before all the other initializers, so I don't think we need to.

PR #1132, opened 2017-05-17
Consolidate set_site to SitesController and inherit downstream

### LINKED ISSUE #960: Banner image can't be removed
on the appearance form the "file" upload field is marked as required, so trying to submit the form without that prevents form submission.  We should mark the "file" file as not required.

atz:
`SitesController` previously had a call that *required* set_site to exist in itself, but it did not have it defined.  This is required to make `SitesController` useful. At the same time, we can utilize inheritance.

Fixes #960

### DIFF
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -6,7 +6,7 @@ class SitesController < ApplicationController
   def update
     params.require(:remove_banner_image)
     @site.remove_banner_image!
-    redirect_to hyrax.admin_appearance_path
+    redirect_to hyrax.admin_appearance_path, notice: 'The appearance was successfully updated.'
   end
 
   private
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -1,5 +1,17 @@
 class SitesController < ApplicationController
   before_action :set_site
-  load_and_authorize_resource
+  load_and_authorize_resource instance_variable: :site, class: 'Site' # descendents still auth Site
   layout 'dashboard'
+
+  def update
+    params.require(:remove_banner_image)
+    @site.remove_banner_image!
+    redirect_to hyrax.admin_appearance_path
+  end
+
+  private
+
+    def set_site
+      @site ||= Site.instance
+    end
 end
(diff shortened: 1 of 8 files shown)


atz:
Implementation looks like:
screen shot 2017-05-17 at 2 31 11 pm

For when you have a Devo banner image.  I will leave it to design to modify the location/position of the remove button should they see fit.

app/controllers/sites_controller.rb
```diff
@@ -1,5 +1,17 @@
 class SitesController < ApplicationController
   before_action :set_site
-  load_and_authorize_resource
+  load_and_authorize_resource instance_variable: :site, class: 'Site' # descendents still auth Site
   layout 'dashboard'
+
+  def update
```
cbeer: It strikes me as a little strange that this base class exposes an `#update` method that will get inherited by all the subclasses. Maybe we should have a separate site banner controller?

mjgiarlo:
@atz When you have time, can you respond to @cbeer's question? Let's resolve his concern before merging.

app/controllers/sites_controller.rb
atz: It is a little strange, but not much weirder than having a model base class wrapping/overriding ActiveRecord methods, or all of ActiveFedora.  The inheriting classes here all override, so as a pragmatic thing, it doesn't matter.  I could imagine taking the base class lower and having the controllers (including this one) inherit it, so that no `#update` was inherited.  But the idea that mis-inheriting will prove weird is a hypothetical.  I am satisfied to defer the refactor until such time as we have another descendant that doesn't override `#update` (doesn't seem likely/imminent).

jcoyne: I'm with @cbeer. This should only be on the controller that deals with banner image.  Let's make this code easy to understand and avoid creating troublesome issue like you mention.

atz: Yeah, let's hold up my actual improvement for your hypothetical problem... again.  Not feeling it.  You can file an issue to refactor and we'll let it rot on the backlog like we do with mine.

Issue #1158, opened 2017-05-26
Docker workers won't start


jcoyne:
I see this message repeated over and over: 

```
workers_1         | 2017-05-26T17:57:52.273Z 1 TID-gn564p2y0 INFO: Redis is online, 33.442146484 sec downtime
workers_1         | 2017-05-26T17:57:53.241Z 1 TID-gn56r70pg ERROR: Error fetching job: Waited 5 sec
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool/timed_stack.rb:86:in `block (2 levels) in pop'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool/timed_stack.rb:78:in `loop'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool/timed_stack.rb:78:in `block in pop'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool/timed_stack.rb:77:in `synchronize'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool/timed_stack.rb:77:in `pop'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool.rb:89:in `checkout'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool.rb:61:in `block in with'
workers_1         | 2017-05-26T17:57:53.242Z 1 TID-gn56r70pg ERROR: /usr/local/bundle/gems/connection_pool-2.2.1/lib/connection_pool.rb:60:in `handle_interrupt'
[...]

PR #1160, opened 2017-05-26
Use Sidekiq by default in dev, and pull value from Settings

mjgiarlo:
Refs #1158 

Shot in the dark.


mjgiarlo:
Nvm. Barking up the wrong tree.

Issue #1193, opened 2017-06-02
Problem on generate thumbnails for PDFs or PPTXs


haozeng0:
Freshly installed in docker, couldn't generate thumbnails for PDF or PPTX files. Images work fine.

Here is the hyku_workers log:

 I, [2017-06-02T14:54:16.082386 #1]  INFO -- : [ActiveJob] [CreateDerivativesJob] [08e84faf-a3dd-47f3-baf5-4b372c32369e] Performing CreateDerivativesJob (Job ID: 08e84faf-a3dd-47f3-baf5-4b372c32369e) from Sidekiq(default) with arguments: #<GlobalID:0x0055b566519ea8 @uri=#<URI::GID gid://hyku/FileSet/afe7f62f-e740-4691-99ef-4adf1c9779c9>>, "afe7f62f-e740-4691-99ef-4adf1c9779c9/files/bd9db74f-5d75-4a1e-a8b6-d6707b86d44a", "/data/tmp/uploads/e9753484-1371-4e1f-aa73-c51045c989d1/hyrax/uploaded_file/file/3/EDS_VuFind_Integration_Filipe_Bento_EBSCO_Sept2015.pptx"
 **E, [2017-06-02T14:54:18.644273 #1] ERROR -- : [ActiveJob] [CreateDerivativesJob] [08e84faf-a3dd-47f3-baf5-4b372c32369e] Error performing CreateDerivativesJob (Job ID: 08e84faf-a3dd-47f3-baf5-4b372c32369e) from Sidekiq(default) in 2561.49ms: MiniMagick::Invalid (`identify /tmp/mini_magick20170602-1-1rpog0s.pdf` failed with error:
 ):**
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:203:in `rescue in validate!'
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:201:in `validate!'
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:113:in `block in create'
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:112:in `tap'
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:112:in `create'
 /usr/local/bundle/gems/mini_magick-4.7.0/lib/mini_magick/image.rb:34:in `read'
[...]

PR #1194, opened 2017-06-02
Add ghostscript to docker build for PDF/etc. derivatives

mjgiarlo:
Refs #1193


mjgiarlo:
I can confirm that PDF derivatives now work, so I think this is good evidence that adding ghostscript to the build had the desired effect.

mjgiarlo:
@jcoyne can you re-review/merge given :point_up: ?

PR #1240, opened 2017-06-14
Update Hyrax and other dependencies

mjgiarlo:
Update translations now that Hyrax supports French, Italian, German, and Portuguese.


mjgiarlo:
@darrenleeweber Done. Re-review?

PR #1247, opened 2017-06-15
Add a rake task to create a superadmin

### LINKED ISSUE #1242: Rake task to create global (non-tenant) user
The current method for creating a super-admin user to manage tenants, requires signing up for a global user account first: https://github.com/samvera-labs/hyku/wiki/Create-super-admin-user.

Now that https://github.com/samvera-labs/hyku/issues/1203 is complete, disabling open sign-up on stack build also removes the ability to sign up as a global user. There needs to be another method for creating these global user accounts. The simplest option is likely a rake task, but that may need discussion.

jcoyne:
Fixes #1242

### DIFF
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -1,0 +1,38 @@
+namespace :hyku do
+  namespace :superadmin do
+    desc 'Create a superadmin user'
+    task create: :environment do
+      puts 'Creating a superadmin user.'
+      user = prompt_to_create_user
+      user.add_role(:superadmin)
+      puts "User (#{user.user_key}) created and granted superadministrator privileges"
+    end
+
+    def prompt_to_create_user
+      User.find_or_create_by!(email: prompt_for_email) do |u|
+        puts 'User not found. Enter a password to create the user.'
+        u.password = prompt_for_password
+      end
+    rescue => e
+      puts e
+      retry
+    end
+
+    def prompt_for_email
+      print 'Email: '
+      $stdin.gets.chomp
+    end
+
+    def prompt_for_password
+      begin
+        system 'stty -echo'
+        print 'Password (must be 8+ characters): '
+        password = $stdin.gets.chomp
+        puts "\n"
+      ensure
+        system 'stty echo'
+      end
+      password
+    end
+  end
+end
--- a/spec/tasks/rake_spec.rb
+++ b/spec/tasks/rake_spec.rb
@@ -12,21 +12,21 @@ RSpec.describe "Rake tasks" do
     end
 
     it 'requires user_list argument' do
-      expect { run_task('superadmin:grant') }.to raise_error(ArgumentError)
+      expect { run_task('hyku:superadmin:grant') }.to raise_error(ArgumentError)
     end
 
     it 'warns when a user is not found' do
-      expect(run_task('superadmin:grant', 'missing@example.org')).to include 'Could not find user'
+      expect(run_task('hyku:superadmin:grant', 'missing@example.org')).to include 'Could not find user'
     end
 
     it 'grants a single user the superadmin role' do
-      run_task('superadmin:grant', user1.email)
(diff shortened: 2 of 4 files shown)


spec/tasks/rake_spec.rb
```diff
@@ -12,21 +12,21 @@
     end
 
     it 'requires user_list argument' do
-      expect { run_task('superadmin:grant') }.to raise_error(ArgumentError)
+      expect { run_task('hyrax:superadmin:grant') }.to raise_error(ArgumentError)
```
mjgiarlo: Shouldn't these all be `hyku:`?

lib/tasks/user.rake
```diff
@@ -0,0 +1,38 @@
+namespace :hyku do
+  namespace :superadmin do
+    desc 'Create an initial superadmin user'
```
mjgiarlo: `s/an initial/a/`?

lib/tasks/user.rake
```diff
@@ -0,0 +1,38 @@
...
+    task create: :environment do
+      puts 'Creating an initial superadmin user.'
```
mjgiarlo: `s/an initial/a/`?

lib/tasks/user.rake
```diff
@@ -0,0 +1,38 @@
...
+      user = prompt_to_create_user
+      user.add_role(:superadmin)
+      puts 'User created.'
+    end
+
+    def prompt_to_create_user
```
mjgiarlo: So, am I reading this PR right that we now have two rake tasks that can elevate an existing user to superadmin? I'm surprised the new task works for existing users given we already have the above rake task for that. Maybe I'm misreading this.

lib/tasks/user.rake
```diff
@@ -0,0 +1,38 @@
+namespace :hyku do
+  namespace :superadmin do
+    desc 'Create an initial superadmin user'
+    task create: :environment do
+      puts 'Creating an initial superadmin user.'
+      user = prompt_to_create_user
+      user.add_role(:superadmin)
+      puts 'User created.'
```
mjgiarlo: How about `puts "User (#{user.user_key}) created and granted superadministrator privileges"`?

spec/tasks/rake_spec.rb
```diff
@@ -12,21 +12,21 @@
     end
```
mjgiarlo: Should test coverage be added here for the new `hyku:superadmin:create` task?

lib/tasks/user.rake
jcoyne: No, this makes a new user and elevates them. The other elevates an existing user.

mjgiarlo: @jcoyne If it only creates new users, do we need `#find_or_create_by` on line 12? The `find` there suggests that this task can act on existing users.

jcoyne: Let me rephrase, this elevates a user.  If they don't exist it creates them.  I believe this is preferable to: "Nope, that user exists. Go find a different rake task to do what you want"

mjgiarlo: @jcoyne Sounds good. Should we retire the other rake task as part of this branch?

jcoyne: @mjgiarlo your call. That one takes a list of users and elevates, this does one at a time but can create them & set passwords too.

spec/tasks/rake_spec.rb
jcoyne: Honestly, I don't know how to test stuff that's reading from `$stdin`.  This code is copied from https://github.com/projectblacklight/spotlight/blob/master/lib/tasks/spotlight_tasks.rake#L2 and it wasn't tested their either.

lib/tasks/user.rake
mjgiarlo: Since you're adding functionality, my preference is to go with this new rake task for both use cases.

PR #1249, opened 2017-06-15
Move Solr container volume to where Solr data lives.

### LINKED ISSUE #1246: Solr cores are lost when Docker container is restarted.
### Descriptive summary

The Solr docker container saves index data at `/opt/solr/server/solr` but this location is not included in a volume in `docker-compose.yml`. Moving the index files to `/data` (which is the current solr volume) is difficult because docker volumes are mounted as root, but Solr does not run as root and so cannot write to that volume.

### Expected behavior
No Solr index data lost after container restarts.
### Actual behavior
No Solr cores present after container restart.

1. Create tenant.
1. `docker-compose stop solr`
1. `docker-compose start solr`
1. Tenant shows "Sorry, I don't understand your search." message.

dlpierce:
Fixes #1246 
Moves mount point of the Solr Docker container's solr volume.

The Solr container saves its index data at `/opt/solr/server/solr`, instead of `/data` where the solr volume is mounted currently. Moving the index data to `/data` is not feasible because of volume permission issues, so keeping the data at its default location (and persisting it) is the simplest fix.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -79,7 +79,7 @@ services:
       - 8983
     volumes:
       - .:/app
-      - solr:/data
+      - solr:/opt/solr/server/solr
     networks:
       internal:
 


qjhart:
As a point of info, previously we had trouble with the docker-compose file for solr.  we ended up with the following solr description.  

``` sh
  solr:
    image: solr
    ports:
      - "${SOLR_PORT}:8983"
    volumes:
      - ../..:/app
      - solr:/opt/solr/server/solr/mycores
    entrypoint:
      - docker-entrypoint.sh
      - solr-precreate
      - hydra_production
      - /app/solr/config
```

mjgiarlo:
@jcoyne Shall we? :point_up:

PR #1263, opened 2017-06-21
Merge volume directives

jcoyne:
Previously line 185 was overwriting line 180.
The volume previously on line 180 is critical for running on
docker-machine, but was not needed on Docker CE.
See https://hub.docker.com/r/dockercloud/haproxy/ "Running with Docker
Compose v2 (Compose Mode)"

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -177,13 +177,12 @@ services:
       - DOCKER_TLS_VERIFY
       - DOCKER_HOST
       - DOCKER_CERT_PATH
-    volumes:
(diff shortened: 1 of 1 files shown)


jcoyne:
This causes problems with the CE Install:

```
ERROR: for lb  Cannot create container for service lb: invalid volume spec ".": invalid volume specification: '.': invalid mount config for type "volume": invalid mount path: '.' mount path must be absolute
ERROR: Encountered errors while bringing up the project.
```

PR #1280, opened 2017-06-23
Only make iiif thumbnails for images.

jcoyne:
Standard thumbnails for PDFs. This was needed because RIIIF doesn't
flatten pdfs, making them look all black. Similar to:
https://github.com/samvera/hydra-derivatives/issues/110


mjgiarlo:
I believe @dlpierce was headed in the same direction. Pinging him here so he can see this work, in hopes that he might review.

PR #1297, opened 2017-06-26
Wrap viewer so height can be set without using viewer class.

### LINKED ISSUE #1251: Universal Viewer in full screen mode does not use the full screen
Issue:

When a user selects the “Full Screen” option in the UV, the viewer renders on only part of the screen (as seen in the attached screenshot). This issue appears on a Mac in both Chrome and Safari. The UV does work as expected in full screen mode in Firefox. Example URL: http://stlawu.pilot.hykudirect.org/concern/images/1f6af6ca-21c6-414a-ae67-6f6a46863011?locale=en

Analysis:

After a little checking around, it appears that the styling for height of 500px is still applied in full screen mode, which then gets centered vertically, and it overflows down from there.

Background Specs:

OS: macOS Sierra Version 10.12.5 
Hardware: Mac Pro 
Browsers:

* Chrome: Version 58.0.3029.110 (64-bit) 
* Safari: Version 10.1.1 (12603.2.4) 
* Firefox: 52.2.0 (64-bit)

This issue was reported by Eric Williams-Bergen
of St. Lawrence University

---

I can verify that this issue also presents in Chrome on Windows 10.

The unnecessary header is created by the CSS element:

.viewer {height: 500px}

acting on the div with class="uv viewer"

andjsmit:
Fixes #1251

Adds a wrapper class around viewer div so height can be set without using viewer class.

### DIFF
--- a/app/assets/stylesheets/viewer.scss
+++ b/app/assets/stylesheets/viewer.scss
@@ -1,5 +1,9 @@
-.viewer {
+.viewer-wrapper {
   height: 500px;
+}
+
+.viewer {
+  height: 100%;
   padding: 10px;
   iframe {
     // position: absolute;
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Thank you, @andjsmit!

PR #1302, opened 2017-06-27
Do not assume that everything has a thumbnail available.

dlpierce:
Fixes samvera/hyrax#1095

Do not assume all files have thumbnails available.

IIIFThumbnailPaths assumes that a thumbnail derivative was created for all filesets, however derivative creation depends on the file file mime type being included in the list of supported mime types. Falling back on Hyrax in this case should display the default work icon in place of a thumbnail.

* Use behavior from hyrax when `thumbnail?` is called on a fileset with an unsupported mime type.

### DIFF
--- a/app/services/concerns/iiif_thumbnail_paths.rb
+++ b/app/services/concerns/iiif_thumbnail_paths.rb
@@ -22,8 +22,13 @@ module IIIFThumbnailPaths
       )
     end
 
-    def thumbnail?(_thumbnail)
-      true
+    # @param [FileSet] thumbnail the object that is the thumbnail
+    # @return [boolean] true when a thumbnail (either generated or a common asset)
+    #                   is expected to be available on the file system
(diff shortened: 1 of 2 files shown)


app/services/concerns/iiif_thumbnail_paths.rb
```diff
@@ -22,8 +22,10 @@ def iiif_thumbnail_path(file_set, size)
       )
     end
 
-    def thumbnail?(_thumbnail)
-      true
+    def thumbnail?(thumbnail)
```
jcoyne: Please use Yard to add method documentation especially about the type of `thumbnail`.

mjgiarlo:
@dlpierce might you have cycles to come back to this? Thanks!

orangewolf:
@jcoyne this looks complete, can you confirm?

PR #1326, opened 2017-07-06
Fix for #1171 - only include images in the IIIF Manifest

### LINKED ISSUE #1171: Universal Viewer should show only things that it can render
### Descriptive summary

If there is a mix of content attached to a work, including images, Universal Viewer attempts to show them all in the viewer. Those is cannot render display as an endless spinning grey square.

### Expected behavior

Only show things that Universal Viewer can render

### Actual behavior

Viewer spins on things it cannot render.

1. Upload several items to a Work including at least one image
2. View the work

### Related work

#1170 related issue on display of access restricted items in Universal Viewer

geekscruff:
Fixes #1171 

Only include images in the IIIF Manifest

At present, all FileSets are included in the IIIF Manifest. This creates RIIIF URLs that won't resolve when the files to which they refer are not images, and affects universal viewer as it can't render the content. The simple fix is to only include FileSets that contain image files in the manifest.

*  Only return the an IIIFManifest::DisplayImage object when calling display_image where solr_document responds 'true' to solr_document.image?

### DIFF
--- a/app/presenters/concerns/displays_image.rb
+++ b/app/presenters/concerns/displays_image.rb
@@ -3,8 +3,11 @@
 module DisplaysImage
   extend ActiveSupport::Concern
 
+  # Creates a display image only where FileSet is an image.
+  #
+  # @return [IIIFManifest::DisplayImage] the display image required by the manifest builder.
   def display_image
-    return nil unless FileSet.exists?(id)
+    return nil unless FileSet.exists?(id) && solr_document.image?
     # TODO: this is slow, find a better way (perhaps index iiif url):
     original_file = FileSet.find(id).original_file
 
--- a/spec/presenters/hyku/file_set_presenter_spec.rb
+++ b/spec/presenters/hyku/file_set_presenter_spec.rb
@@ -5,22 +5,35 @@ RSpec.describe Hyku::FileSetPresenter do
   let(:request) { double(base_url: 'http://test.host') }
   let(:presenter) { described_class.new(solr_document, nil, request) }
   let(:id) { CGI.escape(file_set.original_file.id) }
-  before do
-    Hydra::Works::AddFileToFileSet.call(file_set,
-                                        fixture_file('images/world.png'),
-                                        :original_file)
-  end
 
   describe "display_image" do
-    subject { presenter.display_image }
-    it "creates a display image" do
-      expect(subject).to be_instance_of IIIFManifest::DisplayImage
-      expect(subject.url).to eq "http://test.host/images/#{id}/full/600,/0/default.jpg"
+    context "when the file is not an image" do
+      subject { presenter.display_image }
+      before do
+        Hydra::Works::AddFileToFileSet.call(file_set,
+                                            fixture_file('csv/sample.csv'), :original_file)
+      end
+      it { is_expected.to be_nil }
+    end
+    context "when the file is an image" do
+      subject { presenter.display_image }
+      before do
+        Hydra::Works::AddFileToFileSet.call(file_set,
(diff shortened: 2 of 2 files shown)


app/presenters/concerns/displays_image.rb
```diff
@@ -3,8 +3,11 @@
 module DisplaysImage
   extend ActiveSupport::Concern
 
+  # Creates a display image only where FileSet is an image.
+  #
+  # @return [IIIFManifest::DisplayImage] the display image required by the manifest builder.
   def display_image
-    return nil unless FileSet.exists?(id)
+    return nil unless FileSet.exists?(id) && FileSet.find(id).image?
```
mjgiarlo: Calling `#find` here will hit Fedora and be expensive.

geekscruff: Ah, I wondered about that - it was just so neat this way! I can do something with solr. Will work on it once back from hols next week.

mjgiarlo: OK, great!

geekscruff: switched to solr_document.image? :)

spec/presenters/hyku/file_set_presenter_spec.rb
```diff
@@ -5,10 +5,17 @@
   let(:request) { double(base_url: 'http://test.host') }
   let(:presenter) { described_class.new(solr_document, nil, request) }
   let(:id) { CGI.escape(file_set.original_file.id) }
+
+  let(:file_set_not_image) { FactoryGirl.create(:file_set) }
+  let(:solr_document_not_image) { SolrDocument.new(file_set_not_image.to_solr) }
+  let(:presenter_not_image) { described_class.new(solr_document_not_image, nil, request) }
```
jcoyne: Can this just be the presenter in a context that is about "a file that is not an image"? For example:
```ruby
describe "#display_image" do
  subject { presenter.display_image }
  context "when the file is not an image" do
    before do
       Hydra::Works::AddFileToFileSet.call(file_set,
          fixture_file('csv/sample.csv'), :original_file)
     end
     it { is_expected.to be nil }
  end

  context "when the file is an image" do
    before do
       Hydra::Works::AddFileToFileSet.call(file_set,
          fixture_file('images/world.png'), :original_file)
     end
    ...
  end
end
```

this is preferable in two regards:
1. it's faster because we don't have to set up two sets of objects for each test.
1. we don't need to keep track of a second definition of presenter and solr_document.

geekscruff: Yes, that's a much better way, thanks. Have updated.

spec/presenters/hyku/file_set_presenter_spec.rb
```diff
@@ -5,22 +5,33 @@
...
+      end
+      it { is_expected.to be_nil }
+    end
+    context "when the file is an image" do
+      subject { presenter.display_image }
+      before do
+        Hydra::Works::AddFileToFileSet.call(file_set,
+                                            fixture_file('images/world.png'), :original_file)
+      end
+      it { is_expected.to be_instance_of IIIFManifest::DisplayImage }
```
jcoyne: Would you mind combining these two back to a single it block?  This way is easier to read, but the combined way is a bit faster and we are concerned about how long the build takes to run.

geekscruff: Done :)

geekscruff:
Any idea on the failures I'm seeing on UploadedFiles? I see comments in the uploaded_file_spec about this but I don't know why this would fail in travis for me when I haven't changed anything locally?

PR #1373, opened 2017-07-14
Fix for #1171 - only include images in the IIIF Manifest

geekscruff:
See PR #1326 ... trying a new PR as I think my failures are due to the original PR coming from a fork rather than a branch.


geekscruff:
Finally Green! Can I get a merge? I think I addressed the changes.

Issue #1372, opened 2017-07-14
Support downloads (eg. a PDF of the whole paged item) via Universal Viewer by including renderings in the IIIF Manifest


geekscruff:
### Descriptive summary

I would like to add support for downloading files via Universal Viewer. This will support the use case of a paged item where there are images of each page to display in the viewer and a PDF download of the whole to be offered as a single download.

This is already supported in Universal Viewer and would display  like so:

![iiif_download](https://user-images.githubusercontent.com/722117/28206519-219a9458-687e-11e7-96db-c4954bb05a95.PNG)

In IIIF there are various linking properties. Providing a display/download URL is achieved with a [rendering linking property](http://iiif.io/api/presentation/2.1/#linking-properties), like so:

```
"rendering": {
    "@id": "http://example.org/iiif/book1.pdf",
    "label": "Download as PDF",
    "format": "application/pdf"
  }
```

### Proposed approach:

In Hyku:

* Add a new property 'rendering' mirroring the code for representative. This property would only be shown for items responding true to `image?`. This property would be used to select one or more FileSets to be used as renderings for the Work. 
* In the ManifestEnabledWorkShowPresenter add a 'rendering' method to build a hash as per the above fragment.

There's a bit more detail in this [gist](https://gist.github.com/geekscruff/4c3c89a674300eee8983381db18b0f26)

In IIIF_Manifest:

* Add code to process the rendering method and add the rendering to the sequence.

I would like feedback on the idea and proposed approach @samvera/hyrax-code-reviewers

mjgiarlo:
I have to defer to @jcoyne, who has done the most work in and around this part of Hyku. (Justin is working on another sprint this week, so hang in there.)

I would love to support this!

geekscruff:
@mjgiarlo Thanks! I'll probably do a bit of prototyping over in my fork. Even if it all changes, it's good Hyku/Hyrax learning stuff.

mjgiarlo:
@geekscruff So @jcoyne is away for a number of weeks at this point, thus I'd encourage you to move forward with this work and pls do feel free to share PRs early and often over here! Happy to chime in and test as you go.

PR #1376, opened 2017-07-21
Implementation of #1372: support downloads (eg. a PDF of the whole paged item) via Universal Viewer

geekscruff:
Fix for #1372

Support downloads (eg. a PDF of the whole paged item) via Universal Viewer by including rendering in the IIIF Manifest

Adds a rendering property. Once one or more 'renderings' (file_sets) have been selected for the work and saved, the IIIF manifest will then contain a rendering element in the sequences block. This will be picked up by Universal Viewer and the specified files listed for download.

capture

### Main work done:

* Added a new HABM property 'renderings' onto GenericWork and Image via a concern: hasRendering
* Following the code for representative (with slight differences because that is a singular belongs_to and this is a HABM), added rendering_ids into the image and generic work forms and solr_document, file_set and file_set indexer
* Added a view form in app/views/hyrax/base and included the form in _form_media; configured to include the rendering_ids field in the edit form only when the object already exists AND when it will use the universal viewer (ie. when the representative FileSet responds true to 'image?')
* Add a method called 'manifest_extras' into the manifest_enabled_work_show_presenter - this method generates a hash containing :sequence_rendering, :sequence_rendering contains an array of hashes, one for each selected 'rendering'. 

When the download appears in Universal Viewer, I've configured the label to be: "Download whole resource: label from file_set".

download

### IIIF_Manifest Gem

In order to work fully, the current PR depends on changes in iiif_manifest, to consume the new method and populate the manifest with the rendering block. The changes to iiif_manifest have been made and accepted into iiif_manifest, although a new gem release is pending. 

PR here: https://github.com/samvera-labs/iiif_manifest/pull/3

### Tests

I've added tests for the forms and presenters, and one for the view.

### File Manager

I haven't incorporated this new field into the file manager. It would make sense to allow the selection of renderings in the file manager but it would involve pulling quite a bit up from hyrax into hyku and I was a little concerned about the complexity of doing so. Thoughts?


mjgiarlo:
I have to defer to @jcoyne here, when he has cycles for Hyku-ish things. Thanks for taking on this work, @geekscruff!

app/views/hyrax/base/_form_media.html.erb
```diff
@@ -0,0 +1,8 @@
+<% if f.object.persisted? && f.object.member_ids.present? %>
+  <%= render 'form_representative', f: f %>
+  <%= render 'form_thumbnail', f: f %>
+  <%# display the rendering dropdown only if the representitive responds to image? (ie. Universal Viewer is used) %>
+  <%if curation_concern.representative.image? %>
```
jcoyne: add a space here.

spec/factories/generic_works.rb
```diff
@@ -8,7 +8,10 @@
 
     factory :work_with_one_file do
       before(:create) do |work, evaluator|
-        work.ordered_members << FactoryGirl.create(:file_set, user: evaluator.user, title: ['A Contained Generic File'])
+        work.ordered_members << FactoryGirl.create(:file_set,
+                                                   user: evaluator.user,
+                                                   title: ['A Contained Generic File'],
+                                                   label: 'test label')
```
jcoyne: Why is this being added?

app/views/hyrax/base/_form_rendering.html.erb
```diff
@@ -0,0 +1,15 @@
+  <div class="row">
+    <div class="col-md-12">
+      <fieldset id="representative-media">
+        <legend>
+          Rendering
+        </legend>
+        <p>Select file(s) to be offered as as downloads for every image in Universal Viewer, for example a PDF of the whole <%= f.object.human_readable_type %>.</p>
+        <%= f.select :rendering_ids,
+                     @form.select_files,
+                     {include_blank: true},
```
jcoyne: add a space inside the braces.

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
```diff
@@ -25,4 +25,25 @@
...
     end
   end
+
+  describe "manifest_extras" do
+    subject do
+      presenter.manifest_extras
+    end
+
+    let(:work) { FactoryGirl.create(:work_with_one_file) }
+    let(:document) { work.to_solr }
```
jcoyne: Ideally, we'd do this once ahead of time and avoid having to persist this work just so we can generate a solr document.  This helps keep the test suite fast.

spec/views/hyrax/base/_form_rendering.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  let(:page) do
+    view.simple_form_for form do |f|
+      render 'hyrax/base/form_rendering', f: f
+    end
+    Capybara::Node::Simple.new(rendered)
+  end
+
+  # TODO: get help! (undefined method `select_files' for nil:NilClass)
```
jcoyne: What sort of help do you need?

app/views/hyrax/base/_form_rendering.html.erb
```diff
@@ -0,0 +1,15 @@
+  <div class="row">
+    <div class="col-md-12">
+      <fieldset id="representative-media">
+        <legend>
+          Rendering
+        </legend>
+        <p>Select file(s) to be offered as as downloads for every image in Universal Viewer, for example a PDF of the whole <%= f.object.human_readable_type %>.</p>
+        <%= f.select :rendering_ids,
+                     @form.select_files,
```
jcoyne: Use `f.object` rather than `@form`. I think that'll fix your `TODO` in the test down below.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -2,16 +2,46 @@ module Hyku
...
+        end
+        renderings.flatten
+      end
+
+      def rendering_helper(id)
+        ActiveFedora::SolrService.query("id:#{id}",
+                                        fl: [ActiveFedora.id_field, 'label_ssi', 'mime_type_ssi'],
+                                        rows: 1).map do |x|
+          { '@id' => request.base_url +
+            Hyrax::Engine.routes.url_helpers.download_path(x.fetch(ActiveFedora.id_field)),
```
jcoyne: switch this to download_url and pass the `host` option.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -2,16 +2,46 @@ module Hyku
...
+            'format' => x.fetch('mime_type_ssi'),
+            'label' => "Download whole resource: #{x.fetch('label_ssi')}" }
```
jcoyne: are we okay with a KeyError here if these keys aren't found?

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -2,16 +2,46 @@ module Hyku
...
+        renderings = []
+        if solr_document.rendering_ids.present?
+          solr_document.rendering_ids.map do |id|
+            renderings << rendering_helper(id)
+          end
+        end
+        renderings.flatten
+      end
+
+      def rendering_helper(id)
```
jcoyne: Add yard doc to this method please.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -2,16 +2,46 @@ module Hyku
...
+        ActiveFedora::SolrService.query("id:#{id}",
```
jcoyne: extract this query to a method.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -2,16 +2,46 @@ module Hyku
...
     Hyrax::MemberPresenterFactory.file_presenter_class = Hyku::FileSetPresenter
 
-    delegate :extent, to: :solr_document
+    delegate :extent, :rendering_ids, to: :solr_document
 
     def manifest_url
       manifest_helper.polymorphic_url([:manifest, self])
     end
 
+    def manifest_extras
```
jcoyne: add yard doc here

app/presenters/hyku/file_set_presenter.rb
```diff
@@ -1,5 +1,7 @@
 module Hyku
   class FileSetPresenter < Hyrax::FileSetPresenter
     include DisplaysImage
+    # CurationConcern methods
+    delegate :rendering_ids, :to_s, to: :solr_document
```
jcoyne: Doesn't Hyrax::FileSetPresenter already delegate `:to_s`?

app/models/file_set.rb
```diff
@@ -2,7 +2,26 @@
...
   # It overrides Hyrax to inject IIIF behavior.
   self.indexer = FileSetIndexer
+
+  def rendering_ids
+    to_param
+  end
+
+  # If any parent objects are pointing at this object as their
+  # rendering, remove that pointer.
+  def remove_rendering_relationship
```
jcoyne: Should this be a private method?

app/models/concerns/has_rendering.rb
```diff
@@ -0,0 +1,11 @@
+module HasRendering
+  extend ActiveSupport::Concern
+
+  included do
+    # rubocop:disable Rails/HasAndBelongsToMany
+    has_and_belongs_to_many :renderings,
+                            predicate: ::RDF::URI.new("http://london.ac.uk/ontologies/terms#hasRendering"),
```
jcoyne: I'd like the opinion of other as to the best predicate for this.

app/indexers/file_set_indexer.rb
```diff
@@ -1,3 +1,8 @@
 class FileSetIndexer < Hyrax::FileSetIndexer
   self.thumbnail_path_service = IIIFWorkThumbnailPathService
+  def generate_solr_document
+    super.tap do |solr_doc|
+      solr_doc['hasRendering_ssim'] = object.rendering_ids
```
jcoyne: Can we do this using a snake-case? That's the convention we're using almost everywhere else.

Gemfile
```diff
@@ -102,7 +102,8 @@ gem 'zk'
 gem 'mods', '~> 2.1'
 gem 'riiif', '~> 1.1'
 
-gem 'iiif_manifest', '~> 0.2.0'
+# gem 'iiif_manifest', '~> 0.2.0'
+gem 'iiif_manifest', git: 'https://github.com/ULCC/iiif_manifest', branch: 'hyku_1372_rendering'
```
jcoyne: Is this work going to be merged upstream?

geekscruff: Yes, this PR is blocked until that happens. Just waiting for feedback on https://github.com/samvera-labs/iiif_manifest/pull/3

app/indexers/file_set_indexer.rb
geekscruff: It's a HABM so I think the convention of camel case is correct for this case?

app/views/hyrax/base/_form_media.html.erb
geekscruff: done :)

spec/factories/generic_works.rb
geekscruff: I've sorted out the original problem, this is now gone.

app/views/hyrax/base/_form_rendering.html.erb
geekscruff: done :)

app/views/hyrax/base/_form_rendering.html.erb
geekscruff: done :)

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
geekscruff: done :)

spec/views/hyrax/base/_form_rendering.html.erb_spec.rb
geekscruff: fixed via comment above

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: done :)

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: done :)

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: done :)

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: done :)

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: I rewrote this to avoid errors.

app/presenters/hyku/file_set_presenter.rb
geekscruff: fixed

app/models/file_set.rb
geekscruff: fixed

app/models/concerns/has_rendering.rb
geekscruff: Changed via feedback from IIIF community.

geekscruff:
@jcoyne @mjgiarlo ... if you are happy with the code in this PR, I think there are two final things for me to do:

1) cut a new version of iiif_manifest
2) add another issue to extend the implementation to the file manager - I think that would need more discussion as it would mean bringing up quite a bit of Hyrax code to Hyku

I'd quite like to do (1) as I haven't released a gem before, but I'd need some instruction.

mjgiarlo:
I'm a bit out of my depth on this PR, so I need to defer to @jcoyne.

Thanks for working on this, @geekscruff!

geekscruff:
Any chance of a review on this PR @jcoyne :)

jcoyne:
@geekscruff looks good. Can you squash the first two commits and get rid of the "first pass" language?  It appears that the second commit is just reverting some changes in the first commit.

 Instead of saying "1372" can you say what it does in the commit message so that I don't have to reference the issue to understand why you made these changes?

geekscruff:
@jcoyne how's that? ... also see my comment on Aug 10th re cutting a new version of iiif_manifest

mjgiarlo:
Thank you, @geekscruff!

PR #1391, opened 2017-08-04
Add a rake task that runs other tasks on all or selected tenants

### LINKED ISSUE #1357: Write a new rake task that can run non-Hyku-based rake tasks across all or select tenants

mjgiarlo:
Fixes #1357

To run the `rake user:list[email,foobar]` task within the context of all tenants, invoke `rake tenantize:task[user:list,email,foobar]`. To run the task only in specified tenants, use an environment variable named `tenants`, with each tenant cname separated by a space: `rake tenantize:task[user:list,email,foobar] tenants="foo.localhost baz.localhost quuuuux.localhost"`

### DIFF
--- a/spec/support/rake.rb
+++ b/spec/support/rake.rb
@@ -1,16 +1,11 @@
 require 'rake'
 
 module RakeHelper
-  def load_rake_environment(files)
-    @rake = Rake::Application.new
-    Rake.application = @rake
+  def run_task(task_name, *args)
     Rake::Task.define_task(:environment)
-    files.each { |file| load file }
-  end
-
-  def run_task(task, *args)
     capture_stdout_stderr do
-      @rake[task].invoke(*args)
+      Rake.application[task_name].reenable
+      Rake.application[task_name].invoke(*args)
     end
   end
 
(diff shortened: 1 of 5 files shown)


atz:
Seems like the kind of thing that could use some tests?

spec/support/rake.rb
```diff
@@ -1,16 +1,11 @@
 require 'rake'
 
 module RakeHelper
-  def load_rake_environment(files)
```
mjgiarlo: The spec support for rake tasks was showing its age and complicating my ability to test *my* rake task (which calls other rake tasks) independent of testing the code in this helper. We now have simpler code, and less code, in this module.

spec/tasks/rake_spec.rb
```diff
@@ -1,12 +1,15 @@
 require 'rake'
 
 RSpec.describe "Rake tasks" do
+  before(:all) do
+    Rails.application.load_tasks
```
mjgiarlo: Using this idiomatic call in favor of our custom `load_rake_environment` method, which I have excised above.

mjgiarlo:
@atz It's got tests now. Thanks! Care to review?

mjgiarlo:
See above, @bbranan!

PR #1436, opened 2017-09-22
Add metadata into IIIF Manifest

geekscruff:
The IIIF Presentation API supports adding arbitrary metadata within a `metadata` array.

This small change in Hyku add a manifest_metadata method into ManifestEnabledWorkShowPresenter to support adding metadata into the IIIF manifest. It takes the `required_fields` for the object and returns a metadata array. This method is then consumed by the `iiif_manifest` gem (via this [PR](https://github.com/samvera-labs/iiif_manifest/pull/9)) and the metadata is added to the manifest. In Universal Viewer, the metadata will be displayed in the **_More Information_** tab to the right.

image

The mechanism for getting the required fields is a bit brittle and wouldn't work with works generated with a namespace. Having said that, it does not error, it just means metadata won't be retrieved by the `manifest_metadata` method if the Form class is not in the format `Hyrax::MyModelForm`.

In addition, there is a small change to `sequence_rendering` to simplify the implementation and align it with the current implementation of `iiif_manifest`.

`Gemfile` and `Gemfile.lock` updated to use `iiif_manifest` 0.3.0 which includes support for these new features.

### DIFF
--- a/app/presenters/hyku/manifest_enabled_work_show_presenter.rb
+++ b/app/presenters/hyku/manifest_enabled_work_show_presenter.rb
@@ -8,32 +8,37 @@ module Hyku
       manifest_helper.polymorphic_url([:manifest, self])
     end
 
-    # Hash to populate IIIF manifest with extras. Currently only sequence rendering is implemented.
+    # IIIF rendering linking property for inclusion in the manifest
     #
-    # @return [Hash] hash including sequence_rendering
-    def manifest_extras
-      {
-        sequence_rendering: sequence_rendering
-      }
+    # @return [Array] array of rendering hashes
+    def sequence_rendering
+      renderings = []
+      if solr_document.rendering_ids.present?
+        solr_document.rendering_ids.each do |file_set_id|
+          renderings << build_rendering(file_set_id)
+        end
+      end
+      renderings.flatten
+    end
+
+    # IIIF metadata for inclusion in the manifest
+    #
+    # @return [Array] array of metadata hashes
+    def manifest_metadata
+      metadata = []
+      metadata_fields.each do |field|
+        metadata << {
+          'label' => field.to_s.humanize.capitalize,
+          'value' => get_metadata_value(field)
+        }
+      end
+      metadata
     end
 
     private
 
       def manifest_helper
         @manifest_helper ||= ManifestHelper.new(request.base_url)
-      end
-
-      # IIIF rendering linking property for inclusion in the manifest
-      #
-      # @return [Array] array of rendering hashes
-      def sequence_rendering
-        renderings = []
-        if solr_document.rendering_ids.present?
-          solr_document.rendering_ids.each do |file_set_id|
-            renderings << build_rendering(file_set_id)
-          end
-        end
-        renderings.flatten
       end
 
       # Build a rendering hash
@@ -57,6 +62,27 @@ module Hyku
         ActiveFedora::SolrService.query("id:#{file_set_id}",
                                         fl: [ActiveFedora.id_field, 'label_ssi', 'mime_type_ssi'],
                                         rows: 1)
+      end
+
+      # Retrieve the required fields from the Form class. Derive the Form class name from the Model name.
+      #   The class should exist, but it may not if different namespaces are being used.
+      #   If it does not exist, rescue and return an empty Array.
+      #
+      # @return [Array] required fields
+      def metadata_fields
+        ns = "Hyrax"
+        model = solr_document['has_model_ssim'].first
+
+        "#{ns}::#{model}Form".constantize.required_fields
+      rescue
+        []
+      end
+
+      # Get the metadata value(s).
+      #
+      # @return [Array] field value(s)
+      def get_metadata_value(field)
+        Array.wrap(send(field))
       end
   end
 end
--- a/spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
+++ b/spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
@@ -8,7 +8,9 @@ RSpec.describe Hyku::ManifestEnabledWorkShowPresenter do
 
   describe "#manifest_url" do
     subject { presenter.manifest_url }
+
     let(:document) { { "has_model_ssim" => ['GenericWork'], 'id' => '99' } }
+
     it { is_expected.to eq 'http://test.host/concern/generic_works/99/manifest' }
   end
 
@@ -25,9 +27,9 @@ RSpec.describe Hyku::ManifestEnabledWorkShowPresenter do
     end
   end
 
-  describe "manifest_extras" do
+  describe "#sequence rendering" do
     subject do
-      presenter.manifest_extras
+      presenter.sequence_rendering
     end
 
     before do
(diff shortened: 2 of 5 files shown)


app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -58,5 +63,30 @@ def query_for_rendering(file_set_id)
...
+      end
+
+      # Get the metadata value; return a String for single values and an Array of hashes for multiple values.
+      #
+      # @return [String or Array] field value(s)
+      def get_metadata_value(field)
+        if send(field).count == 1
+          send(field).first
+        else
+          send(field).collect { |f| { "@value" => f } }
```
cjcolvar: I'm not finding an example of this exact construct in the presentation API.  The examples I see of an array of hashes is for associating values with language codes.  There is one example hidden in the spec of an array of strings which might be a better fit here:
```
{
  "label": "Notes",
  "value": ["Text of note 1", "Text of note 2"]
}
```
That being said I tried to see if an array of hashes without language codes was valid using the online validator and it looks like it isn't really checking the contents of a metadata hash's value if it is an array.  But it does look like it is okay with hashes that only have "@value" and no "@language" in general.
https://github.com/iiif-prezi/iiif-prezi/blob/master/iiif_prezi/loader.py#L161-L190

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
```diff
@@ -37,7 +37,18 @@
...
+      expect(subject).to be_an Array
+    end
+  end
+
+  describe "#manifest metadata" do
+    subject do
+      presenter.manifest_metadata
+    end
+
+    it "returns a hash containing the metadata" do
```
cjcolvar: There should probably be a test for the case of multi-valued fields.

geekscruff: Update incoming ...

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: Thanks @cjcolvar - I just went with @value without @language - it rendered in UV fine. But I think your suggestion is a little cleaner.  Update incoming ...

cjcolvar: 👍 How it looks in UV is probably the best test since I'm just looking at the json.

geekscruff:
@cjcolvar Would you feel able to do a final review / merge of this PR? I've updated the iiif_manifest gem now so once this is merged, the metadata will start showing in UV.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -58,5 +63,30 @@ def query_for_rendering(file_set_id)
                                         fl: [ActiveFedora.id_field, 'label_ssi', 'mime_type_ssi'],
                                         rows: 1)
       end
+
+      # Retrieve the required fields from the Form class. Derive the Form class name from the Model name.
+      #   The class should exist, but it may not if different amespaces are being used.
```
cjcolvar: Typo: `amespaces` -> `namespaces`

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
```diff
@@ -58,5 +63,30 @@ def query_for_rendering(file_set_id)
...
+      end
+
+      # Get the metadata value; return a String for single values and an Array of hashes for multiple values.
+      #
+      # @return [String or Array] field value(s)
+      def get_metadata_value(field)
+        if send(field).count == 1
+          send(field).first
+        else
+          send(field).collect { |f| f }
```
cjcolvar: I don't think it hurts for the value to be an array of a single string so it might just be easiest to do 
```
def get_metadata_value(field)
  Array.wrap(send(field))
end
```

config/solr_wrapper_test.yml
```diff
@@ -1,5 +1,7 @@
 # Place any default configuration for solr_wrapper here
 port: 8985
+# TODO: Relax this when we're Solr 7-compliant
+version: 6.6.1
```
cjcolvar: This got fixed in some other places.  Does this restriction still make sense for Hyku?

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
```diff
@@ -37,7 +39,30 @@
...
+      expect(subject).to be_an Array
+    end
+  end
+
+  describe "#manifest metadata" do
+    subject do
+      presenter.manifest_metadata
+    end
+
+    context do
```
cjcolvar: This first context could be removed if values were always arrays.

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
```diff
@@ -37,7 +39,30 @@
...
+      it "returns a hash containing the metadata with a single value" do
+        expect(subject).to be_a Array
+        expect(subject[0]['value']).to eq 'Test title'
+      end
+    end
+
+    context do
```
cjcolvar: This wrapping context would be unnecessary if the previous context is removed.

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
```diff
@@ -37,7 +39,30 @@
...
+      before do
+        work.title = ['Test title', 'Another test title']
+      end
+
+      it "returns an array of metadata values" do
+        expect(subject[0]['value']).to include('Test title', 'Another test title')
+      end
```
cjcolvar: It would be good to add a test for the metadata pair's label as well.

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: done!

app/presenters/hyku/manifest_enabled_work_show_presenter.rb
geekscruff: good point, done!

config/solr_wrapper_test.yml
geekscruff: I've pulled in the updated solr schema from master and removed this, should pass

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
geekscruff: done!

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
geekscruff: done!

spec/presenters/hyku/manifest_enabled_work_show_presenter_spec.rb
geekscruff: done!

geekscruff:
@cjcolvar those are done :)

PR #1511, opened 2017-12-08
Fix invalid foreign key error in languages importer

geekscruff:
See Hyrax [issue 1034](https://github.com/samvera/hyrax/issues/1034)

This is caused by an incorrect foreign key. This PR adds a new migration to remove the incorrect foreign key and add the correct one.

Tested by running `hyrax:controlled_vocabularies:language` which now successfully imports lexvo.


geekscruff:
This PR OK and all, but it's not working in the form: #1512

PR #1526, opened 2018-04-23
Add username field on sign up

cldambrosio:

Add name field on sign up in order to prepopulate the contact form with user name instead of an email address.

screen shot 2018-04-23 at 12 55 07

screen shot 2018-04-23 at 12 57 47

We implemented this change in [our fork of Hyku.](https://github.com/ubiquitypress/hyku)
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyku/registrations_controller.rb
+++ b/app/controllers/hyku/registrations_controller.rb
@@ -1,5 +1,6 @@
 module Hyku
   class RegistrationsController < Devise::RegistrationsController
+    before_action :configure_permitted_parameters
     def new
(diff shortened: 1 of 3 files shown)


orangewolf:

cldambrosio:
It's required at the moment, that could certainly be amended.

geekscruff:
@cldambrosio It's not a massive thing, but I wonder if anyone else has a view @orangewolf @mjgiarlo

mjgiarlo:
I'm :ok: with this change, even with the name field being required.

PR #1527, opened 2018-05-21
Hook up supported metadata tags

cldambrosio:
This already exists in [Hyrax](https://github.com/samvera/hyrax/blob/ca0766b3cc1b83859ce1b08d7cafbda18be07c84/app/views/hyrax/base/show.html.erb#L3)

before change:
without_meta
The change proposed populates the metadata tags drafted [here](https://github.com/samvera/hyrax/blob/ca0766b3cc1b83859ce1b08d7cafbda18be07c84/app/views/shared/_citations.html.erb).
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -1,4 +1,5 @@
 <% provide :page_title, @presenter.page_title %>
+<%= render 'shared/citations' %>
 <div itemscope itemtype="http://schema.org/CreativeWork" class="row">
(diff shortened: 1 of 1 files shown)


cldambrosio:
Also see additional fields added in PR #1528

orangewolf:

cldambrosio:
Thank you @orangewolf for adding a note about the 4 documented failures on all the PRs! I'll do that in the future.

PR #1528, opened 2018-05-21
Add more citations tags

cldambrosio:
More fields added to the gscholar metadata tags: description, keywords, publisher.

Blank fields don't display, as per Google Scholar recommendation to "skip/drop all missing fields".

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -1,4 +1,5 @@
 <% provide :page_title, @presenter.page_title %>
+<%= render './shared/additional_citations' %>
 <div itemscope itemtype="http://schema.org/CreativeWork" class="row">
   <div class="col-xs-12">
     <header>
--- a/app/views/shared/_additional_citations.html.erb
+++ b/app/views/shared/_additional_citations.html.erb
@@ -1,0 +1,11 @@
+<% content_for(:gscholar_meta) do %>
+  <% unless @presenter.description.blank? %>
+    <meta name="description" content="<%= @presenter.description.join %>" />
+  <%end %>
+  <% unless @presenter.keyword.blank? %>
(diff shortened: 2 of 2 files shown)


cldambrosio:
Also see PR #1527

orangewolf:

geekscruff:
This one would be ideal for Hyrax ... added directly to https://github.com/samvera/hyrax/blob/master/app/views/shared/_citations.html.erb. I know that will delay it getting into Hyku, but would you be happy to port it over to Hyrax?

cldambrosio:
Thanks Julie, yes it would be ideal to put this in Hyrax's main citations view file. I don't have a Hyrax installation though so I can't port it over easily. I think we'll need to be able to contribute to Hyrax as well as Hyku so let me see with Francesco what we can do (when he's back from vacation).

cjcolvar:
This was merged in Hyrax master with https://github.com/samvera/hyrax/pull/4382.  Does it need to be backported into a release that Hyku uses in order to close this one?

orangewolf:
@cjcolvar I closed this on accident (was cleaning up the other one) and can't reopen it because the original branch was removed. I would say there are 2 ways forward. Get  samvera/hyrax#4382 in a 2.x release or just open a new PR against Hyku that moves the one view file over with a comment at the top that says something like `<%# override hyrax 2.9 to backport changes made in master as of 2020-07-30 for additional google scholar functionality %>` so that we know why we did it.

cjcolvar:
We have this change already locally so we're not in a rush for it.  I'm fine with either way but might lean towards the backport so we don't have to remember to cleanup anything.  I'm happy to do the backport if you think that's the best way to go.

orangewolf:
Makes sense to me

PR #1529, opened 2018-05-24
[WIP] Hyrax 2.1 upgrade

### LINKED ISSUE #1512: Autosuggest fields fail with JQuery / JS error
### Descriptive summary

Having fixed that in #1511 I have testing the auto-suggest feature on the languages field. This is not working. The following error is observed in the Javascript: 

`Uncaught TypeError: elem.getClientRects is not a function`

image

The issue was widely reported in June/July 2016, for example [here](https://stackoverflow.com/questions/37914869/jquery-ui-error-f-getclientrects-is-not-a-function) and the cause seems to be the version jQueryUI and jQuery 3.0.0.

I am not seeing this error in Hyrax, so there must be some jQuery version difference going on.

### LINKED ISSUE #1517: Universal Viewer Bug: Clicking iiif icon throws error
clicking the iiif icon throws error. 

See: https://github.com/curationexperts/nurax/issues/186

### LINKED ISSUE #1518: Search "My shares" gives error messages
### Descriptive summary

/dashboard/shares page is not present in hyrax/nurax/hyku. 

Nurax: QA Testing for Hyrax
Hyrax Version: 2.1.0.beta1
Hyrax Branch: v2.1.0.beta1_qa
Hyrax Revision: 8fe35f54e067fd499a4467b9ccbefa48233d43a2 

Hyku v1.0.0.beta2 Docker 

Firefox 58.0.2 (64-bit) canonical 1.0

### Rationale

This is a bug.

### Expected behavior
Search within the with the option "My shares" selected presents results on my shared work. 
The search fields which allow this selection of "My shares" are located at    
* hyrax starting page (eg. see https://nurax.curationexperts.com/) and 
* that very field on the top of the Edit dialog of a single work.

### Actual behavior
Search with the  option "My shares" selected directs to /dashboard/shares which is not preset.

1. Go to  
* hyrax starting page (eg. see https://nurax.curationexperts.com/)  
 or
* that very field on the top of the Edit dialog of a single work.
1. Then select "My Shares" from the pull down right next to the search "Go" button
1. (optional) enter search term 
1. Press the "Go" button 

### Related work

Link to related tickets or prior related work here.

https://github.com/samvera/hyrax/issues/2767

orangewolf:
Fixes #1518; Fixes #1517; Fixes #1512; ... more to come, just haven't had time to open them all and see yet.

Branch for work on the Hyrax 2.1 upgrade.

Currently upgraded to rc3 on the theory that we can time the 2.1 release fairly closely if we get a lot of it done up front. Right now there are 40 failing specs (thus the WIP up top ;-) ), but we wanted to get this up and visible so that there isn't any duplicate effort.

@samvera/hyrax-code-reviewers


orangewolf:
Worth noting that this depends on the solr_wrapper work, since I couldn't get solr_wrapper to work w/o it.

orangewolf:

.gitignore
```diff
@@ -46,3 +46,4 @@ config/environments/*.local.yml
 
 # Avatars and such
 public/uploads
+.DS_Store
```
jcoyne: This should be in your global ignore https://help.github.com/articles/ignoring-files/#create-a-global-gitignore  as this is specific only to certain operating systems and it not part of the project.

orangewolf: no problem

orangewolf:
@jcoyne not sure how far you got review-wise, but rake tasks tests were added to get coveralls happy and the gitignore has been updated.

jcoyne:
My only remaining question is why did you add `rack-test` as a dependency.  I don't see where you've used `Rack::Test` anywhere in the code.

orangewolf:
@jcoyne great question: it was already a dependency (it is a dependency of hyrax as well), but was uncapped version wise. Newer versions are not compatible with the version of Capybara we are using.  From the rack-test README:
```
rack-test >= 0.71 does not work with older Capybara versions (< 2.17). See #214 for more details.
```
Thus we needed to call that out explicitly to get the app to bundle.

jcoyne:
@orangewolf can you add a comment to that effect to the Gemfile?  Is upgrading to a newer version of Capybara not an option?

orangewolf:
@jcoyne sorry, finally got a chance to get back to this.

jcoyne:
@orangewolf now this is looking good. Do you want to squash any of these commits together?

orangewolf:
@jcoyne I'd recommend squishing the whole thing together on the merge. Are you ok with that or would you rather have them squished incrementally?

PR #1533, opened 2018-06-11
Log downloads

cldambrosio:
Logging and displaying download statistics independently from Google Analytics

Refs [Hyrax#2539](https://github.com/samvera/hyrax/issues/2539)

*  In the FileDownloadStats database table, add a `title` and change the `date` field to an array; to enable filtering the data by Work and by Date.
* Create and update records when a file is downloaded
* Render these stats in the dashboard
stats

@samvera/hyrax-code-reviewers


app/decorators/controllers/hyrax/downloads_controller_decorator.rb
```diff
@@ -0,0 +1,30 @@
+module Hyrax
+  DownloadsController.class_eval do
```
jcoyne: This is an unusual way to do this.  I like to do this same sort of thing by making a module and include it in the `DownloadsController`

in `config/initializers/hyku.rb` (or something) add:
`Hyrax::DownloadsController.include DownloadsControllerDecorator`

app/decorators/controllers/hyrax/downloads_controller_decorator.rb
```diff
@@ -0,0 +1,30 @@
...
+    # `asset` is inherited from hydra-head/hydra-core/app/controllers/concerns/hydra/controller/download_behavior.rb
+    def send_content
+      super
+      update_download_stats(asset)
+    end
+
+    def download_stats(current_file)
```
jcoyne: Because these methods have nothing to do with HTTP interaction they could move out of the controller to a service class, so that the only addition on line 6 needs to be: 
```
DownloadStats.update(asset)
```

app/decorators/controllers/hyrax/downloads_controller_decorator.rb
```diff
@@ -0,0 +1,30 @@
...
+      if FileDownloadStat.find_by(file_id: current_file.id)
+        FileDownloadStat.find_by(file_id: current_file.id)
```
jcoyne: One common practice to avoid long `if/else` blocks is do to:
```ruby
stat = FileDownloadStat.find_by(file_id: current_file.id)
return stat if stat
... (the stuff that was in the else block)...
```

app/decorators/controllers/hyrax/admin/stats_controller_decorator.rb
```diff
@@ -0,0 +1,11 @@
+module Hyrax
+  Admin::StatsController.class_eval do
+    def show
+      super
+      @works_to_display ||= ::FileDownloadStat.all.select do |e|
```
jcoyne: This will load all database records into memory and iterate over them trying to find records that match this condition. But if you do `::FileDownloadStat.where(user_id: current_user.id)` then the database will do all the work and you'll only load into memory the records you care about.

app/decorators/controllers/hyrax/admin/stats_controller_decorator.rb
```diff
@@ -0,0 +1,11 @@
...
+        e[:user_id] == current_user.id
+      end
+      @total_downloads ||= @works_to_display.map(&:downloads).reduce(:+)
```
jcoyne: why are you using `||=` in this method? Is there somewhere else that could have already set `@total_downloads`?  If so, can you add a comment about where that might occur?  Otherwise just use simple assignment (`=`)

app/views/hyrax/admin/stats/_downloads.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2>Total Downloads</h2>
+<ul>
+  <%= @total_downloads %>
+</ul>
+<br/>
+<h2>Downloads per Work</h2>
+<ul>
+  <% @works_to_display.each do |dwnld| %>
+    <li><%= dwnld.title %></li><span class="count">(<%= dwnld.downloads %>)</span>
```
jcoyne: This markup seems odd. I'm not sure `ul` can have any children besides `li`. Did you want to tuck the `span` within the `li`?

db/migrate/20180606101810_change_file_download_stats_date_to_array.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeFileDownloadStatsDateToArray < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :file_download_stats, :date, :datetime
+    add_column :file_download_stats, :date, :datetime, array: true, default: []
```
jcoyne: I think this change may end up breaking this code: https://github.com/samvera/hyrax/blob/be223ff41debc86bac9a5e74ac4151102513e9ac/app/models/hyrax/statistic.rb#L53

Are you sure you don't want to just make a new database table rather than trying to reuse one?

app/views/hyrax/admin/stats/_downloads.html.erb
cldambrosio: Yes `span` within `li` is nicer thanks. I'll see how else we may want this styled.

app/decorators/controllers/hyrax/downloads_controller_decorator.rb
cldambrosio: Sounds good, I'll try this :+1:

app/decorators/controllers/hyrax/downloads_controller_decorator.rb
cldambrosio: :100:

app/decorators/controllers/hyrax/admin/stats_controller_decorator.rb
cldambrosio: :100:  thanks again!!

cldambrosio:
I'll work on it again tomorrow. @jcoyne thanks a ton for helping me improve this :+1:

cldambrosio:
Style changed to:

downloads_report

orangewolf:
@cldambrosio looks like this needs to have master merged in to it and that will likely help with these test failures.

cldambrosio:
Thank you @orangewolf I will do that soon!

Issue #1524, opened 2018-04-19
CI builds from forks never pass


cjcolvar:
### Descriptive summary

The AWS tests (https://github.com/samvera-labs/hyku/blob/master/spec/models/uploaded_file_spec.rb#L39-L108) fail when building from a fork of the repo but pass when run from a branch.  My guess is that there are travis ENV variables which aren't available when building from a fork since that would be a security issue.  These tests should either be turned off or made to work when building from a fork.

### Rationale

Developers that don't have commit access to Hyku need to be able to submit PRs and have their builds reflect the actual test results from their code change.  Reviewers shouldn't have to dig into the travis build and inspect the failures to know whether the changed code caused failed tests or not.

### Expected behavior

Tests pass when making unrelated code changes.

### Actual behavior

Tests fail.

### Steps to reproduce the behavior

1. Fork Hyku
1. Make a PR to Hyku

### Related work

See the same PR made from a fork (https://github.com/samvera-labs/hyku/pull/1522) and from a branch (https://github.com/samvera-labs/hyku/pull/1523).

PR #1539, opened 2018-06-29
Fix s3 specs - part 2

orangewolf:
As hypothesized, the specs are not running in master any more.  That appears to be because TRAVIS_PULL_REQUEST_SLUG if you are not talking about a pull request. This should work in both cases now.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -55,8 +55,11 @@ RSpec.configure do |config|
   config.run_all_when_everything_filtered = true
 
   # only run aws tests from CI (or w/ `--tag aws`) and only run it on the main repo, since that
-  # is where the valid aws keys live
-  config.filter_run_excluding(aws: true) unless ENV['CI'] && ENV['TRAVIS_PULL_REQUEST_SLUG'].match('samvera-labs/hyku')
+  # is where the valid aws keys live. TRAVIS_PULL_REQUEST_SLUG is "" when the job is a push job
+  unless ENV['CI'] &&
+         (ENV['TRAVIS_PULL_REQUEST_SLUG'].match('samvera-labs/hyku') || ENV['TRAVIS_PULL_REQUEST_SLUG'].blank?)
+    config.filter_run_excluding(aws: true)
+  end
 
   # Allows RSpec to persist some state between runs in order to support
   # the `--only-failures` and `--next-failure` CLI options. We recommend


orangewolf:
I'll check master again once this is merged to make sure the AWS specs are running there again.

spec/spec_helper.rb
```diff
@@ -55,8 +55,11 @@
   config.run_all_when_everything_filtered = true
 
   # only run aws tests from CI (or w/ `--tag aws`) and only run it on the main repo, since that
-  # is where the valid aws keys live
-  config.filter_run_excluding(aws: true) unless ENV['CI'] && ENV['TRAVIS_PULL_REQUEST_SLUG'].match('samvera-labs/hyku')
+  # is where the valid aws keys live. TRAVIS_PULL_REQUEST_SLUG is "" when the job is a push job
+  unless ENV['CI'] &&
+         (ENV['TRAVIS_PULL_REQUEST_SLUG'].match('samvera-labs/hyku') || ENV['TRAVIS_PULL_REQUEST_SLUG'].blank?)
```
jcoyne: In this case, could you just remove the `TRAVIS_PULL_REQUEST_SLUG` entirely?  I'm not understanding what situation this is trying to prevent. Maybe you could modify the commit message to explain a bit more.

orangewolf: @jcoyne here is the description from part 1.  The issue is that only the travis ci from this repo is set to run the S3 specs successfully. So all forked pull request spec sets fail on the S3 connection tests.  BUT we want those test to run after a branch is merged to make sure there is no regression. This commit prevents the tests from running locally (as it was before) and from running on forks, but makes sure they run on pull requests from within this repo and on merged branches. This is part 2, simply because I can't test that the specs actually run like they are supposed to until it is merged it is a little tricky to know if we got the variables right for all the cases w/o getting a merge =-)

---
Fixes #1524

No longer run S3 related specs on non-samvera-labs repos. This will allow specs to pass on forks.

This is a first swing at getting Travis to not run the specs unless the code is in samvera-labs/hyku. Specs on forks will thus pass and the 4 failing S3 specs from forks will get run properly once code is merged.

Note - need to check that the specs in question are run after this is merged.
---

jcoyne: Sounds good. 👍

jcoyne:
Do you want to squash these commits together?

orangewolf:
@jcoyne finally got back to squashing this.

Issue #1238, opened 2017-06-13
CLI importers are broken


jcoyne:
This commit removes the actor that handled the `:files` attribute: https://github.com/samvera/hyrax/commit/3f1b58195d4381c51fde8b9149016c5b09f0c9b4

The ObjectFactory and its descendants expect the `:files` attribute to be handled by the work actor.

https://github.com/samvera-labs/hyku/blob/master/lib/importer/factory/object_factory.rb#L133

This probably affects the CSVImporter but not the PURLImporter since the latter uses remote files: https://github.com/samvera-labs/hyku/blob/master/app/jobs/import_work_from_purl_job.rb#L47

jcoyne:
@dunn I think you'll have this problem if you upgrade to Hyrax.

dunn:
Thanks for the heads-up!

atz:
Not sure I quite follow this.

The `files` attrib from [object_factory.rb#L136](https://github.com/samvera-labs/hyku/blob/master/lib/importer/factory/object_factory.rb#L136) goes through:
```ruby
def file_attributes
  files_directory.present? && files.present? ? { files: file_paths } : {}
end
...
def transform_attributes
  StringLiteralProcessor.process(attributes.slice(*permitted_attributes)).merge(file_attributes)
end
```
to end up used in `create` and `update`.  `update` always uses `work_actor`, `create` does unless the thing being created is a Collection.

`work_actor` is `Hyrax::CurationConcern.actor`, defaulting to:
```ruby
Hyrax::DefaultMiddlewareStack.build_stack.build(Actors::Terminator.new)
``` 
which has to be one of the **least** intelligible pieces of indirection in our stack, layering on **17 other middleware actors**.  @jcoyne, your contention is that none of these actually does what we need anymore?  Not even `CreateWithFilesActor`?

@mjgiarlo What is the right direction to fix this:
- Revert the Hyrax commit? 
- Add another actor similar to the old one in Hyku and append to the `@work_middleware_stack`?
- Convert raw files to `Hyrax::UploadedFile` objects and rely on an existing actor?
- Amend existing actor to catch `:files` attribute again?

mjgiarlo:
@atz I'd be most inclined in favor of `Convert raw files to Hyrax::UploadedFile objects and rely on an existing actor`. You?

atz:
[...]

Issue #1164, opened 2017-05-28
Fully document import by CVS


hannahfrost:

mjgiarlo:
@hannahfrost @bbranan 

I've tested the CSV importer and here's what I found.

First, I created a simplified version of the CSV with a single image-related object (which actually maps to the `GenericWork` type rather than the `Image` type, which confused me). You can see that in #1395

Next, I copied a smallish PNG to `tmp/hyku-objects`, and then ran the command: `./bin/import_from_csv foo.localhost spec/fixtures/csv/simple.csv tmp/hyku-objects`

I found that I needed to make a small number of code changes to get the script to run. `AccountElevator` expects there to be a global `logger` object, so I added that to the script, and there's also some undocumented dependency between imported objects and collections. So I added a way to ignore that dependency.

Here are some shortcomings that will/may need addressing before we consider this useful in a hosted service context:

* [Files are never attached to the works that are imported](https://github.com/samvera-labs/hyku/blob/master/lib/importer/factory/object_factory.rb#L132-L137)
* CSV import should allow a depositor (email) to be specified on the command-line. By default, the depositor is set to a system user, so you can only see it if you are an administrator.
* CSV import could allow an AdminSet ID to be specified on the command-line. It defaults to the default AdminSet.
* [CSV import should accept a column for visibility](https://github.com/samvera-labs/hyku/issues/193)
* CSV import may need to handle workflow in some way if an item is deposited into an AdminSet that isn't using the self-deposit workflow
* It is possible to import data that does not match our vocabularies, e.g., [resource type](https://github.com/samvera-labs/hyku/blob/master/lib/importer/factory/image_factory.rb#L10-L13)
* It is possible to import data that would not pass form validation. For instance, the current CSV format does not include a field for rights statements. (This doesn't cause errors, but does create data that looks odd, and may make it difficult to edit an imported object without needing to add values for these fields.)

atz:
This blocks #1238

orangewolf:
replaced by Bulkrax circa 2023

PR #1541, opened 2018-07-06
CSVImporter file upload bug fix

cldambrosio:
Enables uploading files via the [`import_from_csv`](https://github.com/samvera-labs/hyku/blob/master/bin/import_from_csv) script.

refs #1238 #1164 

I included assets which I used to see the CSV importer at work, with the command:
```ruby bin/import_from_csv tenant ./lib/assets/csv_test.csv ./lib/assets/batch-upload-test/```
(where `tenant` needs to be replaced by your actual one - for instance `foo.localhost`)
If you try it you may need to adapt the command to your set up. These assets could be removed if preferred.

- Notes on tests:

I commented out 4 breaking tests from the codebase because they don't appear useful. The functionality was broken while the tests passed previously.

Thanks @geekscruff for sharing your code in [leaf_addons](https://github.com/leaf-research-technologies/leaf_addons/blob/master/lib/generators/leaf_addons/templates/lib/importer/files_parser.rb#L65-L69) - I used it :)

@samvera/hyrax-code-reviewers

### CODE REVIEW

File: lib/importer/attach_files_to_work.rb [hunk 1]
```diff
@@ -0,0 +1,14 @@
+# Called in 'object_factory'
+module Importer
+  class AttachFilesToWork < AttachFilesToWorkJob
```
jcoyne: Rather than doing this, would it be possible to coerce the input data to an Array of files (with only one element)?  That means that any upstream changes won't have to be ported down to Hyku?

lib/importer/factory/object_factory.rb
```diff
@@ -25,12 +26,13 @@ def run
         object
       end
 
+      ## FOR CONSIDERATION: handle a row (i.e. Work) with more than one file:
+      ## currently the file_set is replaced on update
       def update
         raise "Object doesn't exist" unless object
-        run_callbacks(:save) do
```
jcoyne: Why are all the callbacks removed?

lib/importer/factory/object_factory.rb
```diff
@@ -64,23 +66,21 @@ def search_by_identifier
...
-      def log_created(obj)
-        msg = "Created #{klass.model_name.human} #{obj.id}"
-        Rails.logger.info("#{msg} (#{Array(attributes[system_identifier_field]).first})")
-      end
-
-      def log_updated(obj)
-        msg = "Updated #{klass.model_name.human} #{obj.id}"
-        Rails.logger.info("#{msg} (#{Array(attributes[system_identifier_field]).first})")
-      end
+      ## below methods are commented out to pass rubocop inspection ("Class has too many line");
```
jcoyne: Rather than commenting out the logging feature, how about changing the threshold in `.rubocop_todo.yml`?

cldambrosio: It didn't occur to me to change the enforced practice - will do thanks!

lib/importer/factory/object_factory.rb
cldambrosio: I didn't understand their use! Everything seems to work without. With them the method's Assignment Branch Condition size throws a rubocop error - would you prefer me to change the rubocop threshold for this too?

lib/importer/attach_files_to_work.rb
cldambrosio: I see how overriding upstream changes could become undesirable - I'll work on this thanks!

lib/importer/factory/object_factory.rb
jcoyne: Yes, adjust the thresholds as needed.  If the callbacks aren't used, then make a separate PR to   "remove unused callbacks"  It's easer to understand PRs if they constrain themselves to a single change.

cldambrosio: Ok thanks! I just restored the callbacks - that didn't cause a rubocop failure anymore. I think I was getting that on my initial solution before extracting logic in new methods.

lib/importer/factory/object_factory.rb
```diff
@@ -25,11 +25,16 @@ def run
         object
       end
 
+      ## FOR CONSIDERATION: handle a row (i.e. Work) with more than one file:
+      ## currently the file_set is replaced on update
       def update
         raise "Object doesn't exist" unless object
+        @attr = update_attributes
```
jcoyne: Why have you chosen to use an instance variable here? Why not a local?

lib/importer/factory/object_factory.rb
```diff
@@ -25,11 +25,16 @@ def run
...
         run_callbacks(:save) do
-          work_actor.update(environment(update_attributes))
+          work_actor.update(environment(@attr))
         end
+        destroy_existing_file_set if object.file_sets.present?
+        attach_file_to_work
```
jcoyne: attaching a file to a work is the responsibility of the work actor. Why are you adding it here in duplicate?

lib/importer/factory/object_factory.rb
```diff
@@ -25,11 +25,16 @@ def run
...
+      ## FOR CONSIDERATION: handle a row (i.e. Work) with more than one file:
+      ## currently the file_set is replaced on update
       def update
         raise "Object doesn't exist" unless object
+        @attr = update_attributes
         run_callbacks(:save) do
-          work_actor.update(environment(update_attributes))
+          work_actor.update(environment(@attr))
         end
+        destroy_existing_file_set if object.file_sets.present?
```
jcoyne: Are you sure you want to destroy an existing file set?  Why not update it?  If you replace one fileset with another, people who have cited an existing URI are going to have their references broken.  What's the benefit?

lib/importer/factory/object_factory.rb
cldambrosio: It is called in `attach_file_to_work` (line 137).

lib/importer/factory/object_factory.rb
cldambrosio: Good question; I am not sure that is the best way forward. This is how I envision the use of the csv importer: 
- If I want to update a work with an added file, I would add columns to my csv. I could keep the 1st file or reorder file columns to pick a different one (the thumbnail will be rendered from the 1st).
- If I delete a file from the csv I would expect not to have it appear in the work anymore.

cldambrosio:
I see; since the Actors are in Hyrax and we basically only [point at them in Hyku](https://github.com/samvera-labs/hyku/blob/master/app/actors/hyrax/actors/generic_work_actor.rb) the ObjectFactory seemed to be the right place to add code to fix this. I'll try moving the logic in an Actor instead.

lib/importer/factory/object_factory.rb
jcoyne: It is done in the actor stack here: https://github.com/samvera/hyrax/blob/f43e7048bf3961f48812809897b36d84b6577f68/app/actors/hyrax/actors/create_with_files_actor.rb#L10  now.  So you may need to upload the file and then set the uploaded file id in the `:uploaded_files` parameter before calling the actor.

lib/importer/factory/object_factory.rb
jcoyne: I'm not sure that's universal though.  What if I only want to update titles. Reuploading all the files is going to be very slow.  And I don't want to have to do that just to update some metadata.

cldambrosio: Hi! Apologies for a late reply. Good point.
The method actually destroys the file set only if the file itself needs to be changed - the check is line 142 `f.destroy if attributes[:file] != f.title`. I will rename this method for clarity.
But the file _is_ re-uploaded; which is inefficient for sure. I'll make changes.

lib/importer/factory/object_factory.rb
cldambrosio: Thanks! Will work on this.

cldambrosio:
The code is currently broken - I am stuck and feedback is welcome. I tried to rely on an existing actor but for now I get an ActiveFedora::AssociationTypeMismatch Error: `Hydra::PCDM::File expected, got Hyrax::UploadedFile`

geekscruff:
Hi @cldambrosio ... I had that issue with attributes[:files] so I switched to using attributes[:uploaded_files] and attributes[:remote_files], cf. https://github.com/research-technologies/leaf_addons/blob/master/lib/generators/leaf_addons/templates/lib/importer/factory/base_factory.rb#L15

cldambrosio:
Thank you @geekscruff for coming to the rescue :)
I had not tried removing `:files` from the attributes, now with a few tweaks it works!
I have not included [:remote_files] so far, simply because I don't understand where that comes from - any hint?
I'll run the tests and push a commit later.

jcoyne:
`:remote_files` would be used if the files were in box.com or google drive or something like that (not locally uploaded)

cldambrosio:
Makes sense :sweat_smile: I just had not come across it in Hyku so far so I was not sure I'd need to include this.

cldambrosio:
Hi @jcoyne, do you think this is ready to merge?

PR #1543, opened 2018-08-22
update to hyrax 2.2

orangewolf:
Fixes #1542 

As it says on the tin - New Hyrax Version

From upstreams release notes:

Hyrax 2.1 was released on May 31st, 2018. Since then, there have been new features and bug fixes. This release contributes the following updates to Hyrax:

## New Features

Repository admins can configure the visibility of the registered user list, hidden by default (#3197)
Repository admins can sort the "All Works" dashboard by date, status, and visibility (#3173)
User Interface

Hides deposit agreement progress when disabled (#3141)
Adds Display Name to Workflow Roles management page (#3140)
Change "Update Attached File" button to "Save" on file_set edit view (#3145)
Hides private work metadata from public view during approval/ review workflow (#3148)
Disables Work "Save" button during multi-file uploads (#3150)
Hides cancel upload button when file upload(s) are finished (#3156)
Reduces size of Delete Button the file uploads page for better alignment (#3157)
The "Review Submissions" dashboard displays Depositor name, not system ID (#3170)

## Accessibility:

Removes the aria-hidden flag that hides useful elements for screen readers (#3114)
Make Single-Use Links more accessible (#3150)
Enables zooming (#3158)
File upload buttons are now keyboard clickable (#3186)
Makes breadcrumbs accessible with appropriate labels (#3188)
Dependencies:

Formal support for Ruby 2.3, 2.4, and 2.5 (#3109)
Adds support for SolrWrapper v2.x (#3125)
Consolidate dependency on ActiveFedora::Base.logger (#3142)

## Bug fixes:

Duplicate exported graphs from nested works (#3110)
Removes a duplicate pagination view from Blacklight (#3115)
Adds Hyrax namespace to collection_path route helper (#3118)
Better handling of nil responses to CollectionPresenter#description (#3137)
File counts in admin dashboard accuracy (#3154)
Allows depositor to always see their own work, even during an approval process (#3161)
Refactors:

Inverts dependencies for AdminStatsPresenter (#3140)
Expanded support for "inactive" authority terms to Rights Statements (#3122)
Updates the ImportUrlJob to use a GET instead of a HEAD request for cloud file ingest (#3130)
Moves Hyrax::WorkRelation dependency closer in the stack (#3143)
Removes instances of data: { turbolinks: false } (#3150)

@samvera/hyrax-code-reviewers

PR #1545, opened 2018-10-03
Upgrade to Hyrax ~>2.3.0

cjcolvar:
This is just a refresh of #1543 to see if the failing tests are magically fixed.


orangewolf:
damn no magic. I have high hopes of digging in to this next week after I run my workshops

geekscruff:
@cjcolvar I fixed the importer errors in #1548 but there are still the 4 AWS S3 ones which I had thought were fixed a while back by @orangewolf

cjcolvar:
Maybe #1539 didn't actually help but caused these aws failures?  How about reverting that PR for now?  It seems like it might unblock this PR and thus also #1545 since they are both local branch and not in forks.

geekscruff:
Let me try that now ...

geekscruff:
@cjcolvar no joy #1549 same 4 errrors. I'm stumped!

PR #1548, opened 2018-10-04
Only add collection_type attribute to Collections in importer.

geekscruff:
Hopefully this fixes the failed importer tests.


geekscruff:
merged via #1543

PR #1550, opened 2018-10-09
update FactoryGirl to FactoryBot

### LINKED ISSUE #1530: Update FactoryGirl to FactoryBot
### Descriptive summary

FactoryGirl has been renamed FactoryBot. We should switch out the name to remain compatible with future updates and to better match upstream dependencies like Hyrax.

nwoodward:
Resolves #1530

Replaces FactoryGirl with FactoryBot

@samvera/hyrax-code-reviewers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,7 +55,7 @@ group :test do
   gem 'capybara'
   gem 'chromedriver-helper'
   gem 'database_cleaner'
-  gem 'factory_girl_rails'
+  gem 'factory_bot_rails'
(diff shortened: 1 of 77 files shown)


orangewolf:
@nwoodward thank you for this! looks like we have some rubocop failures. Can you work through those?

nwoodward:
Sure. I'll take a look.

nwoodward:
I think the rubocop failures are fixed. I'm not sure about the source of the new errors.

orangewolf:
this is part of a set of failing specs that have nothing to do with a change. Going to investigate, I'm betting a dns entry changed or something like that

PR #1555, opened 2018-10-21
Use polymorphic_url instead of _path for UV data-uri 

geekscruff:
UV data-uri needs full URI otherwise the embed code won't work on other sites

### DIFF
--- a/app/views/hyrax/base/_representative_media.html.erb
+++ b/app/views/hyrax/base/_representative_media.html.erb
@@ -1,7 +1,7 @@
 <% if presenter.representative_id.present? %>
   <% if presenter.representative_presenter.image? %>
     <%= UniversalViewer.script_tag %>
-    <div class="viewer-wrapper"><div class="uv viewer" data-uri="<%= main_app.polymorphic_path [main_app, :manifest, presenter] %>"></div></div>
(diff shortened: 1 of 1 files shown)


jcoyne:
I'm not sure I understand "other sites". Can you explain that to me?

geekscruff:
@jcoyne the embed code from Universal Viewer won't work because it will have the path and not the full URL on anything other than the current Hyku instance.

jcoyne:
@geekscruff But isn't the consumer of the `data-uri` attribute the UV script.  And since both are on the current host...aren't they the same?

geekscruff:
@jcoyne but if you take the embed code published by Universal Viewer and use that on an external site, surely then you need the full uri?

Example code:

```<div class="uv" data-locale="en-GB:English (GB),cy-GB:Cymraeg" data-config="" data-uri="/concern/published_works/000123946/manifest?locale=en" data-collectionindex="0" data-manifestindex="0" data-sequenceindex="0" data-canvasindex="0" data-xywh="-1557,-98,4353,1948" data-rotation="0" style="width:560px; height:420px; background-color: #000"></div><script type="text/javascript" id="embedUV" src="https://archive.kingsfund.org.uk/bower_includes/universalviewer/dist/uv-2.0.1/lib/embed.js"></script><script type="text/javascript">/* wordpress fix */</script>```

PR #1556, opened 2018-10-22
Support pinging fedora endpoints with basic auth

geekscruff:
We have a live Hyku where fedora shows as `down` in the system status. I think this must be because it uses authentication. When I try the ping directly in the rails console, I get `Ldp::NotFound`.

The current `ping` method is:
```
def ping
    ActiveFedora::Fedora.instance.connection.head('/').response.success?
  rescue StandardError
    false
end
```

Changing it to the following works with our authenticated fedora:
```
def ping
    ActiveFedora::Fedora.instance.connection.head('/').response.success?
  rescue Ldp::NotFound
    ActiveFedora.fedora.connection.http.head.success?
  rescue StandardError
    false
end
```

Feedback welcome on this change. The tests still pass but they *don't* currently test for an authenticated fedora. Advice on how to approach that, or on how to do this in a better way, would be welcome.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/fcrepo_endpoint.rb
+++ b/app/models/fcrepo_endpoint.rb
@@ -10,7 +10,9 @@ class FcrepoEndpoint < Endpoint
   end
 
   def ping
-    ActiveFedora::Fedora.instance.connection.head('/').response.success?
+    ActiveFedora::Fedora.instance.connection.head(
+      ActiveFedora::Fedora.instance.connection.connection.http.url_prefix.to_s
+    ).response.success?
   rescue StandardError
     false
   end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -805,7 +805,7 @@ GEM
       httparty (>= 0.11.0)
       oauth2 (>= 0.9.2)
     slop (4.6.2)
-    solr_wrapper (2.0.0)
+    solr_wrapper (2.1.0)
       faraday
       retriable
       ruby-progressbar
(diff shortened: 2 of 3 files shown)


app/models/fcrepo_endpoint.rb
```diff
@@ -11,6 +11,8 @@ def self.reset!
 
   def ping
     ActiveFedora::Fedora.instance.connection.head('/').response.success?
+  rescue Ldp::NotFound
+    ActiveFedora.fedora.connection.http.head.success?
```
orangewolf: is there a reason not to just always use the fedora.connection.http.head.success? why make the clearly not complete one first?

geekscruff: I'm not entirely sure it's testing the right thing. I need to have another dig at this.

orangewolf:
@geekscruff ran in to this situation today and we do not have auth on our Fedora.  I think that the current test is only looking at "/" which depending on how tomcat is running may or may not have a result. I'm suspicious that this test doesn't actually hit the fcrepo war at all. There has to be a cleaner way to get the URL part, but for now, could you test `ActiveFedora::Fedora.instance.connection.head(ActiveFedora::Fedora.instance.connection.connection.http.url_prefix.to_s).response.success?` on your instance and see if that passes properly for you?

geekscruff:
I get a true for that.

PR #1571, opened 2019-02-25
Remove IIIF code from Hyku, rely on Hyrax functionality

geekscruff:
Removes all local code and specs that duplicate IIIF functionality now in Hyrax. Also removes the universal viewer code in public/bower_includes - this now comes via hyrax from `pulibrary/pul_uv_rails`

Configures iiif support in hyrax to true.


orangewolf:
@geekscruff would love @cbeer promised eyes on this, but it otherwise looks ready to me.  Is there anything else that makes this "WIP"?

geekscruff:
@orangewolf nope, I just forgot to rename it!

orangewolf:
fixed the fact that I @ messages the wrong Chris here earlier.

orangewolf:
merged because Coveralls was broken, but is fixed now.

Issue #1577, opened 2019-03-18
Dockerization Overhaul Proposal


orangewolf:
### Descriptive summary

Support a fully dockerized workflow for Hyku. Docker should be usable for development, testing and production deployment.

### Rationale

Some goals including being able to run the tests against the same image you plan to take to production and the ability to quickly start up Hyku without having to build the images and bundle gems on your machine.  This should also include a real 'production ready' and a real 'development environment ready' configuration for people to utilize Hyku.

### Expected behavior

- Run Docker for development
- Run Docker for tests
- Run Docker for production

PR #1578, opened 2019-03-18
Overhauled Docker Configuration

### LINKED ISSUE #1562: Get rid of dockercloud/haproxy
[docker-compose.yml](https://github.com/samvera-labs/hyku/blob/bc5117444d957864b4df98a51c9473bd9a4fcfb6/docker-compose.yml) currently [pulls in](https://github.com/samvera-labs/hyku/blob/bc5117444d957864b4df98a51c9473bd9a4fcfb6/docker-compose.yml#L180) `dockercloud/haproxy:1.5.3`.

This is problematic for several reasons:
- The [`dockercloud/haproxy:1.5.3`](https://github.com/docker/dockercloud-haproxy/tree/1.5.3) image was released in july 2016, meaning this hasn't gotten any security updates for almost two and a half years. The [latest version](https://github.com/docker/dockercloud-haproxy/tree/master) version of that docker image would be 1.7.6.
- The `dockercloud/haproxy` docker image has been [officially unmaintained since December 2017](https://github.com/docker/dockercloud-haproxy/commit/24e5a4ab6aeac25392e921398fe837de4c4e7019).
- Even haproxy 1.7.6 would be [horrendously outdated](https://www.haproxy.org/download/1.7/src/CHANGELOG) (released in june 2016), the [current](https://www.haproxy.org/download/1.8/src/CHANGELOG) haproxy release is 1.8.14 from september 2018.

So please get rid of this.

orangewolf:
This commit is in preparation of beginning to release Docker images that can be used via Docker Hub with each new release. It also allows for fully dockerized development environment and production environments.

Two phase Docker build - base and main. The goal here are much faster build times by moving seldom changing items to the base image.

Moves to Passenger base image which allows a few real niceties:

  - for dev, you can restart the web process w/o restarting the container, saving a lot of time
  - for prod passenger is more effecient by default in a variety of configurations than puma. Puma frequenty requires more tuning
  - means nginx is serving assets, so no extra proxy is required

Note, this makes the build a little bigger, but with libreoffice and fits in there this change is a pretty small fraction in my opinion.

Uses a nulldb to compile assets during build, making deployment much faster. Also set up to allow for optional assets volume, which makes keeping cached assets possible

This Dockerfile moves fits to always be in the same place in the container. There is still a version file present, so we don't really need it in the path and that means we don't have to update multiple places when we change versions. Just
 in the Dockerfile.base

Sets the admin and default host in dockerfile and virtual_host attributes. That allows this setup to work in dev by default with common proxies like Dory and others.

Makes test adjustments to allow for Docker environment, adds an external Chrome driver docker container to the dev environment

Splits the production and dev docker-compose.yml files. Reduced dev zookeeper to 1, removed lb and memcached from dev. Splitting futher refinement of production docker-compose in to a seperate task as part of #1577.

Fixes #1562 ; refs #1577

@samvera/hyrax-code-reviewers


geekscruff:
@orangewolf looks like this is caused by a change to the solr collection name from hyku-test

PR #1579, opened 2019-03-18
Add Blacklight OAI to Hyku

### LINKED ISSUE #1499: OAI-PMH integration
To support standard harvesting by other systems.

orangewolf:
Fixes #1499 This work was sponsored by [PALNI](http://www.palni.org)

Brings in and configures the existing blackight-oai gem. 

This allows each tenant to have a /catalog/oai endpoint which provides OAI-PMH access to the collections within.  This is an initial feature offering, so I'm expecting it to become more sophisticated over time.

Screen Shot 2019-03-17 at 8 14 45 PM

@samvera/hyrax-code-reviewers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,7 @@ group :development do
 end
 
 gem 'blacklight', '~> 6.7'
+gem 'blacklight_oai_provider', '~> 6.0'
 
 gem 'hyrax', '~>2.3.0'
 gem 'rsolr', '~> 2.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -111,6 +111,10 @@ GEM
       bootstrap-sass (~> 3.0)
       openseadragon (>= 0.2.0)
       rails
+    blacklight_oai_provider (6.0.0)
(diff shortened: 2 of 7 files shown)


geekscruff:
@orangewolf  re-based on master to see if tests pass.

orangewolf:
@geekscruff I agree. Was rushing to get this up as it looks like someone else in slack needed it. Do you think an integration spec is sufficient? I’m not sure what unit tests would look like as we are just bringing in and configuring the library.

geekscruff:
@orangewolf yes, I don't think it needs much testing, just a "is it there and returning stuff"

orangewolf:
@geekscruff you mean I have to actually add the file to the repo? ;-) Expanded it a bit while I was at it. Let me know what you think.

geekscruff:
@orangewolf looks good, failures were a rubocop offense so have fixed that and am re-running

orangewolf:
finally jury rigged things locally enough to get around whatever is keeping fcrepo_wrapper from starting and was able to run the tests outside of dockerland.

PR #1608, opened 2020-01-03
add bulkrax behind feature flipper

orangewolf:
Put in bulkrax behind a feature flag


geekscruff:
@orangewolf I've made a few tweaks and it seems to be working (tested CSV and OAI) - I had a few issues with jobs, but I think that's a dev env problem not bulkrax.

Ideally I'd like to get [bulkrax#123](https://github.com/samvera-labs/bulkrax/pull/127) merged and in before we merge this one as it fixes some issues with the stats on the importers.

Issue #517, opened 2016-12-05
Add Contact page


jcoyne:
Per designs at https://docs.google.com/document/d/1WZIOL8omCseweV6OSP65SkSQCxtoXbtP65UVN9doBPY/edit#heading=h.4cnihetzl8p4

ggeisler:
I'm not sure that we want to add the Contact page to the Content Blocks configuration after all. Isn't the current Contact page a form, and in that sense different from the About and Help pages?

I know including the Contact page is in the design documented in the Google doc you're referencing, so the confusion is my fault, but I thought I had updated #440 with more recent designs (my intention for all designs is to only use the Google docs for discussing the preliminary design, and final design is always in a ticket). #440 doesn't have them so something happened there, but I'll update the other tickets with my newer designs (the same as in the Google Doc, I think, except I updated the tabs to reflect no Contact page).

hannahfrost:
@ggeisler Are you thinking that "Contact Form" is its own facet under Configuration > Settings?

ggeisler:
In the short term I'm thinking we just use the current Contact page we (I believe) inherit from Sufia/Hyrax. So there is no UI consideration necessary.

Longer term, we might want to add some UI configuration, which could be as simple as entering the email addresses to which the form submission should be sent, but that would probably just be a tab on another Settings page (i.e., might not warrant its own Configuration menu item). 

Of course we could completely reconsider how we want to provide a Contact page and make its configuration via the UI more powerful, but my thought is this isn't an important enough feature for MVP to do anything other than use what we inherit from Sufia/Hyrax. Happy to add it to the design needed list if you feel it should be higher-priority, though.

hannahfrost:
Agreed, this not MVP material, I just wanted some clarity on what you were thinking.  Eventually I think a tab with these two things would be good: (1) email address(es) for where to send the form submissions and (2) customizable text (which supports HTML links) for the message at the top of the Contact form.

hannahfrost:
@ggeisler I am renaming this ticket to "Configure contact form", flagging it as design needed, so we can make it actionable for development (though still not a top priority)

hannahfrost:
This ticket is related: https://github.com/samvera-labs/hyrax/issues/113

PR #1622, opened 2020-04-18
Contact page - customizable email to per tenant

### LINKED ISSUE #1207: Allow tenant admins to set contact form email address
As a tenant administrator, I should be able to set the email address to which contact form submissions are sent for my tenant.

orangewolf:
Fixes #1207 #517 

Split settings of default contact page email "from" and email "to".  There is a Hyrax.config.contact_email as well, but we really need both a "from" and a "to" email address and we want to be able to set both from ENV.

Override Hyrax default so that the from address isn't whatever is put in the form, as many MTAs do now allow non-specified from addresses. Instead include the contact page email in the reply-to and the subject. 

Allow changing the "to" email address on a per tenant basis. Provides an admin form for a given tenant to set where they want contact emails to be sent to. 

@samvera/hyrax-code-reviewers

Please note contribution for this PR comes from Hykuup.com and NYC's DORIS project


no-reply:
are there changes from this we'd like to port back to Hyrax, especially to help avoid the copied-code situation?

i think maybe opening up `ContactFormController` to dependency injection for its models would be a good start? `ContactFormController.model_class = MyCustomContactForm` would at least let you inherit the existing model and extend it, instead of copying it in?

orangewolf:
@no-reply yes, seems like a good call going forward.  I'm tempted to get this merged and open a ticket for it as part of the broader upgrade task set.

no-reply:
oh yeah. i definitely wouldn't delay this on upstream improvements. just wanting to set you up for a cleaner solution down the road.

jeremyf:
Approved, but I don't want to merge as I won't have space to further help out.

PR #1623, opened 2020-04-20
Fix CollectionThumbnailPathService

bkiahstroud:
## As a point of clarification: 

All references to "default collection images" or the like within this PR refer to theming default images that are uploaded under the Appearance Dashboard tab, _not_ the Hyrax default collection image. 

# Summary

The `Hyrax::CollectionThumbnailPathService` was attempting to shove the entire `Hyrax::AvatarUploader` object into the `thumbnail_path_ss` Solr field: 

```ruby
RSolr::Error::Http: RSolr::Error::Http - 400 Bad Request
Error parsing JSON field value. Unexpected OBJECT_START at [513], field=thumbnail_path_ss
```

# Expected Behavior

`Collection` creation should no longer fail in production environments. Default collection images should be applied to newly created `Collections` (and `AdminSets`) successfully and resize appropriately. 

# Additional Info 

1.  The determining factor of whether this bug occurred or not was whether a Tenant had a default collection image or not. 

1.  This error was **not** reproducible locally; attempting to create `Collections` before applying the fix did not result in the creation failing catastrophically. Instead, the only noticeable difference was whether the `thumbnail_path_ss` Solr field was set or not. This did, however, cause visual bugs:

Screen Shot 2020-04-15 at 2 39 51 PM

@samvera/hyrax-code-reviewers

Please note contribution for this PR comes from the PALNI/PALCI project.

### DIFF
--- a/app/services/hyrax/collection_thumbnail_path_service.rb
+++ b/app/services/hyrax/collection_thumbnail_path_service.rb
@@ -4,7 +4,9 @@ module Hyrax
   class CollectionThumbnailPathService < Hyrax::ThumbnailPathService
     class << self
       def default_image
-        Site.instance.default_collection_image&.url&.presence || ActionController::Base.helpers.image_path('collection.png')
+        hyrax_default_collection_image = ActionController::Base.helpers.image_path('collection.png')
+
+        Site.instance.default_collection_image&.url&.presence || hyrax_default_collection_image
       end
     end
   end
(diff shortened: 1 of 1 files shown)


app/services/hyrax/collection_thumbnail_path_service.rb
```diff
@@ -4,7 +4,9 @@ module Hyrax
   class CollectionThumbnailPathService < Hyrax::ThumbnailPathService
     class << self
       def default_image
-        Site.instance.default_collection_image.presence || ActionController::Base.helpers.image_path('collection.png')
+        hyrax_default_collection_image = ActionController::Base.helpers.image_path('collection.png')
+
+        Site.instance.default_collection_image&.url&.presence || hyrax_default_collection_image
```
dewey4iv, line 9: I don't think we need that last `&` there. 

```suggestion
        Site.instance.default_collection_image&.url.presence || hyrax_default_collection_image
```

PR #1632, opened 2020-06-23
Upgrade to Hyrax 2.8.0

cjcolvar:
Upgrade Hyrax to 2.8.0 and other dependencies including a full rebuild of Gemfile.lock.

Also included is added support for ruby 2.7 via changes in hyrax 2.8.0.  Upgrade of secure_headers also happens here so this would supersede #1611.  Had to copy in some Hyrax fixtures for feature tests (see https://github.com/samvera/hyrax/issues/4393)

@samvera/hyku-code-reviewers


orangewolf:
this is looking really solid @cjcolvar let me know when its ready to come down from wip and I'll pull it locally for testing

orangewolf:
I feel like doing these things together makes sense and saves work even through in other scenarios it would be better to split them apart. I like the idea of this + the user stuff i'm trying to get merged being out as a Hyku 3.0 and then iterating from there with additive features for a bit.

cjcolvar:
It looks like Hyrax 2.8 is about to be released with Rails 5.2 support so I think it makes sense to do that in this PR and get off the branch.  I'll do that as soon as the release happens.

Gemfile
```diff
@@ -85,7 +85,8 @@ gem 'willow_sword', git: 'https://github.com/notch8/willow_sword.git'
 gem 'blacklight', '~> 6.7'
 gem 'blacklight_oai_provider', '~> 6.0'
 
-gem 'hyrax', '~> 2.5.1'
+gem 'hyrax', github: 'samvera/hyrax', branch: 'release-2.8'
```
no-reply, line 88: this branch may now be deleted. use `~> 2.8.0` instead?

it may also be okay to use `~> 2.8`, if the various adopters are comfortable with that.

cjcolvar:
@orangewolf Are you able to take a look at this again soon?  I'm really keen on bringing over features from Ubiquity's fork but would like to base them on this PR.

spec/features/create_image_spec.rb
```diff
@@ -51,8 +51,8 @@
       expect(page).to have_content "Add files"
       expect(page).to have_content "Add folder"
       within('span#addfiles') do
-        attach_file("files[]", "#{Hyrax::Engine.root}/spec/fixtures/image.jp2", visible: false)
-        attach_file("files[]", "#{Hyrax::Engine.root}/spec/fixtures/jp2_fits.xml", visible: false)
+        attach_file("files[]", File.join(fixture_path, 'hyrax', 'image.jp2'), visible: false)
+        attach_file("files[]", File.join(fixture_path, 'hyrax', 'jp2_fits.xml'), visible: false)
```
orangewolf, line 55: keeping these locally is a good call

PR #1639, opened 2020-07-29
volume for tmp/cache significantly speeds up dynamic asset building on windows/mac and some linux file systems

orangewolf:
Anything that makes a lot of small files on a bind mounted directory can cause slowness in certain Docker environments such as Docker.app for Windows or Mac and some Linux file system types.  Dynamically compiling the assets is exactly the sort of small file IO that triggers this slowness. By moving the tmp/cache directory to a volume we significantly speed up page upload, especially when editing CSS, JS or view files (think 1 minute pages loads down to < 15s). 

@samvera/hyku-code-reviewers

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ volumes:
   zkconf:
   app:
   assets:
+  cache:
 
 networks:
   internal:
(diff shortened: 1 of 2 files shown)


orangewolf:
I only did not do this because there was no yarn.lock or package.json when I looked

cjcolvar:
That makes sense.  I'm a little puzzled though since I was expecting one given that the UV is being pulled in by yarn now.

PR #1640, opened 2020-07-30
Rubocop fixes

orangewolf:
This PR pulls in a bunch of Rubocop fixes from HykuUp's code base. Doing this as a separate PR since one gets a little bored looking at the same magic comment in 252 files. The only interesting bits (in my opinion) are the spec/support file changes.  There are two things (one removed file and one removed section in `multitenancy_metadata.rb` that massively sped up our spec runs and did not appear to be doing anything.  See https://github.com/samvera/hyku/pull/1640/commits/fef25caf4b71cce5d2f82bc4cc2e3db340e31772 to look at just that part.

@samvera/hyku-code-reviewers


app/models/endpoint.rb
```diff
@@ -1,3 +1,9 @@
+# frozen_string_literal: true
+
 class Endpoint < ApplicationRecord
   has_one :account
+
+  def switchable_options
```
cjcolvar, line 6: Can you explain this change more?  Why is this needed?

app/forms/hyrax/forms/admin/appearance.rb
```diff
@@ -259,6 +259,7 @@ def self.customization_params # rubocop:disable Metrics/MethodLength
         end
 
         def font_import_body_url
+          headline = headline_font.split('|').first.to_s.tr(" ", "+")
```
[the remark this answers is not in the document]
orangewolf: agreed. looks like accidental copy-pasta

app/models/endpoint.rb
orangewolf: That's the problem with these things being "out" on their own branches so long. I vaguely remember this code.  Let's see what past Rob had to say:

```
commit 825029de758cffaa2425674b88e96eaa91ffbb05
Author: Rob Kaufman <rob@notch8.com>
Date:   Thu Feb 21 12:15:07 2019 -0800

    some endpoints have empty options, clearing (for example) the fedora url, which then causes the process to get and stay stuck
```

I think a code comment to that affect is worth while.

orangewolf: we were seeing that if you had a blank solr or fedora url you could basically cause passenger processes to stop responding. that seems like it was worth fixing and trying to future proof a little.

orangewolf:
Being in a state where more of Hyku is Rubocop compliant makes the jump to current Bixby feel pretty achievable.

app/models/endpoint.rb
cjcolvar: :+1: Thanks!

PR #1644, opened 2020-08-18
Remove guard clause to ensure switch callbacks work

cjcolvar:
Fixes issue where switch callbacks weren't firing in rails console and possibly other cases like activejobs and single tenants.

@samvera/hyku-code-reviewers


orangewolf:
going to pull this in - someones coverage algorithm is a little aggressive!

PR #1648, opened 2020-09-18
Adds jobs that run at Midnight everyday to expire embargos and leases

BertZZ:
As part of Ubiquity Press' Hyku alignment of features developed for the British Library, this PR makes use of new Hyrax Jobs I authored for  [embargo expiry](https://github.com/samvera/hyrax/blob/master/app/jobs/embargo_expiry_job.rb) and [lease expiry](https://github.com/samvera/hyrax/blob/master/app/jobs/embargo_expiry_job.rb)

- Bump Redis version to 4 in docker-compose
- Point Hyrax at '2.x-stable' branch for now
- Add sidekiq-cron to schedule jobs
- Add jobs to auto expire embargoes and leases 

@samvera/hyku-code-reviewers


orangewolf (changes requested):
@BertZZ and I spoke about this in slack. Great progress here, but it looks like this is missing the multi-tenant piece and needs that before its ready for final review. We also agreed to move to a non-sidekiq specific job scheduler to continue to allow users to swap out the ActiveJob backend.

orangewolf (dismissed):

orangewolf (approved):

orangewolf:
@BertZZ did a merge, please give it a final look and hit merge if you feel good about it =-)

BertZZ:
@orangewolf Looks like Circle is unhappy about something but it seems all the relevant tests passed so I am happy if you are :smiley:

orangewolf:
@BertZZ i agree. There is no clear reason why it’s red.

PR #1649, opened 2020-09-24
Registered Users cannot create Works unless given explicit permission

bkiahstroud:
Restrict access to who can create Works.

This change means that registered Users will not be able to create Works unless given explicit permission to do so. For example, a User who is an admin of Tenant A, but not of Tenant B, will no longer be able to deposit Works into Tenant B. 

* Remove [ability logic that allows all registered users to create curation concerns](https://github.com/samvera/hyrax/blob/b9cc7744f37875c7b7762a01deaa53c7ae5a889f/app/models/concerns/hyrax/ability.rb#L96-L101) 

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -5,7 +5,6 @@ class Ability
   include Hyrax::Ability
 
   self.ability_logic += %i[
-    everyone_can_create_curation_concerns
     group_permissions
     superadmin_permissions
   ]
--- a/spec/features/create_work_spec.rb
+++ b/spec/features/create_work_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe 'Creating a new Work', :clean do
-  let(:user) { create(:user) }
+  let(:user) { create(:admin) }
 
   before do
     AdminSet.find_or_create_default_admin_set_id
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -13,10 +13,12 @@ RSpec.describe Ability do
     it { is_expected.not_to be_able_to(:manage, :all) }
   end
 
-  describe 'an ordinary user' do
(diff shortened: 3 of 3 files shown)


cjcolvar (commented):
Thanks for the quick PR!  As I read this the simple workaround to the issue of users in one tenant being able to deposit in another tenant is to disable self-deposit by default.  This makes out of the box hyku better by requiring implementations that want self-deposit to come up with their own solution to the issue.  I wonder how many implementations rely on self-deposit (:raised_back_of_hand: we do) and it might be a blocker for institutions wanting to migrate away from systems that already do self-deposit.

I wonder if there is a different approach that preserves self-deposit while still closing this issue.  It seems to me like the issue is how to know which users "belong" to a tenant or not in order to grant them deposit ability (and the ability to view "Institution"-only visible works/files which is the other issue we raised with elevated users).

One way could be restricting self-registration to email addresses with specific domains then we could grant the ability to self-deposit based upon the same email domain check.  But I don't think such a registration restriction would be globally accepted.

Maybe we could grant roles in the tenant(s) that a user registers with instead of just signing into?  But I think this would require changes to the registration process since registration is global and a user would get an error if trying to register in host B when already registered in host A.

Or maybe forgetting about "belonging" to a tenant and instead relying on mediated deposit workflow, but this might leave things open to spamming and misleading users to think they are allowed to deposit in a tenant when they might not really be.  It would also put more work on admins to filter deposit requests.

No matter which approach we take, my goal is for a flow that doesn't burden admins or raise barriers to self-deposit.  What do others think?  Any different ideas to solve this issue?

bkiahstroud:
@cjcolvar Thanks for your detailed review! I feel like I have a better understanding of what the problem is now. We're going to meet with Amanda and Gretchen from PALNI/PALCI to get a better understanding of what their concerns were.

PR #1650, opened 2020-10-12
Helps with Solr sometimes falling over in dev/test

orangewolf:
This change makes solr run more reliably in our specs and was recommended in dev chat with the folks working on the Solr docker image.

@samvera/hyku-code-reviewers

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,10 @@ services:
   solr:
     image: solr:7.5
     command: solr -c -f -z zoo1:2181
+    environment:
+      - OOM=script
+      - VIRTUAL_PORT=8983
+      - VIRTUAL_HOST=solr.hyku.docker
     depends_on:
       zoo1:
         condition: service_healthy


docker-compose.yml
```diff
@@ -38,6 +38,10 @@ services:
   solr:
     image: solr:7.5
     command: solr -c -f -z zoo1:2181
+    environment:
+      - OOM=script
+      - VIRTUAL_PORT=8983
+      - VIRTUAL_HOST=solr.hyku.docker
```
cjcolvar, line 44: Interesting.  Totally an edge case right now, but I wonder if we need to add a reserved name check for creating tenants to avoid collisions with pre-existing subdomains that were manually created like this one.

cjcolvar (approved):
:+1: 
Do you think a restart policy for solr should be added to this docker-compose?  [This comment](https://github.com/docker-solr/docker-solr/issues/217#issuecomment-499733283) on solr docker leads me to think we should:
>So far everything looks good - the solr-foreground script now set's the OOM handling and the oom_solr.sh script successfully logs the error and kills the process.
Together with the proper docker restart policy (always, on-failure or unless-stopped) the container is restarted and solr stays responsive.

bess (approved):

docker-compose.yml
orangewolf: good point. I'll open a separate PR for it

orangewolf:
> 👍
> Do you think a restart policy for solr should be added to this docker-compose? [This comment](https://github.com/docker-solr/docker-solr/issues/217#issuecomment-499733283) on solr docker leads me to think we should:
> 
> > So far everything looks good - the solr-foreground script now set's the OOM handling and the oom_solr.sh script successfully logs the error and kills the process.
> > Together with the proper docker restart policy (always, on-failure or unless-stopped) the container is restarted and solr stays responsive.

doesn't seem to need to be? it seems like solr gets restarted in the container. not sure if that comment lines up with the version of solr we are using.

PR #1652, opened 2020-11-13
Upgrade to Redis v5

bkiahstroud:
Sidekiq requires a newer version of Redis (>= 4), otherwise the workers container falls over locally and background jobs don't get processed. 

* Upgrade Redis to major version 5 
* Update sidekiq/redis gem/redis-namespace to fix `#exists` vs `#exists?` issue 

@samvera/hyku-code-reviewers 

 This code was provided to Hyku by PALNI/PALCI.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -776,8 +776,8 @@ GEM
       rdf (~> 3.1)
     redic (1.5.3)
       hiredis
-    redis (4.2.1)
-    redis-namespace (1.7.0)
+    redis (4.2.2)
+    redis-namespace (1.8.0)
       redis (>= 3.0.4)
     redlock (1.2.0)
       redis (>= 3.0.0, < 5.0)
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
This overlaps with some of the changes in #1648, but I like how this PR has a single focus.  If this gets merged then #1648 will need to be updated even though it has already been approved.
  
I don't see the version of `redis-namespace` changing in `Gemfile.lock` as mentioned in the description.  Did it not actually need to be upgraded?  For reference, #1648 had it upgrading from 1.7.0 -> 1.8.0.

bkiahstroud:
> This overlaps with some of the changes in #1648, but I like how this PR has a single focus. If this gets merged then #1648 will need to be updated even though it has already been approved.
> 
> I don't see the version of `redis-namespace` changing in `Gemfile.lock` as mentioned in the description. Did it not actually need to be upgraded? For reference, #1648 had it upgrading from 1.7.0 -> 1.8.0.

Whoops, I was missing a commit, thanks @cjcolvar

cjcolvar (approved):
:+1: Looks good to me.

PR #1653, opened 2020-12-03
Prevent users who do not have roles on a given tenant from being in the registered group automatically

orangewolf:
- Do not show institution only works to users who do not have at least one role on the tenant in question.
- Do not allow creation of works unless the user has at least one role on the tenant in question.

My understanding is these are the 2 blockers for the 3.0 release.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -42,4 +42,14 @@ class Ability
   def superadmin?
     current_user.has_role? :superadmin
   end
+
+  # Override from blacklight-access_controls-0.6.2 to define registered to include having a role on this tenant
+  def user_groups
+    return @user_groups if @user_groups
+
+    @user_groups = default_user_groups
+    @user_groups |= current_user.groups if current_user.respond_to? :groups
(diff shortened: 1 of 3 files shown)


app/models/ability.rb
```diff
@@ -42,4 +42,14 @@ def superadmin_permissions
...
     current_user.has_role? :superadmin
   end
+
+  # Override from blacklight-access_controls-0.6.2 to define registered to include having a role on this tenant
+  def user_groups
+    return @user_groups if @user_groups
+
+    @user_groups = default_user_groups
+    @user_groups |= current_user.groups if current_user.respond_to? :groups
+    @user_groups |= ['registered'] if (!current_user.new_record? && current_user.roles.count > 0)
```
orangewolf, line 52: lots of tracking down, one tiny little change.  The original method is found [here](https://github.com/projectblacklight/blacklight-access_controls/blob/master/lib/blacklight/access_controls/ability.rb#L104) `        @user_groups |= ['registered'] unless current_user.new_record?
`

jeremyf (approved):

PR #1654, opened 2020-12-03
Subdomains to exclude

orangewolf:
Do not allow solr, admin, www or fedora as subdomain names.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -48,7 +48,8 @@ class Account < ApplicationRecord
             unless: :cname_is_blank?
 
   validates :tenant, presence: true, uniqueness: true
-  validates :cname, presence: true, uniqueness: true, exclusion: { in: [default_cname('')] }
+  validates :cname, presence: true, uniqueness: true,
+                    exclusion: { in: Account.excluded_cnames.map { |c| default_cname(c) } }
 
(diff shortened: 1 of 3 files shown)


orangewolf:
Seems like a load order issue. It ran on locally but I wonder if spring had already loaded apartments.

cjcolvar:
This seems like a really good idea!  I'm happy to approve once CI is fixed.

app/models/account.rb
```diff
@@ -48,7 +48,8 @@ def self.tenants(tenant_list)
             unless: :cname_is_blank?
 
   validates :tenant, presence: true, uniqueness: true
-  validates :cname, presence: true, uniqueness: true, exclusion: { in: [default_cname('')] }
+  validates :cname, presence: true, uniqueness: true,
+                    exclusion: { in: Account.excluded_cnames.map { |c| default_cname(c) } }
```
cjcolvar, line 52: Tests are failing because `excluded_cnames` isn't defined at this point.  Does it need to be a lambda or proc?  Or maybe changing `excluded_cnames` into a constant or just defining it before this `validates`?

orangewolf:
closed in lew of https://github.com/samvera/hyku/pull/2295

PR #1655, opened 2020-12-04
Test institution visible works with elevated user model

cjcolvar:
The changes made in #1653 restict access to avoid users from other
tenants from accessing institution visible works (which should only be
visible to users of the tenant) by requiring the user to have a role on
the tenant.  This test attempts to cover the cases for access to
institution visible works: end users from within and outside the work's
account as well as users with roles within and outside the work's
account.

This PR is for testing and discussion purposes only.

### DIFF
--- a/spec/requests/institution_visibility_spec.rb
+++ b/spec/requests/institution_visibility_spec.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Insitution visiblity work access', type: :request, clean: true, multitenant: true do
+  let(:account) { create(:account) }
+  let(:account2) { create(:account) }
+  let(:tenant_user) { create(:user) }
+  let(:tenant2_user) { create(:user) }
+  let(:work) { create(:work, visibility: 'authenticated') }
+
+  before do
+    WebMock.disable!
+    Apartment::Tenant.create(account.tenant)
+    Apartment::Tenant.switch(account.tenant) do
+      Site.update(account: account)
+      tenant_user
+      work
+    end
+    Apartment::Tenant.create(account2.tenant)
+    Apartment::Tenant.switch(account2.tenant) do
+      Site.update(account: account2)
+      tenant2_user
+    end
+  end
+
+  after do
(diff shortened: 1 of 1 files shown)


spec/requests/institution_visibility_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    Apartment::Tenant.create(account2.tenant)
+    Apartment::Tenant.switch(account2.tenant) do
+      Site.update(account: account2)
+      tenant2_user
+    end
+  end
+
+  after do
+    WebMock.enable!
+    Apartment::Tenant.drop(account.tenant)
```
orangewolf, line 27: is there a reason to drop tenant 1 but not tenant 2?

spec/requests/institution_visibility_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  describe 'as an end-user' do
+    it 'allows access for users of the tenant' do
+      login_as tenant_user, scope: :user
+      get "http://#{account.cname}/concern/generic_works/#{work.id}"
+      expect(response.status).to eq(200)
```
orangewolf, line 34: I do not think this really matches what we should expect. I think the user should have to have a role. however, I think it makes sense to go ahead and expand this spec to cover signing the users up in each tenant instead of trying the declare trick on lines 21 and 15.  It will still fail "as is" but that is a spec that I feel like should pass.

spec/requests/institution_visibility_spec.rb
cjcolvar: Good catch.  Tenant 2 should be dropped here.

spec/requests/institution_visibility_spec.rb
cjcolvar: That makes sense.  I'll switch to user registration through requests instead of test setup.

cjcolvar:
@orangewolf Just saw that you took care of these changes and pulled the test into #1656.  Thanks!

PR #1656, opened 2020-12-05
Institution visibility with roles

orangewolf:
Add a role to users when they register so that they can see institutional works. Also fleshed out the specs @cjcolvar started a little more.


orangewolf:
@cjcolvar this should be ready for your review again now.  Specs are green locally, though I see the feature spec timing issue got me on the last spec run. Hopefully this one will be green.

cjcolvar (approved):
This is nice and clean!  I think this solves the issue with institution visibility works!
It seems to me that future work would be providing a way for a user who was created when registering with one tenant a way to register with another tenant.  But that can come later since it is a new feature.

orangewolf:
I agree Chris - at the moment you have to be invited, and that can be improved.

PR #1657, opened 2020-12-21
Use latest released 2.x Hyrax

cjcolvar:
Use an official release and force hyrax to 2.9.1 or greater to pull in https://github.com/samvera/hyrax/pull/4660 and https://github.com/samvera/hyrax/pull/4656 along with the auto-expiry jobs (https://github.com/samvera/hyrax/pull/4526)
@samvera/hyku-code-reviewers


orangewolf (approved):

orangewolf:
this is actually a really small change, since we already had the autoexpiry jobs. looks good to me.

PR #1658, opened 2021-01-19
set version to 3.0 and fix base build

orangewolf:
This sets the version number to 3.0 and makes sure the bundler version matches the Gemfile.lock both now and in the future as needed to refresh the Docker hub images.


cjcolvar (approved):
:tada:

PR #1659, opened 2021-01-29
Use specific work type presenter since it is a subclass of Hyku::WorkShowPresenter

cjcolvar:
Fixing this allows for Hyrax plugins to work as expected when injecting module mixins into the normal Hyrax presenters.

See #1646 and https://github.com/samvera/hyku/blob/master/app/presenters/hyrax/image_presenter.rb#L6
@samvera/hyku-code-reviewers


orangewolf (approved):

orangewolf:
somebody explain to me how this reduces code coverage? because yay - it doesn't.

PR #1663, opened 2021-02-08
Update instructions to include starting background workers

cjcolvar:
@samvera/hyku-code-reviewers


orangewolf (approved):
I did not know `up` could take 2 arguments! that's a great way to solve that problem.

PR #1665, opened 2021-02-23
Use label tag to allow accessibly access

prdanelli:
By default it seems that the checkboxes on the work type select page are not wrapped in labels and thus far harder to select than it they were. This mini-PR adds a label around each checkbox. 

@samvera/hyku-code-reviewers


orangewolf:
this is great - thank you

PR #1666, opened 2021-02-23
Fix zookeeper bundle install

cjcolvar:
Fixes the CircleCI build which is failing currently.  See https://app.circleci.com/pipelines/github/samvera/hyku/323/workflows/74d691c6-f883-4db9-af65-28ac6f8302aa/jobs/352

@samvera/hyku-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,8 @@ jobs:
       RACK_ENV: test
       FCREPO_TEST_PORT: 8080/fcrepo
       SPEC_OPTS: --profile 10 --format RspecJunitFormatter --out /tmp/test-results/rspec.xml --format progress
+      # CFLAGS need to be set for zookeeper to compile native extensions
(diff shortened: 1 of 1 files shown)


orangewolf (approved):

cjcolvar:
@orangewolf How worried are you about the failing tests?  It looks like they passed on the rerun but the build still failed.  Do you think we could go ahead and merge this or should I restart the build until it goes green?

orangewolf:
the failure here is coveralls - I vote we change the way coverage is done because that's been pretty flakey. do you know if its flakey on Hyku or is it just flakey everywhere?

orangewolf:
@cjcolvar separately there is some flapping feature specs. We're working on a project right now to try and nail down what is causing those, but its super time consuming.

cjcolvar:
@orangewolf I think Coveralls is just flakey everywhere.  I could make a PR to switch to codeclimate if that sounds good to you.

PR #1670, opened 2021-03-27
Use Hyrax docker base setup

no-reply:
Hyrax now provides a minimal base docker image, intended to be reused and
extended by applications. this is a proposal to replace the existing Dockerfile
with one based on that image.

this would switch the application server to `puma`. our thinking is that real
containerized deployments (e.g. via the Helm chart) are likely to be behind a
load balancer and an `nginx` (an ingress controller).

the image is kept minimal with the idea that a smaller set of system
dependencies means a lower footprint for security risks, besides the bandwidth
and storage convenince. Hyrax's setup definitely has space to improve here, but
the idea is greater consolidation will help the community maintain something
really solid.

there's still a little work to do here:

  - ~~remove dependency on `zk` (and therefore java runtime).~~
    - this now removes `zk` as a build and runtime dependency. a slightly reworked copy of Hyrax's SolrCloud scripts is included for handling configset uploads and collection management in CI. if we could move the CI image to use the built container, we could remove those scripts.
  - refine & use Hyrax's `hyrax-worker` image setup for `sidekiq` workers (using
    a separate lets us consider optimizations like `MALLOC_ARENA_MAX=2`).
  - systematically reconcile environment variables (the job here is a bit
    patchwork).

if this is looking like a good direction, I can work on getting this working
with the upstream helm chart as well.

@samvera/hyku-code-reviewers


no-reply:
oh, and there's `docker-compose.production.yaml` to contend with. is this genuinely supported for running a production environment? like for docker swarm? or is it just cruft from the grant work days?

no-reply:
sort of surprisingly (to me) the image built by this Dockerfile deploys successfully with Hyrax's helm chart, all the way up to a Rails server start! the health check is misconfigured, since the root on `localhost`/`0.0.0.0` comes up 404 but i suspect there are some simple adjustments to make to clear that.

this is a much easier deployment process than i expected!

orangewolf:
> oh, and there's `docker-compose.production.yaml` to contend with. is this genuinely supported for running a production environment? like for docker swarm? or is it just cruft from the grant work days?

We used this for Rancher 1 based deployment, which is completely deprecated. I do not think this is correct for docker swarm, so i would vote to remove it. It feels safe in 2021 to recommend k8s based deployment

orangewolf:
> sort of surprisingly (to me) the image built by this Dockerfile deploys successfully with Hyrax's helm chart, all the way up to a Rails server start! the health check is misconfigured, since the root on `localhost`/`0.0.0.0` comes up 404 but i suspect there are some simple adjustments to make to clear that.
> 
> this is a much easier deployment process than i expected!

Glad to hear it. Us working closely together has put us on a good course for consultation. That feels like exactly the kind of win we are looking for.

orangewolf:
Couple of thoughts
- moving to puma makes sense to me, but it means Rails is serving all the assets, doesn’t it? We solve that on other stacks with either varnish or nginx. I’d be happy to submit a pr for hyrax if that’s desirable. 
- I’m good with a separate sidekiq image as long as it remains easy to swap sidekiq with delayed job. I vastly prefer the dependability and easy job lookup of dj over the essentially unrealizable performance benefits (we aren’t doing 20k jobs a second, we’re doing jobs that can take minutes). I get this is a minority opinion, and am content to keep sidekiq as the default, just want the option to be doable. 
- I’m 100% in to removing the Java dependency for zk. This is only a setup step, I can confirm it won’t affect the tenancy. 
- what are your thoughts on that reconciliation of env vars? Hyku’s setting gem seems like a win to me but I’m pretty open to solutions.

no-reply:
> moving to puma makes sense to me, but it means Rails is serving all the assets, doesn’t it? We solve that on other stacks with either varnish or nginx. I’d be happy to submit a pr for hyrax if that’s desirable.

yes. Rails serving the assets is the Hyrax chart's setup at the moment. a lot of folks seem to think this is the way to go (i.e. let rails serve it, then put a CDN or cache in front), but i agree that this would be better to solve as part of the chart. moving assets and other static content to a separate pod would make it possible to scale the asset servers independently, which doesn't seem possible with either setup at the moment(?).

something quite a lot like this seems ideal, in that it avoids the need for an RWX PVC: https://webcloudpower.com/helm-rails-static-files-path/

no-reply:
> I’m good with a separate sidekiq image as long as it remains easy to swap sidekiq with delayed job. I vastly prefer the dependability and easy job lookup of dj over the essentially unrealizable performance benefits (we aren’t doing 20k jobs a second, we’re doing jobs that can take minutes). I get this is a minority opinion, and am content to keep sidekiq as the default, just want the option to be doable.

this seems like a good goal to me. Hyrax certainly doesn't want to be a `sidekiq`-only engine. i don't know of any reason to expect `hyrax-engine-dev-worker` not to work with a dj app, apart from the need for a custom `CMD`.

i'll keep charging forward in this direction and let's see what happens if we put up an app with another ActiveJob provider?

no-reply:
> I’m 100% in to removing the Java dependency for zk. This is only a setup step, I can confirm it won’t affect the tenancy.

fantastic. i do my best to keep a JRE/JDK off my laptop, and i get grumpy when `bundle` fails because of it 😾

no-reply:
> what are your thoughts on that reconciliation of env vars? Hyku’s setting gem seems like a win to me but I’m pretty open to solutions.

i feel like i might be missing something important, re: "Hyku's setting gem"? is this in reference to https://github.com/rubyconfig/config ?

i think there are a couple of things to look at:

  - there are some mismatches in (mostly arbitrary) environment names, e.g. `DATABASE_NAME` vs. `DB_NAME`. there's an opportunity to align support and deprecate some variations.
  - Hyrax's chart needs a hook for adding Hyku-specific variables. i think following the Bitnami pattern would be good: `extraEnvVars`, `extraEnvVarsCM`, `extraEnvVarsSecret`

orangewolf:
> there are some mismatches in (mostly arbitrary) environment names, e.g. DATABASE_NAME vs. DB_NAME. there's an opportunity to align support and deprecate some variations.

This is what I meant by reconcile - I don't feel attached to DATABASE_NAME vs DB_NAME. 

There are places where Hyku users env vars like `SETTINGS__SOLR__URL` and also requires `SOLR_URL` to also be set, but perhaps that is a separate problem and one for another time. I do like the rubyconfig/config gem, but I sometimes think Unix solved this in 1979 and maybe we can stop resolving it.

> Hyrax's chart needs a hook for adding Hyku-specific variables. i think following the Bitnami pattern would be good: extraEnvVars, extraEnvVarsCM, extraEnvVarsSecret

I'm 100% in on using Btinami's pattern. I don't always agree with them, but sometimes doing the same thing other people are is more important than inventing our own path. And don't you dare tell 17 year old Rob that I said that ;-)

no-reply:
after https://github.com/samvera/hyrax/pull/4877, the hyrax chart supports:

  - `extraEnvVars` for arbitrary key-value pairs
  - `extraEnvFrom` for existing ConfigMaps and Secrets that will populate env
  - `extraInitContainers` for (e.g.) the case where we need to ensure additional databases exist before startup
  - `extraVolumeMounts` and `extraVolumes` for application specific drives

orangewolf (approved):
this looks good for me. @no-reply maybe time to take the WIP off?

cjcolvar (commented):
Looks good to me.

Issue #1671, opened 2021-04-08
mimemagic locked on yanked version


no-reply:
https://github.com/samvera/hyku/blob/master/Gemfile.lock#L620

`bundle` fails due to `mimemagic` being yanked.

the best fix is probably to upgrade to Rails 5.2.5.

orangewolf:
thank you for this

PR #1672, opened 2021-04-13
bump rails to 5.2.5

no-reply:
this fixes errors related to the mimemagic GPL yank.

@samvera/hyku-code-reviewers


no-reply:
#1671

orangewolf (approved):

PR #1693, opened 2021-04-27
Docker development adjustments

orangewolf:
Started with a fresh environment today and found a few issues

- Fix app home path everywhere
- Add new docker-compose step for solving the volume file permisison issue
- Move dory domain to .test since Google is resolving .docker now
- Solve issue where apartment initializer would not allow migrations before seeds had been run

@samvera/hyku-code-reviewers


chart/README.md
```diff
@@ -11,7 +11,7 @@ brew install helm
 ```
 
 * kubernetes
-Kubectl is the command line tool for controlling Kubernetes clusters. It is available via (https://docs.docker.com/docker-for-mac/)[Docker for Mac]
+Kubectl is the command line tool for controlling Kubernetes clusters. It is available via (https://docs.test.com/docker-for-mac/)[Docker for Mac]
```
cjcolvar, line 14: Looks like a find/replace that shouldn't have been replaced.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_30_194003) do
+ActiveRecord::Schema.define(version: 2020_06_01_204556) do
```
cjcolvar, line 13: Is the rollback of the db schema version intentional here?

docker-compose.yml
```diff
@@ -108,13 +108,13 @@ services:
...
       # Comment out these 5 for single tenancy / Uncomment for multi
-      - SETTINGS__MULTITENANCY__ADMIN_HOST=hyku.docker
+      - SETTINGS__MULTITENANCY__ADMIN_HOST=hyku.test
       - SETTINGS__MULTITENANCY__ADMIN_ONLY_TENANT_CREATION=false
-      - SETTINGS__MULTITENANCY__DEFAULT_HOST=%{tenant}.hyku.docker
-      - SETTINGS__MULTITENANCY__ROOT_HOST=hyku.docker
-      - SETTINGS__MULTITENANCY__ENABLED=true
+      - SETTINGS__MULTITENANCY__DEFAULT_HOST=%{tenant}.hyku.test
+      - SETTINGS__MULTITENANCY__ROOT_HOST=hyku.test
+      - SETTINGS__MULTITENANCY__ENABLED=false
```
cjcolvar, line 115: Can you explain the switch to disable multi-tenancy by default?  Does it make sense to remove the next three lines (116-118)?

cjcolvar (commented):
I did a quick read through and this looks good so far.  I'll test it out locally and report back.

chart/README.md
no-reply:
```suggestion
Kubectl is the command line tool for controlling Kubernetes clusters. It is available via (https://docs.docker.com/docker-for-mac/)[Docker for Mac]
```

docker-compose.yml
orangewolf: I was testing single tenant mode to make sure the fix i did for the db chicken/egg issue worked. I don't think single should be the default

cjcolvar:
I'm getting errors when trying to bring this up:
```
Attaching to hyku_workers_1, hyku_web_1
workers_1         | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
web_1             | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
hyku_workers_1 exited with code 1
hyku_web_1 exited with code 1
```

no-reply:
> I'm getting errors when trying to bring this up:
> 
> ```
> Attaching to hyku_workers_1, hyku_web_1
> workers_1         | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
> web_1             | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
> hyku_workers_1 exited with code 1
> hyku_web_1 exited with code 1
> ```

is it possible you have pre-existing volumes, leftover from the older container setup? `docker-compose down -v` might clear this up?

---

@orangewolf i think it might actually be a good idea to remove the `chown` from the Hyrax entrypoint script and solve the whole thing another way. these have been problematic for us in Kubernetes, too, where they don't play very well with all CSI drivers, and `fsgroup:` settings are required either way.

no-reply:
this is working for me, and i'm approving on that basis. i'd like to see someone else succeed at running this dev env before we merge it though?

no-reply (dismissed):

orangewolf:
> > I'm getting errors when trying to bring this up:
> > ```
> > Attaching to hyku_workers_1, hyku_web_1
> > workers_1         | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
> > web_1             | chown: /app/samvera/hyrax-webapp/tmp: Operation not permitted
> > hyku_workers_1 exited with code 1
> > hyku_web_1 exited with code 1
> > ```
> 
> is it possible you have pre-existing volumes, leftover from the older container setup? `docker-compose down -v` might clear this up?
> 
> @orangewolf i think it might actually be a good idea to remove the `chown` from the Hyrax entrypoint script and solve the whole thing another way. these have been problematic for us in Kubernetes, too, where they don't play very well with all CSI drivers, and `fsgroup:` settings are required either way.

I agree with you @no-reply, we solve it again in the docker-compose.yml here, since they are mounted for performance and at least on Mac, the mounts start out as root. 

@cjcolvar  `/app/samvera/hyrax-webapp/tmp` shouldn't be mounted. Maybe the location has the wrong permissions locally somehow?

no-reply:
if you use `38aed2d0b89908b1faaf24be2f327a48e577983a` as the `HYRAX_IMAGE_VERSION`, i think the `chown` issue will be gone.

cjcolvar:
@no-reply I set `HYRAX_IMAGE_VERSION` in the first line of `Dockerfile` and did the following but still ended up with the same errors.
> docker-compose down -v
sudo rm -rf tmp/
docker-compose build
docker-compose up web workers

no-reply:
@cjcolvar it seems pretty likely that your `tmp` is setup on `root:root`.

no-reply:
@orangewolf should we have the docker-compose mount setup change permission for `-R /tmp` instead of the individual volumes?

no-reply:
i pushed the new image as `ghcr.io/samvera/hyrax/hyrax-base:3.0.1`, so the custom tag should no longer be necessary

orangewolf:
> @orangewolf should we have the docker-compose mount setup change permission for `-R /tmp` instead of the individual volumes?

It needs to be the individual volumes because -R tmp will cause substantial performance issues once you start having a lot of derivitives or uploads. Basically what happens is that the moment it takes more than 30 seconds to do the chown (which doesn't require that much data on efs/nfs) the containers get cycled for not booting faster than they can complete the ownership change and therefore never come up.

no-reply:
Would a last step to just change the ./tmp directory itself be a workaround?

I guess an alternative would be to put the pidfile elsewhere

orangewolf:
I don't understand how Chris is getting tmp w/ the wrong permissions. So I'm not sure what to recommend to fix it. Let's try the chmod there and see if that solves it for this case.

orangewolf:
@cjcolvar lets give this a shot, if it still doesn't work, I'd be happy to do a pairing session to figure it out =-)

cjcolvar (approved):
Looks good!

PR #1705, opened 2021-05-17
Support multiple domain names per tenant

orangewolf:
This PR allows a given tenant to be referred to by multiple domain names. For example samvera-repo.hykuup.com can be the same tenant as repo.samvera.org. This PR does not currently change the UI for adding or removing domain names. That will come in a later PR. It requires a migration but does not require a manual intervention on a tenant basis. . This work is also a precursor for a `is_public` feature that will allow tenants to be easily hidden behind HTTP auth for demos and pre-live accounts.

@samvera/hyku-code-reviewers


orangewolf:
this is only failing do to a flaky feature spec - not due to any changed code here. it passes locally.

app/middleware/account_elevator.rb
```diff
@@ -11,7 +11,8 @@ def parse_tenant_name(request)
   end
 
   def self.switch!(cname)
-    account = Account.find_by(cname: Account.canonical_cname(cname))
+    account = Account.joins(:domain_names).find_by(domain_names: { is_active: true,
```
smcfarlane, line 14: you can use the `from_cname` class method here

smcfarlane (commented):
Just commenting as I don't have a lot of context on Hyku but it looks good to me

app/middleware/account_elevator.rb
orangewolf: @smcfarlane good point - updated

bkiahstroud (approved):

PR #1710, opened 2021-05-20
update bulkrax

no-reply:
ran `bundle update bulkrax`.

the prior target version for `bulkrax` fails repeated migrations, which motivated this change.

@samvera/hyku-code-reviewers


cjcolvar (approved):
:+1:

PR #1711, opened 2021-05-21
Docker and Helm File Path Issues

orangewolf:
- branding images should be served from public - these currently do not render and break the app when banner image or custom default icons are set
- fix for docker file permissions
- keep x-app from starting when docker-compose up is run (since it does not do anything)


docker-compose.yml
```diff
@@ -1,5 +1,24 @@
 version: '3.8'
 
+x-app: &app
+  build:
+    context: .
+    target: hyku-base
+    args:
+      - EXTRA_APK_PACKAGES=less vim bash # TODO: this doesn't install vim, why?
+  image: registry.gitlab.com/notch8/britishlibrary:${TAG:-latest}
```
cjcolvar, line 9: I'm guessing this should still be old image:
```suggestion
  image: ghcr.io/samvera/hyku:latest
```

docker-compose.yml
```diff
@@ -1,5 +1,24 @@
...
+  env_file:
+    - .env
+  # NOTE: all common env variables moved to .env
+  volumes:
+    - uploads:/app/samvera/hyrax-webapp/tmp/uploads
```
cjcolvar, line 14: Should this be switched to public too?
```suggestion
    - uploads:/app/samvera/hyrax-webapp/public/uploads
```

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_30_194003) do
+ActiveRecord::Schema.define(version: 2020_06_01_204556) do
```
cjcolvar, line 13: Is this rollback intentional?

cjcolvar (commented):
Is there something more than the movement of the `x-app` declaration that keeps it from starting with `docker-compose up`?

docker-compose.yml
```diff
@@ -1,5 +1,24 @@
 version: '3.8'
 
+x-app: &app
+  build:
+    context: .
+    target: hyku-base
+    args:
+      - EXTRA_APK_PACKAGES=less vim bash # TODO: this doesn't install vim, why?
```
no-reply, line 8: `hyku-base` doesn't seem to accept a `EXTRA_APK_PACKAGES` argument(?)

orangewolf: it does not. I haven't looked in to it at all yet - do we have to do something to pass it back up a level?

orangewolf:
> Is there something more than the movement of the `x-app` declaration that keeps it from starting with `docker-compose up`?

no, the act of moving it is what makes it get skipped when you do `docker-compose up` there was talk of respecting the `x-` convention any where in the file and I thought that was in Docker Compose 3.8, but it clearly is not.

db/schema.rb
orangewolf: reverted

docker-compose.yml
orangewolf: figured it out - they do not stack at all, so I added support

db/schema.rb
cjcolvar: I see the version was reverted but `is_public` is still being removed from the accounts table.

no-reply (approved):

PR #1712, opened 2021-05-24
Don't try to assign registered role on global tenant

cjcolvar:
On a fresh checkout of hyku, I was not able to register a user on the global tenant because the
before_create callback errored when trying to add the registered role
due to a validation error on the role: `Resource type is not included in the list`
This was not caught in the tests because there wasn't a check that the
user actually got created.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -73,8 +73,9 @@ class User < ApplicationRecord
   # If this user is the first user on the tenant, they become its admin
   # unless we are in the global tenant
   def add_default_roles
-    add_role :admin, Site.instance unless
-      self.class.joins(:roles).where("roles.name = ?", "admin").any? || Account.global_tenant?
+    return if Account.global_tenant?
+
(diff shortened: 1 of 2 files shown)


orangewolf:
guess I'll cross this off my list of things to get committed back from last week https://share.getcloudapp.com/RBul5JrP =-D

Also, i didn't see it before, but I think its the line feed in the original that is the problem. the `unless` would have a nil after it and the next line would evaluate true or false but not return to anything.  Your way is cleaner and (obv) less error prone.

orangewolf:
guess I'll cross this off my list of things to get committed back from last week https://share.getcloudapp.com/RBul5JrP =-D

Also, i didn't see it before, but I think its the line feed in the original that is the problem. the `unless` would have a nil after it and the next line would evaluate true or false but not return to anything.  Your way is cleaner and (obv) less error prone.

orangewolf (approved):

PR #1713, opened 2021-05-25
Allow per-tenant settings by recreating the Settings on switch!

cjcolvar:
This is a bit more complicated due to the fact that the config gem
removes and sets the Settings constant when reloading from files.  This
isn't thread safe as all threads share the same constant.  I was also
seeing one thread trying to access Settings when another had just
removed it leading to NameErrors.

My approach to dealing with this is to remove the Settings constant in
an initializer then access the configuration through a Settings
singleton.  The Settings class delegates all calls to configuration
stored in the current thread which is created avoiding setting the
Settings constant.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -104,6 +104,7 @@ class Account < ApplicationRecord
     solr_endpoint.switch!
     fcrepo_endpoint.switch!
     redis_endpoint.switch!
+    Settings.switch!(name: name, settings: additional_settings)
   end
 
   def switch
@@ -117,6 +118,7 @@ class Account < ApplicationRecord
     SolrEndpoint.reset!
     FcrepoEndpoint.reset!
     RedisEndpoint.reset!
+    Settings.switch!
   end
 
   # Get admin emails associated with this account/site
@@ -148,5 +150,10 @@ class Account < ApplicationRecord
 
     def canonicalize_cname
       self.cname &&= self.class.canonical_cname(cname)
+    end
+
+    # Allow for overriding this method to inject new settings from DB or elsewhere
+    def additional_settings
+      {}
     end
 end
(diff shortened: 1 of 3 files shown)


orangewolf:
@cjcolvar i know you’ve put quite a bit of effort in to this already and if it’s better to push through, feel free to ignore me. 

With that disclaimer out of the way, I feel like you might be making things way harder than they have to be. It would seem to me that putting these properties on the account or site object would greatly simplify things. This would leave Settings for globals and also make it trivial to allow changing these things in the UI (since they are just active record attributes at that point) If we have (for example) a google analytics code in a yaml file then we have to deploy and reload a 100 tenant application just to pick up the new code? That doesn’t seem right. We could also (down the road) encrypt those columns more easily than encrypting the yaml file.
 current_account is already loaded on each request so you aren’t even making a new db call. Just adding some properties to it. 
 that doesn’t mean I’ll stand in the way of this if you want to move forward with it, just offering a perspective.

orangewolf:
I think Account should be the place for anything that's per tenant. Global is seperate and Settings makes sense to me for that _though_ consolidating down and combining with Hyrax.config would not break my heard either (in order to reduce the number of places even further)

cjcolvar:
To make sure I understand you, you'd like to keep `Settings` for only global (superadmin) configuration and then anything tenant-specific would be stored in the Account model and then accessed via `Hyrax.config`? Or should the Account itself be the main point of accessing tenant-specific config (although some would still have to be available through the Hyrax config like GA id to meet Hyrax's needs)?

cjcolvar:
After discussing with @orangewolf, we decided this should probably be done with a different approach which is spelled out in #1719 .

PR #1715, opened 2021-06-21
Update README.md

archivalGrysbok:
Ran into an issue running docker-compose up, since my user wasn't in the 'docker group'. Added a note to the readme about this requirement, to help save others a google.

@samvera/hyku-code-reviewers


orangewolf:
thank you!

PR #1720, opened 2021-07-02
set default scope for users to exclude guests, remove old guests

### LINKED ISSUE #606: Cron job to run the rake task to prune guest users
https://github.com/cbeer/devise-guests/blob/master/lib/railties/devise_guests.rake#L2-L5

orangewolf:
Fixes #606 

Make sure we do not display guest users in the UI. These are created and used for Blacklight sessions, but have no value to admins or other logged in users of the system. Also prune guest users periodically.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/jobs/delete_old_guests_job.rb
+++ b/app/jobs/delete_old_guests_job.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class DeleteOldGuestsJob < ApplicationJob
+  non_tenant_job
+  after_perform do |job|
+    reenqueue
+  end
+
+  def perform
+    User.where("guest = ? and updated_at < ?", true, Time.now - 7.days).each { |x| x.destroy}
+  end
+
+  private
+
+  def reenqueue
+    DeleteOldGuestsJob.set(wait_until: Date.tomorrow.midnight).perform_later
+  end
+end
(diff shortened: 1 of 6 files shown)


app/jobs/delete_old_guests_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class DeleteOldGuestsJob < ApplicationJob
+  non_tenant_job
+  after_perform do |_job|
+    reenqueue
+  end
+
+  def perform
+    User.where("guest = ? and updated_at < ?", true, Time.current - 7.days).each(&:destroy)
```
prdanelli, line 10: Does the version of Rails you're targetting allow you to use `destroy_all` here?

orangewolf: great question @prdanelli I'm working on fixing the broken spec right now. I'll try it out.  I admit copied the blacklight code from its rake task and didn't look at it too closely.

orangewolf:
@prdanelli can I get a re-review? I think this is GTG now but would like you to see it first if possible

prdanelli:
Looks good to me @orangewolf.

prdanelli (approved):

PR #1724, opened 2021-07-15
Allow uploads to non-acl mounts like Samba or EFS

orangewolf:
some file systems (nfs/samba) do not support chmod. This allows it to be disabled so carrierwave will still work on these shared drives.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/middleware/account_elevator.rb
+++ b/app/middleware/account_elevator.rb
@@ -17,7 +17,7 @@ class AccountElevator < Apartment::Elevators::Generic
     elsif Account.any?
       raise "No tenant found for #{cname}"
     else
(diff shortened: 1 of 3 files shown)


app/middleware/account_elevator.rb
```diff
@@ -17,7 +17,7 @@ def self.switch!(cname)
     elsif Account.any?
       raise "No tenant found for #{cname}"
     else
-      logger.info "It looks like we're in single tenant mode. No tenant found for #{cname}"
+      Rails.logger.info "It looks like we're in single tenant mode. No tenant found for #{cname}"
```
orangewolf, line 20: this was unrelated, but is a very long standing bug - logger is never defined in this file.

ShanaLMoore (approved):

PR #1726, opened 2021-07-19
all users where showing on admin user list instead of just the users of a given tenant

orangewolf:
We missed a page found here: https://share.getcloudapp.com/kpuDxy4K when doing the user scope limiting. This limits the scope of all "registered" calls to make sure there are not other missing

This work is sponsored by the British Library

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,7 +23,7 @@ class User < ApplicationRecord
     joins(:roles)
   }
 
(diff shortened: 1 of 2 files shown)


orangewolf:
This is green - no clue why it shows as failed in circle (it says specifically that the tests all pass)

PR #1733, opened 2021-07-29
Add missing Mailboxer gem migrations

### LINKED ISSUE #1732: Fix mailboxer gem migrations
### Descriptive summary
After Hyku upgraded the Mailboxer gem that Hyku uses as notification system, we didn't include a migration file that the gem needed for some of its latest features. This results in a 500 error being triggered when trying to send emails along the dashboard notifications.

This issue affects all versions of Hyku.

### Rationale
We need to fix this to allow sending notification emails on the approval workflow without the users getting a 500 error.

### Expected behavior
The notification emails are sent without the user getting a 500 error.

### Actual behavior
The notification emails trigger a 500 error.

1. Enable email delivery for Mailboxer (https://github.com/mailboxer/mailboxer#emails
2. Add a job into a mediated workflow
3. Make any action to trigger a dashboard notification
4. Enjoy the 500 error

okeen:
Fixes #1732 

Adds a missing migration of the Mailboxer gem that was making the email notification delivery throw a 500 error.

The migration file has been generated using the instructions at: https://github.com/mailboxer/mailboxer/issues/391

``` ruby
rails g mailboxer:install
```

* Run rails g mailboxer:install to add the missing migration file

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/concerns/account_switch.rb
+++ b/app/models/concerns/account_switch.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module AccountSwitch
   extend ActiveSupport::Concern
 
@@ -6,10 +8,10 @@ module AccountSwitch
       account = if cname_or_name_or_account.is_a?(Account)
                   cname_or_name_or_account
                   # is it a domain name?
-                elsif cname_or_name_or_account =~/^((?!-)[A-Za-z0-9-]{1, 63}(?<!-)\\.)+[A-Za-z]{2, 6}$”/
+                elsif cname_or_name_or_account =~ /^((?!-)[A-Za-z0-9-]{1, 63}(?<!-)\\.)+[A-Za-z]{2, 6}$”/
                   Account.joins(:domain_names).find_by(domain_names: {
-                    is_active: true, cname: Account.canonical_cname(cname)
-                  })
+                                                         is_active: true, cname: Account.canonical_cname(cname)
+                                                       })
                 else
                   Account.find_by(name: cname_or_name_account)
                 end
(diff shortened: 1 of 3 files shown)


orangewolf (approved):
This looks great and thanks for rubocoping the account_switch for us. Since all we are doing is adding a missing migration from the gem, I don't see any real reason it needs special spec coverage.

orangewolf:
@okeen I'm going to merge this, but wanted to bring something to your attention (since you seem to be working on email). We discovered on a different project that you have to specify a mailboxer_email method on the user or this will silently not send notification emails. See https://github.com/mailboxer/mailboxer#user-identities for a prototype.  I think Hyku can probably get away with just setting something like

```ruby
#Returning the email address of the model if an email should be sent for this object (Message or Notification).
#If no mail has to be sent, return nil.
def mailboxer_email(object)
  return "define_email@on_your.model"
end
```

I discovered this last night as the secret to make my notifications send. are you doing something like that in your user model?

okeen:
Thanks for your advise, @orangewolf! Yes, I discovered that after following @cjcolvar's always wise advise a couple of weeks ago. That's how I realised it was triggering 500 erros on mediated workflows (here is the [link to the PR](https://github.com/ubiquitypress/hyku_addons/commit/40739a1be746ca1cbdfcd993ecf11b2e1dafc72b) ).

It's something that is actually well documented at https://samvera.github.io/email_notifications.html. Considering that we both "failed" to find in on the first time, I wonder if we promote the documentation enough.

Thanks for your help, Rob!

orangewolf:
this is actually passing CI - I don't want to hold it up until coveralls decides to work again.

PR #1734, opened 2021-08-03
Draft: Theming from pals rubocop

erik-engstrom:

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

@samvera/hyku-code-reviewers


erik-engstrom:
This PR was created on the wrong fork of this Hyku project.

Issue #1699, opened 2021-04-29
Bring Helm charts and deployment strategy to Hyku


DraxIndustries79:
### Descriptive summary
This will bring helm charts from Hyrax to Hyku. The goal is to make deploying a new Hyku app easier for new adopters.

### Rationale

crisr15:
Included in V4.0 release.

PR #1737, opened 2021-08-11
Helm upgrade

orangewolf:
refs #1699

Logic from the existing Hyku chart is now covered and covered better by the Hyrax chart. Remove this code since it is now available upstream in an improved form.

@samvera/hyku-code-reviewers


.gitlab-ci.yml
```diff
@@ -0,0 +1,296 @@
...
+  when: manual
+  variables:
+    DEPLOY_IMAGE: $CI_REGISTRY_IMAGE
+    DEPLOY_TAG: $CI_COMMIT_SHORT_SHA
+    WORKER_IMAGE: $CI_REGISTRY_IMAGE/worker
+    HELM_EXPERIMENTAL_OCI: 1
+    HELM_EXTRA_ARGS: >
+      --values ops/production-deploy.yaml
+  script:
+    - export KUBECONFIG=$KUBECONFIG_ADVENTIST
```
no-reply, line 273: do we want this variable here?

ops/review-deploy.tmpl.yaml
```diff
@@ -0,0 +1,220 @@
...
+  enabled: true
+  hosts:
+    - host: hyku-$CI_MERGE_REQUEST_ID.example.com
+      paths:
+        - path: /
+    - host: "*.hyku-$CI_MERGE_REQUEST_ID.example.com"
+      paths:
+        - path: /
+
+extraEnvVars:
```
no-reply, line 42: as follow up to this, let's look into getting some of this into the Hyrax chart as a courtesy to Hyku users. especially the `SETTINGS__` stuff is harmless to non hyku users but in some cases duplicates other variables used by the engine.

ops/review-deploy.tmpl.yaml
```diff
@@ -0,0 +1,220 @@
...
+    mountPath: /app/samvera/hyrax-webapp/tmp/network_files
+    subPath: network-files
+
+ingress:
+  enabled: true
+  hosts:
+    - host: hyku-$CI_MERGE_REQUEST_ID.example.com
+      paths:
+        - path: /
+    - host: "*.hyku-$CI_MERGE_REQUEST_ID.example.com"
```
no-reply, line 38: does this work? maybe we just want `"*.hyku.example.com"` for an example?

i don't think this values file will succeed at resolving these environment variables.

ops/review-deploy.tmpl.yaml
```diff
@@ -0,0 +1,220 @@
...
+  - name: SETTINGS__FITS_PATH
+    value: /app/fits/fits.sh
+  - name: SETTINGS__MULTITENANCY__ADMIN_ONLY_TENANT_CREATION
+    value: "true"
+  - name: SETTINGS__MULTITENANCY__DEFAULT_HOST
+    value: "%{tenant}.hyku-staging.example.com"
+  - name: SETTINGS__MULTITENANCY__ENABLED
+    value: "true"
+  - name: SETTINGS__SOLR__COLLECTION_OPTIONS__COLLECTION__CONFIG_NAME
+    value: $CI_COMMIT_SHORT_SHA
```
no-reply, line 84: same question about environment variables here and down

ops/review-deploy.tmpl.yaml
```diff
@@ -0,0 +1,220 @@
...
+  runAsGroup: 101
+  fsGroup: 101
+  fsGroupChangePolicy: "OnRootMismatch"
+
+embargoRelease:
+  enabled: false
+leaseRelease:
+  enabled: false
+
+imagePullSecrets:
```
no-reply, line 208: maybe comment this in general. i think the base case is that the registry is likely to be public, and a pull secret would need to be manually created per namespace, so having a template default feels like it does more to confuse than anything else.

ops/review-deploy.tmpl.yaml
orangewolf: ` - envsubst < ops/dev-deploy.tmpl.yaml > ops/dev-deploy.yaml` means yes it will resolve variables. That's why I renamed them tmpl (this matches the nginx and several other significant dockerize projects do things)

I'd be open to moving that to the helm_deploy if you think that is useful

ops/review-deploy.tmpl.yaml
orangewolf: I agree

ops/review-deploy.tmpl.yaml
orangewolf: I think we are going to go the opposite way - one of the very next steps is to 100% redo the way settings work in Hyku and I think syncing them to use the Hyrax ones where we can and simplifying to parts where we can not can happen at that time.

no-reply: oh, that's great.

ops/review-deploy.tmpl.yaml
no-reply: ah. this makes sense. ty!

PR #1739, opened 2021-08-18
Update model specs

labradford:
Rubocop recommends not using the Let! block, however tests copied over from Hyrax are using them. Reverted test back to use the Let! block. Added the model specs that use it to the rubocop.yml file


labradford:
Opened in error

Issue #1719, opened 2021-07-01
Per-tenant settings


cjcolvar:
### Descriptive summary

Accounts in Hyku should have a place for a variety of configurable settings beyond fedora, solr, redis, and database settings.   This can be done using ActiveRecord's database-independent `store` / `store_accessor` which would allow for arbitrary key/value pairs to be added by customized Hyku applications.  There should be a UI for tenant admins to configure these settings as well as a UI to configure these from the account edit or manage pages in the superadmin dashboard.

Follow on work would be removing the dependency on `rubyconfig` and providing an implementation of `Settings` that checks the account's settings, then ENV variables, then defaults.  The Hyrax config would need to be updated on account switch like is being done in #1713.

### Related work

https://github.com/samvera/hyku/issues/1685
https://github.com/samvera/hyku/pull/1713

Evolved implementation in `hyku_addons`:
Account model behavior mixin: https://github.com/ubiquitypress/hyku_addons/blob/main/app/models/concerns/hyku_addons/account_behavior.rb
Superadmin dashboard edit view: https://github.com/ubiquitypress/hyku_addons/blob/main/app/views/proprietor/accounts/edit.html.erb#L54-L70
Tenant dashboard views: https://github.com/ubiquitypress/hyku_addons/tree/main/app/views/hyku_addons/account_settings
Account settings controller: https://github.com/ubiquitypress/hyku_addons/blob/main/app/controllers/hyku_addons/account_settings_controller.rb

PR #1746, opened 2021-09-05
Per Tenant Settings: Settings Overhaul Part 1

orangewolf:
Ref #1719 

Most of the credit for this work goes to Ubiquity Press. Many of the lines here are extracted from https://github.com/ubiquitypress/hyku_addons

This development was funded by the British Library: https://bl.iro.bl.uk

Ports per tenant settings from HykuAddons and expands on them to make settings definable as properties. This is designed to be json compatible with Ubiquities current data structure. No migration should be needed for those using HykuAddons.  This also lays the groundwork for DOI as a first class feature in Hyku, which is also WIP.  Because DOI is off by default, we feel good that this can be merged and then turned on when we are ready to. 

Next Steps:

- Port or discard implementations of existing settings
- Port the admin interface for settings
- Align all existing settings and remove the Settings class
- Set defaults based on Hyrax env var names

@samvera/hyku-code-reviewers


cjcolvar:
FWIW when I started integrating hyrax-doi into hyku I thought another endpoint made sense which is what I implemented but later I was wondering if simple settings would make more sense since they are lighter weight.

orangewolf:
> FWIW when I started integrating hyrax-doi into hyku I thought another endpoint made sense which is what I implemented but later I was wondering if simple settings would make more sense since they are lighter weight.

I got about half way through porting it and thought the same thing. Honestly I think it is a toss up, so I was going with whatever matched Ubiquity to ease migration.

ShanaLMoore (approved):

PR #1747, opened 2021-09-09
draft of uncompleted implementation of cross tenant search

mankind:
Hi I know you are working on moving settings to Hyku, but to save time I just used the approach to settings in hyku_addons.

Also is **shared_search_enabled**  will be moved to Flipflop as a feature flag later on.

### DIFF
--- a/app/jobs/create_solr_collection_job.rb
+++ b/app/jobs/create_solr_collection_job.rb
@@ -9,18 +9,21 @@ class CreateSolrCollectionJob < ApplicationJob
   def perform(account)
     name = account.tenant.parameterize
 
-    unless collection_exists? name
+    perform_for_cross_search_tenant(account, name) if account.is_shared_search_enabled? && account.tenant_list.present?
+    perform_for_normal_tenant(account, name) unless account.is_shared_search_enabled? && account.tenant_list.present?
+
+    account.add_parent_id_to_child
+  end
+
+  def without_account(name, tenant_list = '')
+    return if collection_exists?(name)
+    if tenant_list.present?
+      client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATEALIAS',
+                                                                             name: name, collections: tenant_list)
+    else
       client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATE',
                                                                              name: name)
     end
-
-    account.create_solr_endpoint(url: collection_url(name), collection: name)
-  end
-
-  def without_account(name)
-    return if collection_exists?(name)
-    client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATE',
-                                                                           name: name)
   end
 
   # Transform settings from nested, snaked-cased options to flattened, camel-cased options
@@ -87,5 +90,40 @@ class CreateSolrCollectionJob < ApplicationJob
       uri = URI(normalized_uri) + name
 
       uri.to_s
+    end
+
+    def add_solr_endpoint_to_account(account, name)
+      account.create_solr_endpoint(url: collection_url(name), collection: name)
+    end
+
+    def perform_for_normal_tenant(account, name)
+      unless collection_exists? name
+        client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATE',
+                                                                               name: name)
+      end
+       add_solr_endpoint_to_account(account, name)
+    end
+
+    def perform_for_cross_search_tenant(account, name)
+      account_children = account.children.map(&:tenant)
+      tenants_from_edit = account.tenant_list - account_children
+      tenant_list = account.tenant_list.join(',')
+
+      if tenants_from_edit.present?
+        solr_options = account.solr_endpoint.connection_options.dup
+        RemoveSolrCollectionJob.perform_now(name, solr_options, 'cross_search_tenant')
+        create_shared_search_collection(tenant_list, name)
+        account.solr_endpoint.update(url: collection_url(name), collection: name)
+      else
+        create_shared_search_collection(tenant_list, name)
+        account.create_solr_endpoint(url: collection_url(name), collection: name)
+      end
+    end
+
+    def create_shared_search_collection(tenant_list, name)
+      unless collection_exists? name
+        client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATEALIAS',
+                                                                               name: name, collections: tenant_list)
+      end
     end
 end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -2,6 +2,12 @@
 
 # Customer organization account
 class Account < ApplicationRecord
+  scope :not_cross_search_tenants, -> {where('settings @> ?', {shared_search: 'false'}.to_json)}
+  has_many :children, class_name: "Account", foreign_key: "parent_id", dependent: :nullify, inverse_of: :parent
+  belongs_to :parent, class_name: "Account", inverse_of: :parent, foreign_key: "parent_id", optional: true
+  store_accessor :settings, :shared_search, :tenant_list
+  after_initialize :initialize_settings
+
   # @param [String] piece the tenant piece of the canonical name
   # @return [String] full canonical name
   # @raise [ArgumentError] if piece contains a trailing dot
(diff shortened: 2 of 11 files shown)


app/jobs/create_solr_collection_job.rb
```diff
@@ -9,18 +9,21 @@ class CreateSolrCollectionJob < ApplicationJob
   def perform(account)
     name = account.tenant.parameterize
 
-    unless collection_exists? name
-      client.get '/solr/admin/collections', params: collection_options.merge(action: 'CREATE',
-                                                                             name: name)
-    end
+    perform_for_cross_search_tenant(account, name) if account.is_shared_search_enabled? && account.tenant_list.present?
+    perform_for_normal_tenant(account, name) unless account.is_shared_search_enabled? && account.tenant_list.present?
```
orangewolf, line 13: I think an if/else might be better here. I've read this 3 times and I'm pretty sure that if you number these methods like so

1 = perform_for_cross_search_tenant
2 = perform_for_normal_tenant
3 = neither one is done

and make a truth table you get:

tenant_list.present? && is_shared_search_enabled?  => 1
just is_shared_search_enabled?  => 3
just tenant_list.present?  => 3
neither tenant_list.present? or is_shared_search_enabled? => 2

It doesn't seem right to me that would be the case.

```suggestion
    if account.is_shared_search_enabled? && account.tenant_list.present?
      perform_for_cross_search_tenant(account, name)
    else
      perform_for_normal_tenant(account, name)
    end
```

app/models/account.rb
```diff
@@ -2,6 +2,12 @@
 
 # Customer organization account
 class Account < ApplicationRecord
+  scope :not_cross_search_tenants, -> {where('settings @> ?', {shared_search: 'false'}.to_json)}
+  has_many :children, class_name: "Account", foreign_key: "parent_id", dependent: :nullify, inverse_of: :parent
+  belongs_to :parent, class_name: "Account", inverse_of: :parent, foreign_key: "parent_id", optional: true
+  store_accessor :settings, :shared_search, :tenant_list
```
orangewolf, line 8: I don't think you need to store this one boolean field in JSON. maybe just a good old fashioned db field? I see why you might want tenant_list here.

orangewolf: I didn't realize that this feature would depend on account settings. it will need a little work to be mergeable but probably not too much

app/views/proprietor/accounts/edit.html.erb
```diff
@@ -34,6 +34,32 @@
             <%= s.input :base_path %>
           <% end %>
 
+          <p>
+						<%= f.check_box :shared_search, {checked: @account.is_shared_search_enabled?, name: "account[settings][shared_search]", id: "account_settings_shared_search"}, 'true' , 'false' %>
```
orangewolf, line 38: this needs some spacing adjustments please.

app/views/proprietor/accounts/edit.html.erb
```diff
@@ -34,6 +34,32 @@
...
+						<%= f.label "Search Only(select tenant on next page)" %>
```
orangewolf, line 39: please use i18n tags

db/migrate/20210802093531_add_parent_id_and_settings_to_accounts.rb
```diff
@@ -0,0 +1,7 @@
+class AddParentIdAndSettingsToAccounts < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :accounts, :parent, foreign_key: { to_table: :accounts }
+    add_column :accounts, :settings, :jsonb, default: {}
```
orangewolf, line 4: this migration will need to be adjusted as the settings stuff is already in

spec/rails_helper.rb
```diff
@@ -149,6 +149,8 @@
     DatabaseCleaner.clean_with(:truncation)
     Account.destroy_all
     CreateSolrCollectionJob.new.without_account('hydra-test') if ENV['IN_DOCKER']
+    CreateSolrCollectionJob.new.without_account('hydra-sample') if ENV['IN_DOCKER']
+    CreateSolrCollectionJob.new.without_account('hydra-cross-search-tenant', 'hydra-test, hydra-sample') if ENV['IN_DOCKER']
```
orangewolf, line 153: 👍

app/models/account.rb
```diff
@@ -2,6 +2,12 @@
 
 # Customer organization account
 class Account < ApplicationRecord
+  scope :not_cross_search_tenants, -> {where('settings @> ?', {shared_search: 'false'}.to_json)}
+  has_many :children, class_name: "Account", foreign_key: "parent_id", dependent: :nullify, inverse_of: :parent
+  belongs_to :parent, class_name: "Account", inverse_of: :parent, foreign_key: "parent_id", optional: true
```
orangewolf, line 7: what does this parent / child thing do in your repo? I see you adding the infra here but it doesn't seem like the whole feature is happening? I'm maybe ok with that, but I would like to understand what the end goal is for it if possible.

orangewolf (changes requested):
this feels super solid, but needs some minor revision and to meet up with the settings code that went in to Hyku. I'm going to try and give you back a draft that has some of that.

orangewolf:
I think I may finally have my head around the parent/child thing - its is _part_ of this cross tenant thing.  I thought it was a separate feature getting mixed in here. I'm concerned about the naming and more over, I don't see a reason why a given normal tenant (maybe we call them Full Tenant and Search Tenant ?)  could not be in more that one search tenant. I think that the fundemental idea that these are somehow children is maybe incorrect.

app/views/proprietor/accounts/edit.html.erb
mankind: Thanks I will use i18n tags in the update

db/migrate/20210802093531_add_parent_id_and_settings_to_accounts.rb
mankind: Ok noted thanks.

app/views/proprietor/accounts/edit.html.erb
mankind: I have actually moved this to a partial that i can shared between new and edit files

app/jobs/create_solr_collection_job.rb
mankind: Thanks that change has been applied. Thanks

app/models/account.rb
mankind: A search tenant is  a parent of all the tenants in its solr collection alias. So it allows us to do somethings:
1. Query for  the tenants in a particular shared search and check those in the edit page
2.  For Ubiquity we  have  a tab next to contacts which is called 'shared search; which points to the parent or the shared search a particular tenant is in.  when you click it, it redirect to the shared search page..

mankind:
> I think I may finally have my head around the parent/child thing - its is _part_ of this cross tenant thing. I thought it was a separate feature getting mixed in here. I'm concerned about the naming and more over, I don't see a reason why a given normal tenant (maybe we call them Full Tenant and Search Tenant ?) could not be in more that one search tenant. I think that the fundemental idea that these are somehow children is maybe incorrect.

Yes it is part of the cross search tenant

mankind:
> 
> 
> I think I may finally have my head around the parent/child thing - its is _part_ of this cross tenant thing. I thought it was a separate feature getting mixed in here. I'm concerned about the naming and more over, I don't see a reason why a given normal tenant (maybe we call them Full Tenant and Search Tenant ?) could not be in more that one search tenant. I think that the fundemental idea that these are somehow children is maybe incorrect.

@orangewolf  We can rename them full_tenant and search_tenant if you want. 

A Full Tenant can be in more than one search. Lets say you have full tenants A, B, C, D, E 
You can have **Search Tenant 1** which will contain **Full tenants A, B, C**
We can still have **Search Tenant 2** which will contain **Full Tenant D** and **Search Tenant 1**.
Note that since **Search Tenant 1** already includes Full Tenants A, B, C and that means **Search Tenant 2** actually has data for **Full Tenants A, B, C, D**.

When you add , remove or update the record in any of those tenants eg **Full tenant A**, then **Search Tenant 1** and **Search Tenant 2**  gets updated as well.

mankind:
> 
> 
> I think I may finally have my head around the parent/child thing - its is _part_ of this cross tenant thing. I thought it was a separate feature getting mixed in here. I'm concerned about the naming and more over, I don't see a reason why a given normal tenant (maybe we call them Full Tenant and Search Tenant ?) could not be in more that one search tenant. I think that the fundemental idea that these are somehow children is maybe incorrect.

I am just thinking out loud here. Since we want a Full-tenant to be able to be in many Shared-Tenant but shared-tenant can not be in a full-tenant,  that is not really a many to many neither is it a one to many association. Therefore do you like the idea of no real database association to show that relationship instead we just a store an array of tenant uuid that are in a shared search tenant in the tenant_list array. This will only stored in a Search Tenant. This approach loses referential integrity but do we need that in this situation, since we only need the tenant_list of full_tenants in a shared search edit page and for listing the shared search a full tenant is in, elsewhere in the shared search tab or page.

What do you think @orangewolf

orangewolf:
this work was merged as part of #1753

PR #1751, opened 2021-09-20
Settings overhaul 3: UI Time

orangewolf:
Ref #1719

Most of the credit for this work goes to Ubiquity Press. Many of the lines here are extracted from https://github.com/ubiquitypress/hyku_addons

This development was funded by the British Library: https://bl.iro.bl.uk

Part 3 of settings overhaul #1746 This PR adds an admin interface to the settings and removes the contact controller which is no longer needed. This code resurrects a controller that already existed but was broken and not connected to the rest of the UI and fills it in with admin settings.

At the moment, string and boolean settings are supported, but array and hash are not yet supported.


orangewolf:
this work was merged as part of #1753

PR #1753, opened 2021-09-20
Settings plus cross search

orangewolf:
This is a combination of #1747 and #1750. It fixes any open bugs in both branches.  Shared search was done by Ubiquity Press with special thanks to @mankind for the majority of this code. The British Libraries sponsored work for Notch8 to integrate the cross site search and the new settings UI in to Hyku's main repository.


mankind:
Thanks @orangewolf for some reasons, I didn't see this notification and only just saw it thought I received it on 20th September.  

 Thanks for pulling this into the settings work. The change to many to many association and adding a search_only boolean field is  fine. 
 
 I will go through the code tomorrow.

PR #1777, opened 2022-01-15
Updates from WUSTL Newman Numismatic Portal

orangewolf:
- create jpgs from pdf
- only escape urls that are unsafe
- newspaper_works gem
- attach_files_to_work performance improvements
- propritor views should reset account properly
- helm template helper command
- load decorator files
- long fcrepo timeout for prod
- fix active_fedora stack overflow on manage files page
- add admin role during seed
- fix hydra deriviatives tripple pdf in previews


orangewolf:
going to need to take a new swing at this after iiif_print is done

PR #1792, opened 2022-04-05
Featured collections

summer-cook:
# Summary
Allows admins to feature a collection in the same way as we can feature works
Allows admin to choose the display order of the featured collections
Updates all 4 themes in the hyku repo

# Acceptance Criteria
- [ ] Admin users can see a feature/unfeature button to the collections page to match the functionality on works
- [ ] Admin users have the ability to sort featured collections on the home page to match the functionality of works
- [ ] Admin users can add any number of featured collections between 0 and 6.  

# Note
this also adds a style to bring the color of the share your work button text & show_actions button texts back to white, since it was improperly scoped after previous work that added body.public-facing to appearance_styles.rb

## Videos

**Default theme:** https://share.getcloudapp.com/X6uEbjl7

**Neutral Theme:** https://share.getcloudapp.com/yAuK0GKY

**IR Theme:** https://share.getcloudapp.com/mXuYPQ1E

**Cultural Repo Theme:** https://share.getcloudapp.com/yAuK0erx

## Screenshots
share work button before:
Screen Shot 2022-04-06 at 2 45 47 PM

share work button a
Screen Shot 2022-04-06 at 2 42 58 PM
fter:

@samvera/hyku-code-reviewers


app/actors/hyrax/actors/featured_collection_actor.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Actors
+    # Removes featured collections if the collection is deleted or becomes private
+    class FeaturedCollectionActor < Hyrax::Actors::AbstractActor
```
orangewolf, line 6: do collections go through the same actor stack as works do or is it a separate flow?

summer-cook: I'm not _entirely_ sure if this answers what you are asking, but the structure is that there is both a `FeaturedCollectionActor` and a `FeaturedWorkActor`. Both of these actors are called in the `DefaultMiddlewareStack`

summer-cook: So I think there is a separate flow.

orangewolf: but you are adding it to the middleware stack for works.  That seems like a bug. lets pair up and investigate this before we merge

orangewolf (changes requested):
hold until we talk through actor stack issues

app/services/hyrax/default_middleware_stack.rb
```diff
@@ -0,0 +1,65 @@
+# OVERRIDE Hyrax 2.9.1 to add Featured Collection Actor to build_stack method
```
orangewolf, line 1: @summer-cook this whole override can be removed now?

orangewolf (dismissed):

orangewolf (approved):

PR #1801, opened 2022-05-11
fix single tenant catalog links

orangewolf:
Louisville reported issues with links on the catalog pointing to single.tenant.default as the URL. This is because all page views were using the account_cname instead of just on shared search views.

### DIFF
--- a/app/helpers/shared_search_helper.rb
+++ b/app/helpers/shared_search_helper.rb
@@ -2,7 +2,7 @@
 
 module SharedSearchHelper
   def generate_work_url(model, request)
-    return model if current_account.search_only?
+    return model unless current_account.search_only?
     # needed because some attributes eg id is a symbol 7 others are string
     model = model.to_h.with_indifferent_access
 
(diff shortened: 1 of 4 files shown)


app/helpers/shared_search_helper.rb
```diff
@@ -2,8 +2,9 @@
 
 module SharedSearchHelper
   def generate_work_url(model, request)
+    return model if current_account.search_only?
```
ShanaLMoore, line 5:
```suggestion
    return model unless current_account.search_only?
```

ShanaLMoore: Confirmed this work with LV with the above change

ShanaLMoore:
Specs need to be updated. I will submit a fix soon.

spec/helpers/shared_search_helper_spec.rb
```diff
@@ -5,12 +5,16 @@
 
   context "shared search records" do
     let(:cname) { 'hyku-me.test' }
-    let(:account) { build(:account, cname: cname) }
+    let(:account) { build(:search_only_account, cname: cname) }
```
ShanaLMoore, line 8: ideally would like to include a context for when account is not shared_search but I don't think I'll get back to this in time. I need to take care of some other issues before PTO! cc @orangewolf

jpengst (approved):

alishaevn (approved):

PR #1838, opened 2022-08-17
1837 better user error handling

alishaevn:
# expected behavior
- stop checking for `@user.errors.details[key].first[:value]` because it does not exist

# related
- https://gitlab.com/notch8/britishlibrary/-/issues/372

@samvera/hyku-code-reviewers

### DIFF
--- a/app/views/proprietor/users/new.html.erb
+++ b/app/views/proprietor/users/new.html.erb
@@ -12,9 +12,8 @@
                 <%= pluralize(@user.errors.count, "error") %> prohibited this repository from being saved:
                 <ul>
                   <% @user.errors.messages.each do |key, messages| %>
-                  <% if @user.errors.details[key].present? %>
-                    <li><%= key %> &quot;<%= @user.errors.details[key].first[:value] %>&quot; <%= messages.join(' and ') %></li>
-                  <% end %>
+                    <% next unless messages.present? %>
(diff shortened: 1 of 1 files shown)


app/views/proprietor/users/new.html.erb
```diff
@@ -12,9 +12,8 @@
                 <%= pluralize(@user.errors.count, "error") %> prohibited this repository from being saved:
                 <ul>
                   <% @user.errors.messages.each do |key, messages| %>
-                  <% if @user.errors.details[key].present? %>
-                    <li><%= key %> &quot;<%= @user.errors.details[key].first[:value] %>&quot; <%= messages.join(' and ') %></li>
-                  <% end %>
+                    <% next unless messages.present? %>
+                    <li><%= key %> <%= messages.join(' and ') %></li>
```
alishaevn, line 16: removed `@user.errors.details[key].first[:value]` because it doesn't exist per the screenshots below. unsure if it is supposed to or not.

if `[:value]` is supposed to exist, we should instead fix why it isn't showing up in the `details` array.

alishaevn: ### user "new" view page
Screen Shot 2022-08-17 at 10 36 20 AM

### user controller
Screen Shot 2022-08-17 at 10 47 32 AM

orangewolf (approved):

PR #1839, opened 2022-08-22
rubocop updates, autocorrects and manual fixes

orangewolf:
Update Rubocop to a more recent release.


orangewolf:
too old, needs to be redone

PR #1840, opened 2022-09-01
add a reset task that wraps the task from bulkrax

orangewolf:
Refs https://github.com/samvera-labs/bulkrax/issues/647

This task wraps Bulkrax's reset task and makes it work for all tenants. Waiting for https://github.com/samvera-labs/bulkrax/pull/648 to be merged and Bulkrax to be updated before we merge this.

@samvera/hyku-code-reviewers

### DIFF
--- a/lib/tasks/reset.rake
+++ b/lib/tasks/reset.rake
@@ -1,0 +1,20 @@
+namespace :hyku do
+  desc 'reset work and collection data across all tenants'
+  task works_and_collections: [:environment] do
+    confirm('You are about to delete all works and collections across all accounts, this is not reversable!')
+    Account.find_each do |account|
+      switch!(account)
+      Rake::Task["hyrax:reset:works_and_collections"].reenable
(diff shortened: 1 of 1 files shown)


lib/tasks/reset.rake
```diff
@@ -0,0 +1,20 @@
+namespace :hyku do
+  desc 'reset work and collection data across all tenants'
+  task works_and_collections: [:environment] do
+    confirm('You are about to delete all works and collections across all accounts, this is not reversable!')
+    Account.find_each do |account|
```
ShanaLMoore, line 5: Do you think we should make a tenant argument optional instead of resetting all of them?

ShanaLMoore (approved):

lib/tasks/reset.rake
ShanaLMoore: task name implies the intention is to do this across all tenants. So I'll take back my above comment and approving this PR

orangewolf: I did change it today shana =-)

PR #1841, opened 2022-09-05
contribute_back_i541-featured-works-save

aprilrieger:
This is contributed back work from the Palni Palci Hyku Upgrade. This update was a bug that didn't allow the admin user the ability to save the featured works list. Ticket: https://gitlab.com/notch8/palni-palci/-/issues/541


labradford (approved):
looks good to me

PR #1844, opened 2022-09-23
Fix shared search URLs

laritakr:
Fixes remaining problems with UTK issue 80

[Blacklight's search_context.js](https://github.com/projectblacklight/blacklight/blob/b37a4642059c4af125372e7faa969b5245b6ecd4/app/assets/javascripts/blacklight/search_context.js#L1-L42) introduces click-jacking behavior for A-tags that conflict with multi-tenant behavior. This click-jacking behavior uses a relative path in its form submission.

The manifestation of this problem is when you click on an item in the catalog index and get a Faraday error that has the port 99999. That 99999 means that there’s no loaded tenant the various endpoints (see [AccountEndpoints module](https://gitlab.com/notch8/utk-hyku/blob/8d0018f06f03e86ecc131209bc43cad91310add7/app/models/concerns/account_endpoints.rb#L21-35)) for Solr and Fedora use the nil handler (see )and our nil endpoint handler is loaded. See the [NilSolrEndpoint](https://gitlab.com/notch8/utk-hyku/blob/8d0018f06f03e86ecc131209bc43cad91310add7/app/models/nil_solr_endpoint.rb#L1-28). This is evident in the work of UTK issue 80, where we coerced an ID (which normally renders as a path) to a URL to the host tenant.

To remedy the link jacking, we have two options:

1. Amend [Blacklight::UrlHelperBehavior#document_link_params](https://github.com/projectblacklight/blacklight/blob/b37a4642059c4af125372e7faa969b5245b6ecd4/app/helpers/blacklight/url_helper_behavior.rb#L32-L35) to ignore the the session_tracking_params method call when dealing with a full URL.
2. Amend [Blacklight::UrlHelperBehavior#session_tracking_path](https://github.com/projectblacklight/blacklight/blob/b37a4642059c4af125372e7faa969b5245b6ecd4/app/helpers/blacklight/url_helper_behavior.rb#L75-L87) to return a URL instead of a relative path.

* Use the first of the two options listed above to allow full links to resolve
* Remove CatalogHelperBehavior overrides which are now obsolete
* Remove partials previously overridden from Hyrax which are no longer needed due to this change

@samvera/hyku-code-reviewers


app/views/catalog/_index_gallery_collection_wrapper.html.erb
```diff
@@ -0,0 +1,11 @@
+<%# OVERRIDE Hyrax 3.4.1: make collection thumbnail a link so it matches works %>
+<div class="document col-xs-6 col-md-3">
+  <div class="thumbnail">
+  <% debugger %>
```
ShanaLMoore, line 4: Oops! spotted a debugger.

ShanaLMoore (changes requested):
I know you are leaving for the day so I plan to correct this

app/helpers/blacklight/url_helper_behavior_decorator.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# OVERRIDE blacklight 6 to allow full url for show links for shared search and override blacklight url tracking
+module Blacklight
```
ShanaLMoore, line 4: Until we figure out UTK's decorator issue, we'll prob have to convert this to module_eval once pulled into UTK. @laritakr

app/views/catalog/_index_gallery_collection_wrapper.html.erb
laritakr: I don't know how that happened because that isn't in my code and it doesn't show any changes. :/

app/helpers/blacklight/url_helper_behavior_decorator.rb
laritakr: Most likely. I will have to test it to make sure when I pull it in.

ShanaLMoore (changes requested):
Taking this for a spin, after I created a work w a fileset for each tenant. I then made a collection. I was going back to the work's show page to add the work to the collection but ran into this error: 

Screen Shot 2022-09-23 at 3 02 54 PM

Not sure if it's related to the changes in this PR but I'd like to take some time to look into it and maybe correct it here if so.

ShanaLMoore:
> Taking this for a spin, after I created a work w a fileset for each tenant. I then made a collection. I was going back to the work's show page to add the work to the collection but ran into this error:
> 
> Screen Shot 2022-09-23 at 3 02 54 PM
> 
> Not sure if it's related to the changes in this PR but I'd like to take some time to look into it and maybe correct it here if so.

The issue is that from the catalog controller, doc is a solr document. but from the work show page it's a fileset presenter and calling to_h doesn't work in that context.

ShanaLMoore:
@laritakr I pushed up what I think may be the fix to the issue described above. Please review at your convenience. 

- [ ] I tested that the shared search links and links via thumbnails are all going to the correct place. I approve this PR if you agree with my [changes](https://github.com/samvera/hyku/pull/1844/commits/d49d6535b4e588cf652e0e4981631a6e78c6de3a). Otherwise ping me again when this is ready, if you decide to do something different to resolve it. 

Demo: https://share.getcloudapp.com/X6uQAJBN
Screenshots: 

Screen Shot 2022-09-23 at 4 03 59 PM
Screen Shot 2022-09-23 at 4 03 47 PM

ShanaLMoore:
> @laritakr I pushed up what I think may be the fix to the issue described above. Please review at your convenience.
> 
>     * [x]  I tested that the shared search links and links via thumbnails are all going to the correct place.
> 
>     * [x]  I also tested that the thumbnails on the work's show page still work
> 
> 
> I approve this PR if you agree with my [changes](https://github.com/samvera/hyku/pull/1844/commits/d49d6535b4e588cf652e0e4981631a6e78c6de3a). Otherwise ping me again when this is ready, if you decide to do something different to resolve it.
> 
> Demo: https://share.getcloudapp.com/X6uQAJBN Screenshots:
> 
> Screen Shot 2022-09-23 at 4 03 59 PM Screen Shot 2022-09-23 at 4 03 47 PM

As an update, I pulled over the fix from UTK. I retested it and it's all still working.

laritakr:
Thanks for following up, @ShanaLMoore The changes you made are what I had in mind.

ShanaLMoore (approved):

PR #1854, opened 2022-12-06
WIP: Install iiif print

laritakr:
Ref https://github.com/scientist-softserv/iiif_print/issues/6
This branch documents the results of running the iiif_print install generator. 

Updated & added files were compared with similar files in Louisville and the [differences I found are noted here.](https://docs.google.com/document/d/1esw3HbnbCDRNgRgwn-QarImtK52JBpjghpq4vsezOhg/edit)


app/controllers/catalog_controller.rb
```diff
@@ -16,7 +17,30 @@ def self.modified_field
...
+
+
+    # IiifPrint index fields
+    config.add_index_field 'all_text_tsimv', highlight: true, helper_method: :render_ocr_snippets
+
+
+    # configuration for Blacklight IIIF Content Search
+    config.iiif_search = {
+      full_text_field: 'all_text_tsimv',
+      object_relation_field: 'is_page_of_s',
```
ShanaLMoore, line 38: @laritakr 

Can we change this to is_page_of_ssim to match LV && to support multi value: https://github.com/scientist-softserv/louisville-hyku/blob/main/app/controllers/catalog_controller.rb#L35

laritakr: This is filled in by `blacklight_iiif_search_generator.rb` during install in iiif_print. It can be changed, but ideally I'd think we probably would want a config file variable rather than leaving it hardcoded in the generator file.

laritakr: Just a note: I did not change ANYTHING in this pull request. It all came directly from the iiif_print install generator.

PR #1862, opened 2022-12-15
Fixes Rubocop errors on groups-with-roles-feature branch

DiemBTran:
Fixes Rubocop errors on groups-with-roles-feature branch


app/controllers/admin/group_roles_controller.rb
```diff
@@ -59,7 +59,8 @@ def cannot_remove_admin_role_from_admin_group
         role = Role.find_by(id: params[:role_id])
         return unless @group.name == ::Ability.admin_group_name && role.name == 'admin'
 
-        redirect_back(fallback_location: edit_admin_group_path(@group), flash: { error: "Admin role cannot be removed from this group" })
+        redirect_back(fallback_location:
+          edit_admin_group_path(@group), flash: { error: "Admin role cannot be removed from this group" })
```
bkiahstroud, line 63: This will work, but could be more readable: 

```suggestion
        redirect_back(
          fallback_location: edit_admin_group_path(@group),
          flash: { error: "Admin role cannot be removed from this group" }
        )
```

app/controllers/hyrax/admin/collection_type_participants_controller.rb
```diff
@@ -1,13 +1,18 @@
+# frozen_string_literal: true
+
 # OVERRIDE FILE from Hryax v2.9.0
-require_dependency Hyrax::Engine.root.join('app', 'controllers', 'hyrax', 'admin', 'collection_type_participants_controller').to_s
+require_dependency Hyrax::Engine.root
+                                .join('app', 'controllers', 'hyrax', 'admin', 'collection_type_participants_controller')
+                                .to_s
```
bkiahstroud, line 6: This isn't terrible by any means, but could probably be a bit more readable. Let's link up on this one.

app/models/concerns/hyrax/ability/collection_ability.rb
```diff
@@ -91,9 +116,7 @@ def collection_roles
           models.each do |collection_model|
             # Permit all actions (same collection permissions as admin users)
             can :manage, collection_model
-            can :manage, ::SolrDocument do |solr_doc|
-              solr_doc.collection?
-            end
+            can :manage, ::SolrDocument, &:collection?
```
bkiahstroud, line 119: I've never seen this syntax before, it's interesting. I think this change is good, I would just want to double check that it's accomplishing the same thing it was before (we can do this together)

app/services/roles_service.rb
```diff
@@ -77,9 +78,11 @@ def create_default_roles!
 
     def create_default_hyrax_groups_with_roles!
       # Prevent Hyrax::Groups from being created in the public schema
-      return '`AccountElevator.switch!` into an Account before creating default Hyrax::Groups' if Site.instance.is_a?(NilSite)
+      return '`AccountElevator.switch!` into an Account before creating default Hyrax::Groups' if Site.instance
+                                                                                                      .is_a?(NilSite)
```
bkiahstroud, line 82: One-line conditionals are nice, but sometimes it makes sense to break it into an `if...end` if the line is too long: 

```suggestion
      if Site.instance.is_a?(NilSite)
        return '`AccountElevator.switch!` into an Account before creating default Hyrax::Groups'
      end
```

app/services/roles_service.rb
```diff
@@ -77,9 +78,11 @@ def create_default_roles!
...
 
-      default_hyrax_groups_with_roles = DEFAULT_HYRAX_GROUPS_WITH_ATTRIBUTES.deep_merge(DEFAULT_ROLES_FOR_DEFAULT_HYRAX_GROUPS)
+      default_hyrax_groups_with_roles = DEFAULT_HYRAX_GROUPS_WITH_ATTRIBUTES
+                                        .deep_merge(DEFAULT_ROLES_FOR_DEFAULT_HYRAX_GROUPS)
```
bkiahstroud, line 85: This works, but I think there might be a more ideal breaking point. This might still be too long, but we can try it: 

```suggestion
      default_hyrax_groups_with_roles =
              DEFAULT_HYRAX_GROUPS_WITH_ATTRIBUTES.deep_merge(DEFAULT_ROLES_FOR_DEFAULT_HYRAX_GROUPS)
```

app/services/roles_service.rb
```diff
@@ -265,7 +275,7 @@ def seed_qa_users!
             end
           end
 
-          puts "Email: #{user.email}\nRoles: #{user.roles.map(&:name)}\n\n"; nil
+          Rails.logger "Email: #{user.email}\nRoles: #{user.roles.map(&:name)}\n\n"
```
bkiahstroud, line 278: I'm not positive this will do the same thing it was before, so we should look at it

bkiahstroud (commented):
Overall this is looking great so far! I left a few notes below:

app/controllers/admin/group_roles_controller.rb
bkiahstroud: (There are a few of this, so simply clicking `Commit suggestion` won't get all of them unfortunately)

DiemBTran: Agreed on this, thank you! Will make those changes

app/models/concerns/hyrax/ability/collection_ability.rb
DiemBTran: This one was an autocorrect one. Yes, let's look at this together!

app/services/roles_service.rb
DiemBTran: This is one that JP and I googled together HA yes, let's look at it together!

app/controllers/admin/group_roles_controller.rb
DiemBTran: I think I found them all and made those changes. Thank you!

labradford (approved):

PR #1886, opened 2023-01-10
add conditional to account for when iiif print gem is installed

kirkkwang:
With all the work going in the [IIIF Print](https://github.com/scientist-softserv/iiif_print) gem, we need a way to safely override methods.  Instead of updating the `IiifManifestPresenterDecorator` with a injections from a generator, the safer approach is to check `super` to see if there's another implementation of it, in our case, coming from the IIIF Print gem.  If not then carry on as usual.


alishaevn (approved):

kirkkwang:
@alishaevn i had to pull `main` to pass CI, can you take a look again please?

alishaevn (approved):

PR #1859, opened 2022-12-13
Groups with Roles feature

bkiahstroud:
Introduce ability to add roles to groups. Group members inherit the
permissions attached to the group's roles.

Notable changes:
- Several new Roles for Work, Collection, and User management
- Role permission integration with Hyrax permission logic (Sipity,
  access controls, CanCan, etc.)
- Hyku::Group replaced with already existing Hyrax::Group
- Easily assign roles to individual users
- Extended UI
- Both pre-configured and customizable groups
- Scripts to migrate existing Hyku applications
- More

This work is a joint contribution of PALNI and PALCI's Hyku for Consortia project.

PR #1898, opened 2023-01-25
i395-cross-ref-overrides

aprilrieger:
cross reffed versions against v3.4.2 and updated the version as none of them changed from their previous version to v3.4.2


bkiahstroud:
@aprilrieger I think your local copy of the `groups-with-roles-feature` branch might be out of date; the versions in these comments were recently updated in this commit: https://github.com/samvera/hyku/pull/1859/commits/c27ac8ed8909e189e624c8fbb411ce358ee97b45

PR #1899, opened 2023-01-26
rubocop fixes

labradford:
Fixes #396 in the Scientist SoftServ repository https://github.com/scientist-softserv/palni-palci/issues/396

Reverts the rubocop-todo.yml file and fixes resulting rubocop violations


bkiahstroud (dismissed):

app/models/hyrax/group.rb
```diff
@@ -90,7 +90,7 @@ def description_label
       label
     end
```
jeremyf, line 91:
```suggestion
    end

    alias is_default_group? default_group?
    
    self.deprecation_horizon = 'Hyku v6.0.0'
    deprecation_deprecate :is_default_group?
```

jeremyf: I'm thinking that we want to provide some support for those who might have customizations around the `:is_default_group?`.

This gives some space, notifying folks of the change.

jeremyf (approved):

app/models/hyrax/group.rb
bkiahstroud: I like that pattern. For this change specifically though, it's merging into a feature branch where this method is new; this method never existed in `main`, so no one has been using it yet

PR #1905, opened 2023-02-01
DO NOT MERGE: for testing I11 deploy iiif print

laritakr:
use for testing https://github.com/scientist-softserv/iiif_print/pull/83


app/models/generic_work.rb
```diff
@@ -2,6 +2,8 @@
 
 class GenericWork < ActiveFedora::Base
   include ::Hyrax::WorkBehavior
+  include IiifPrint::SetChildFlag
+  include IiifPrint.model_configuration(pdf_split_child_model: Image)
```
ShanaLMoore, line 6: Should line 6 go on the Image model?

ShanaLMoore: or, is this saying that the child work types for a generic work parent will be images? @laritakr

PR #1907, opened 2023-02-03
upgrade to hyrax 3.5, make graph indexer on by default

orangewolf:
This skips the very slow nested indexing and will be default behavior in Hyrax 4.0. The only downside is it will not work if the Solr Cloud shard count is > 1

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -264,7 +264,7 @@ module Hyrax
 
         @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
         # we don't have to reindex the full graph when updating collection
-        @collection.reindex_extent = Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX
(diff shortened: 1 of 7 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -264,7 +264,7 @@ def update_active_fedora_collection
 
         @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
         # we don't have to reindex the full graph when updating collection
-        @collection.reindex_extent = Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX
+        @collection.try(:reindex_extent=, Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX)
```
laritakr, line 267: should a similar change happen in app/jobs/reindex_collections_job.rb ?

aprilrieger (approved):

app/controllers/hyrax/dashboard/collections_controller.rb
orangewolf: 💯

laritakr: (or is the job obsolete now?)

jeremyf (dismissed):

PR #1912, opened 2023-02-08
DO NOT MERGE: I98 iiif print fileset row

ShanaLMoore:
this branch was created by doing the following. 

Merge main into hyku-iiif-base
Created this branch
Update iiif_print gem to pull in LaRita's work
Run iiif_print:install
Run migrations and yarn


laritakr:
THIS is the branch that contains the migrations which are deployed to the iiif_print staging site. The migrations and schema here should be used for future deploys to that site.

PR #1918, opened 2023-02-20
bring back a production worthy docker-compose example

orangewolf:
Needs a few things
- check readme to see if it needs an update
- verify that this comes up first try from a clean env

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -1,0 +1,199 @@
+version: '3.8'
+
+x-app: &app
+  image: ghcr.io/samvera/hyku:${TAG:-latest}
+  env_file:
+    - .env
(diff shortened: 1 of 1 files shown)


bkiahstroud (commented):
If this file is only supposed to be an example, should we rename it from `docker-compose.production.yml` to `docker-compose.production.example.yml` for clarity?

aprilrieger:
Checked in the README and it does mention the docker-compose.production.yaml file, it also has a blurb on how to deploy via helm on kubernetes cluster.
documentation links:
- https://github.com/samvera/hyku#with-kubernetes
- https://github.com/samvera/hyku#with-docker

bkiahstroud (approved):

PR #1924, opened 2023-02-25
Update FileSetDerivatives override

kirkkwang:
`FileSetDerivatives` used the `#class_eval` method of overriding.  We have since favored the decorator pattern instead.  This commit will also add a test for `CreateDerivativesJob` and a couple of new fixtures to make sure we are getting the thumbnail sizes.


app/services/hyrax/file_set_derivatives_service_decorator.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# OVERRIDE: increase size of thumbnails for better viewing in Neutral theme home page
```
alishaevn, line 3:
```suggestion
# OVERRIDE: Hyrax v2.9.0 to increase size of thumbnails for better viewing in Neutral theme home page
```

alishaevn (changes requested):

orangewolf (approved):

kirkkwang:
Change already made

PR #1946, opened 2023-05-03
Some Build Adjustments

orangewolf:
- fill in java_opts by default
- remove need for sed now that ruby is fixed in hyrax image
- update hyrax-doi branch to main as hyrax_upgrade is merged

@samvera/hyku-code-reviewers


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_02_08_001228) do
+ActiveRecord::Schema.define(version: 2023_01_31_202855) do
```
alishaevn, line 13: why revert the schema?

.env
```diff
@@ -13,6 +13,7 @@ FCREPO_PORT=8080
 FCREPO_REST_PATH=rest
 INITIAL_ADMIN_EMAIL=admin@example.com
 INITIAL_ADMIN_PASSWORD=testing123
+JAVA_OPTS=-Xmx4g -Xms1g
```
alishaevn, line 16: should this include a space?

docker-compose.yml
```diff
@@ -132,6 +129,8 @@ services:
 
   web:
     <<: *app
+    # Uncommend command to access container with out starting Rails. Useful for debugging
```
alishaevn, line 132:
```suggestion
    # Uncomment command to access container with out starting Rails. Useful for debugging
```

alishaevn (changes requested):

.env
orangewolf: yes, it needs the space

db/schema.rb
orangewolf: i dont see a migration here. I dont know how the schema got rolled forward but the number seems wrong

PR #1958, opened 2023-06-14
introduce per tenant analytics

alishaevn:
# Story
allow for per tenant analytics. still has a fallback to use the project default values though.
you can establish an analytics account for each tenant, or only some tenants while allowing the others to use the defaults in the environment variables.

related
- https://github.com/scientist-softserv/palni-palci/issues/408
- https://github.com/scientist-softserv/palni-palci/issues/409

@samvera/hyku-code-reviewers


jeremyf:
This work is superseded by:

- https://github.com/scientist-softserv/palni-palci/issues/910

PR #1961, opened 2023-06-23
Contribute back reindex and tenants rake tasks

summer-cook:
As per slack convo: https://assaydepot.slack.com/archives/C0311DNF3MG/p1687538076547959

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -1,0 +1,61 @@
+namespace :hyku do
+  desc "reindex just the works in the background"
+  task reindex_works: :environment do
+    Account.find_each do |account|
+      puts "=============== #{account.name}============"
+      next if account.name == "search"
+      switch!(account)
+      begin
+        Site.instance.available_works.each do |work_type|
+          title = "~ #{account.name} - #{work_type}"
+          klass = work_type.constantize
+          progressbar = ProgressBar.create(total: klass.count, title: title, format: "%t %c of %C %a %B %p%%")
+          klass.find_each do |w|
+            ReindexWorksJob.perform_later(w)
+            progressbar.increment
+          end
+        end
+      rescue => e
+        puts "(#{e.message})"
+      end
+    end
+  end
+
+  desc "reindex just the collections in the background"
+  task reindex_collections: :environment do
+    Account.find_each do |account|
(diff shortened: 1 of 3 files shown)


lib/tasks/index.rake
```diff
@@ -0,0 +1,61 @@
...
+  desc "reindex just the filesets in the background"
+  task reindex_filesets: :environment do
+    Account.find_each do |account|
+      puts "=============== #{account.name}============"
+      next if account.name == "search"
+      switch!(account)
+      title = "~ #{account.name}"
+      progressbar = ProgressBar.create(total: FileSet.count, title: title, format: "%t %c of %C %a %B %p%%")
+      begin
+        FileSet.find_each do |file_set|
```
ShanaLMoore, line 52: You'll need to bring this job over, I don't think it's in hyku.

lib/tasks/tenants.rake
```diff
@@ -0,0 +1,61 @@
+namespace :hyku do
```
ShanaLMoore, line 1: Is there a difference between this and the index.rake file?

lib/tasks/index.rake
```diff
@@ -0,0 +1,61 @@
...
+  task reindex_works: :environment do
+    Account.find_each do |account|
+      puts "=============== #{account.name}============"
+      next if account.name == "search"
+      switch!(account)
+      begin
+        Site.instance.available_works.each do |work_type|
+          title = "~ #{account.name} - #{work_type}"
+          klass = work_type.constantize
+          progressbar = ProgressBar.create(total: klass.count, title: title, format: "%t %c of %C %a %B %p%%")
```
ShanaLMoore, line 12: In Adventist, I was able to run this task locally but wasn't able to in prod without requiring progressbar at the top of this file. 

ref: https://github.com/scientist-softserv/adventist-dl/pull/465/files

summer-cook: thank you i will look at this!

lib/tasks/tenants.rake
summer-cook: oops, i was doing this way too fast. I just updated it!

summer-cook:
@ShanaLMoore ok - i just updated everything ･ᴗ･ thank you for your review

PR #1970, opened 2023-07-20
Bulkrax importing work

orangewolf:
Work to make Bulkrax work properly for a standard CSV file.


.env
```diff
@@ -1,12 +1,14 @@
 CHROME_HOSTNAME=chrome
+COMPOSE_DOCKER_CLI_BUILD=1
```
orangewolf, line 2: @aprilrieger this is the secret to eternal youth. or maybe it was faster builds. it was one of those two things

PR #1976, opened 2023-08-05
SSO Interface and Setup for SAML, OpenID Connect and CAS

orangewolf:
1) Ability to add SSO providers in the UI
Screenshot 2023-08-06 at 01 30 35
Screenshot 2023-08-06 at 01 30 43

2) If no providers the log in button goes to the login form
3) If one provider only is added, the login button auto sends you to the sso provider
4) If more than one provider is added it takes you to a selection page
https://share.getcloudapp.com/YEu4bPoX


docker-compose.yml
K8Sewell, line 1: Should this be flipped back?  Ignore me if not so.

orangewolf: good catch Kait!

PR #1979, opened 2023-08-24
resolve-rails-command-servercommand-error

alishaevn:
cherry-pick commit `96a88ab` from hyrax.

when starting the server today I recevied an uninitialized constant Rails::Command::ServerCommand::APP_PATH (NameError) error. a search of the samvera slack returned this message: https://samvera.slack.com/archives/C75LH2VEH/p1691595910445459. @dlpierce responded and I found https://github.com/samvera/hyrax/pull/6146. I cherry picked the merge commit, but I am still unable to start up this app.

we didn't have either of these two files anyway, so I don't think this would have been the solution. but leaving them here as breadcrumbs.

Screenshot 2023-08-24 at 2 12 55 PM

### DIFF
--- a/.regen
+++ b/.regen
@@ -1,0 +1,2 @@
+# When updating CI regen seed, set to current date.
+2023-08-09T14:34:59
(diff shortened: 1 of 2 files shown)


alishaevn:
user error. I had two panes open, one was in hyku and one was in hyrax. 🤦🏾‍♀️

PR #1988, opened 2023-09-21
make sure layers are flattened to fix pdf tripple issue

orangewolf:


lib/hydra/derivatives/processors/image_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+# Fix PDF tripple issue
+
+module Hydra
+  module Derivatives
+    module Processors
+      module ImageDecorator
+        protected
+
+          # When resizing images, it is necessary to flatten any layers, otherwise the background
+          # may be completely black. This happens especially with PDFs. See #110
```
alishaevn, line 12: idk where this code was copied over from, but can this `#110` reference be fleshed out to point at the issue? in the event we need to reference it in the future.

alishaevn (changes requested):

lib/hydra/derivatives/processors/image_decorator.rb
laritakr: This is the issue: https://github.com/samvera/hydra-derivatives/issues/110

alishaevn:
```suggestion
          # may be completely black. This happens especially with PDFs. See https://github.com/samvera/hydra-derivatives/issues/110
```

alishaevn (approved):

PR #1993, opened 2023-09-21
add blacklight advances search, blacklight date ranges and order already

orangewolf:
to the gem file and configure them


ShanaLMoore:
I attempted to test this from hyku proper since adventist knapsack currently can't create works. 

I see the date range facet displayed but entering dates creates an error: 

image


image

orangewolf:
added with a different merge

PR #1995, opened 2023-09-21
register and add missing methods to avoid content block errors

ShanaLMoore:
on adventist_dev we received an error when visiting the tenant page: ie dev.hyku.test

image


ShanaLMoore:
refer to PR 1994. Rob simultaneously fixed this

laritakr (approved):

PR #1997, opened 2023-09-22
add option to support good job for background jobs instead of sidekiq

orangewolf:


ShanaLMoore:
This is working for adventist_knapsack (which should be using good_jobs instead of sidekiq)

image

ShanaLMoore (approved):

ShanaLMoore:
This looks fine but I wasn't able to get good jobs running in adventist knapsack. The jobs would queue in the terminal, but even w the worker container running they were not moving or showing up in the good_job dashboard. When I switched the env var to sidekiq, everything worked fine. 

Testing this branch now to isolate if it's an adventist_knapsack problem or not. cc @orangewolf

ShanaLMoore (changes requested):
I pulled down this branch and see sidekiq when visting /jobs even tho the env var says good_job

orangewolf:
@ShanaLMoore do you agree this is working now?

orangewolf:
this got merged some other way

PR #2001, opened 2023-09-29
Load Knapsack factories

kirkkwang:
This commit will load the Knapsack factories in rails_helper if they are present.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -41,6 +41,10 @@ require 'shoulda/matchers'
 #
 Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
 
+# loads Knapsack factories
+knapsack_factory_path = Rails.root.join('..', 'spec', 'factories', '*.rb')
+Dir[knapsack_factory_path].each { |f| require f } if Dir.exist?(File.dirname(knapsack_factory_path))
+
 # Ensure the Hyrax::Admin constant is loaded. Because testing is done using autoloading,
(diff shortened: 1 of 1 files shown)


spec/rails_helper.rb
```diff
@@ -41,6 +41,10 @@
 #
 Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
 
+# loads Knapsack factories
+knapsack_factory_path = Rails.root.join('..', 'spec', 'factories', '*.rb')
```
orangewolf, line 45: If you use the knapsack path (see lib/hyku_knapsack.rb) you don’t need to deal with this relative path thing. Also you should handle us directories with ** and lastly you don’t need the if statement if you are doing an each. Dir[i_dont_exist] == [] and [].each is a no op already

orangewolf: Also, why do you need factories from the knapsack in the main app? Aren’t knapsack tests run with their own rails helper in the directory one up?

kirkkwang: If that's then case then i must not be running the specs correctly, i just bash into my web container and run `rspec` so if there's something else I need to do then my mistake.

https://github.com/samvera/hyku/blob/load-knapsack-factories/spec/abilities/work_ability_spec.rb#L17
in particular was running through all the work types and erroring out because there were no factories for Adventist work types in Hyku

kirkkwang:
from the messages on this PR seems like I was doing something wrong, closing this PR

PR #2009, opened 2023-10-04
:bug: subject can't be blank for the contact form

ShanaLMoore:
Previously there was a bug because even if you typed in a subject, the contact form would error saying that it was blank.

image

Part of issue:
- https://github.com/scientist-softserv/adventist-dl/issues/608


app/controllers/hyrax/contact_form_controller.rb
```diff
@@ -50,6 +50,7 @@ def create
       # not spam, form is valid, and captcha is valid
       @captcha.values[:category] = params[:contact_form][:category]
       @captcha.values[:contact_method] = params[:contact_form][:contact_method]
+      @captcha.values[:subjects] = params[:contact_form][:subject]
```
ShanaLMoore, line 53:
```suggestion
      @captcha.values[:subject] = params[:contact_form][:subject]
```

jeremyf (approved):

PR #2030, opened 2023-10-23
:bug: Depositor role cannot deposit into admin set fix

ShanaLMoore:
We implemented a fix for this bug in PALS and are contributing it back to Hyku.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/864

Prior to this commit, even if a user was granted the role of depositor to an admin set, they could not add a work.  If they were added in a group that had the depositor role, they still were not able to create a work.

Now, users who have the depositor role or in a group with the depositor role can deposit to admin sets.

# Screenshots / Video

https://github.com/scientist-softserv/palni-palci/assets/19597776/dcb324fa-02f5-4743-a804-ea18709bce73

# NOTE

By default Hyku doesn’t actually include everyone_can_create_curation_concerns in the ability logic. It’s behind an [ENV-controlled feature flipper](https://github.com/samvera/hyku/blob/main/app/models/ability.rb#L22-L24), which [by default is true](https://github.com/samvera/hyku/blob/776cf9e9cd3d50b480779c78ee1c5cf0496f078b/.env#L52) (true in this case does not include everyone_can_create_curation_concerns since true == more restrictive permissions)

### DIFF
--- a/app/models/concerns/hyrax/ability/work_ability.rb
+++ b/app/models/concerns/hyrax/ability/work_ability.rb
@@ -15,8 +15,23 @@ module Hyrax
             doc = permissions_doc(id)
             all_work_types_and_files.include?(doc.hydra_model)
           end
-        elsif work_depositor?
+        elsif work_depositor? || admin_set_with_deposit?
           can %i[create], all_work_types_and_files
+        end
+
+        # OVERRIDE HYRAX 3.5.0 to return false if no ids are found
+        # @return [Boolean] true if the user has at least one admin set they can deposit into.
+        def admin_set_with_deposit?
+          ids = PermissionTemplateAccess.for_user(ability: self,
+                                                  access: ['deposit', 'manage'])
+                                        .joins(:permission_template)
+                                        .select(:source_id)
+                                        .distinct
+                                        .pluck(:source_id)
+
+          return false if ids.empty?
+
+          Hyrax.custom_queries.find_ids_by_model(model: Hyrax::AdministrativeSet, ids: ids).any?
         end
       end
     end
--- a/spec/features/create_work_spec.rb
+++ b/spec/features/create_work_spec.rb
@@ -18,4 +18,54 @@ RSpec.describe 'Creating a new Work', type: :feature, clean: true do
     click_link "Share Your Work"
     expect(page).to have_button "Create work"
   end
+
+  context 'as a user with no roles' do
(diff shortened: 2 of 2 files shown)


app/models/concerns/hyrax/ability/work_ability.rb
```diff
@@ -15,9 +15,24 @@ def work_roles
...
+          ids = PermissionTemplateAccess.for_user(ability: self,
+                                                  access: ['deposit', 'manage'])
+                                        .joins(:permission_template)
+                                        .select(:source_id)
+                                        .distinct
+                                        .pluck(:source_id)
+
+          return false if ids.empty?
+
+          Hyrax.custom_queries.find_ids_by_model(model: Hyrax::AdministrativeSet, ids: ids).any?
```
bkiahstroud, line 34: What was the error & error message being thrown here? It was only happening outside any tenants, right?

ShanaLMoore: ref this hot fix: https://github.com/scientist-softserv/palni-palci/pull/885

app/models/concerns/hyrax/ability/work_ability.rb
```diff
@@ -15,9 +15,24 @@ def work_roles
...
             all_work_types_and_files.include?(doc.hydra_model)
           end
-        elsif work_depositor?
+        elsif work_depositor? || admin_set_with_deposit?
           can %i[create], all_work_types_and_files
         end
+
+        # OVERRIDE HYRAX 3.5.0 to return false if no ids are found
+        # @return [Boolean] true if the user has at least one admin set they can deposit into.
+        def admin_set_with_deposit?
```
bkiahstroud, line 24: The `#admin_set_with_deposit?` method is defined inside the `#work_roles` method

bkiahstroud (changes requested):
@ShanaLMoore I spotted an indentation issue, but once that's fixed this should be ready to go!

app/models/concerns/hyrax/ability/work_ability.rb
ShanaLMoore: oh man, good catch! 🙈

ShanaLMoore:
Replacing this PR with [this](https://github.com/samvera/hyku/pull/2032). I noticed the branch name was wrong.

PR #2044, opened 2023-12-06
pin-json-canonicalization-0.3.1

aprilrieger:
Pinning the json-canonicalization to 0.3.1

# The maintainers yanked 0.3.2 version (see https://github.com/dryruby/json-canonicalization/issues/2)
gem 'json-canonicalization', "0.3.1"

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,7 +13,10 @@
 ActiveRecord::Schema.define(version: 2023_08_04_073106) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "hstore"
+  enable_extension "pgcrypto"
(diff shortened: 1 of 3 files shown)


db/schema.rb
aprilrieger, line 1: @bkiahstroud this is the schema updates that were created after bringing the application up fully with dc up, which runs bundle and migrations.

bkiahstroud: I don't think these should break anything, but I also don't think they belong in this PR. Can we revert the change to this file and open a separate PR for it? (I can take that as a TODO)

aprilrieger: I agree with you, just wanted you to see them.

PR #2052, opened 2023-12-13
♻️ Introduce Hyku::Application.theme_view_path_roots

jeremyf:
With the introduction of [HykuKnapsack][1], we are adjusting how we create instances of Hyku.  Namely we don't clone Hyku but instead we incorporate Hyku as a submodule into a Knapsack.

The Knapsack is a Rails engine that is mounted in the Hyku application. What this means is that when we want to

Prior to this commit, the only way to adjust themed views would have been to add them to the Rails application (e.g. Hyku) directly.  Which would work in a non-Knapsack ecosystem.

However, with Knapsack we need a means of saying "Hey, for themes we want to be able to add/adjust views within the knapsack."  Hence this change.

[1]: https://github.com/samvera-labs/hyku_knapsack

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

@samvera/hyku-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -505,8 +505,10 @@ module Hyrax
       def inject_show_theme_views
         if show_page_theme && show_page_theme != 'default_show'
           original_paths = view_paths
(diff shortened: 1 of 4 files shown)


kirkkwang:
i've already merged your theme fixes into this branch

PR #2054, opened 2023-12-13
Use github actions in favor of circle ci

ShanaLMoore:
Per Jeremy, 

there is a desire from the community to switch to github actions instead of circle ci. Doing so will save us a lot of $$.

should the softserv actions repo be forked into samvera labs?


.circleci/config.yml
```diff
@@ -98,3 +98,7 @@ workflows:
           ruby_version: "2.7.8"
           name: "ruby2-7-8"
           solr_config_path: 'solr/conf'
+          filters:
```
ShanaLMoore, line 101: Remove this file?

ShanaLMoore: And delete the circle ci webhook fromt his repo once this is merged into main.

PR #2062, opened 2023-12-15
🧹 Hyrax 5 get additional specs passing

laritakr:
Ref https://github.com/scientist-softserv/hykuup_knapsack/issues/56

- Fix rails_helper
- Remove `google_analytics_id` references from `reload_library_config`, pulling back [a piece of the pals analytics work](https://github.com/scientist-softserv/palni-palci/pull/946) to get specs passing. Additional analytics backporting will take place elsewhere.


.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     parameters:
       ruby_version:
         type: string
-        default: 2.7.8
+        default: 3.2.2
```
ShanaLMoore, line 13: I believe Kirk has a commit for this file? If not, I think we may need to update the rails version too.

laritakr: He did not. I will update rails version as well.

ShanaLMoore (approved):

PR #2067, opened 2023-12-18
🧹 Hyrax 5 upgrade additional specs

laritakr:
Refs https://github.com/scientist-softserv/hykuup_knapsack/issues/55

Ongoing work to fix broken upgrade specs.


app/controllers/hyrax/contact_form_controller.rb
```diff
@@ -1,71 +1,65 @@
...
-      @captcha.values[:contact_method] = params[:contact_form][:contact_method]
-      @captcha.values[:subject] = params[:contact_form][:subject]
-      @contact_form = model_class.new(@captcha.values)
-      if @contact_form.valid? && @captcha.valid?
+      # not spam and a valid form
+      if @contact_form.valid?
         ContactMailer.contact(@contact_form).deliver_now
         flash.now[:notice] = 'Thank you for your message!'
         after_deliver
+        @contact_form = model_class.new
```
jeremyf, line 54: Interesting that this is made new again

app/models/role.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Role < ApplicationRecord
-  has_many :users, through: :users_roles
+  has_and_belongs_to_many :users, join_table: :users_roles
```
jeremyf, line 4: Umm…these may not be the same macros…let's be cautious.  But there's no `UserRole` model (nor `RoleUser`) so; I guess we'll be alright with this.

app/controllers/hyrax/contact_form_controller.rb
laritakr: I just copied this part from Hyrax. This was not one of the overridden methods in the controller.

app/models/role.rb
laritakr: They are NOT the same... that's why I had to revert it back to this and exclude it in rubocop yml.

app/controllers/hyrax/contact_form_controller.rb
laritakr: Actually I may have removed a portion of overridden code when I pulled in the Hyrax updates... checking this again.

app/controllers/hyrax/contact_form_controller.rb
```diff
@@ -58,14 +59,12 @@ def create
         after_deliver
       else
         flash.now[:error] = 'Sorry, this message was not sent successfully. ' +
-                            @contact_form.errors.full_messages.map(&:to_s).join(", ") +
-                            "" + @captcha.error
+                            @contact_form.errors.full_messages.map(&:to_s).join(", ")
```
jeremyf, line 62: Are we accounting for any captcha errors?

spec/abilities/collection_ability_spec.rb
```diff
@@ -321,7 +321,7 @@
              permission_template: collection.permission_template,
              agent_type: 'user',
              agent_id: user.user_key)
-      collection.reset_access_controls!
+      collection.permission_template.reset_access_controls_for(collection:)
```
jeremyf, line 324: Rhetorical/Venting: Upstream change brings the Yuck.  Shouldn't the collection be able to do this?

I mean if it gets things working.  But saying `A` tells `B` to do `C` with `A` is a very circuitous way; but

spec/factories/collections.rb
```diff
@@ -152,56 +141,26 @@
   #                      :allow_multiple_membership, # OR :not_allow_multiple_membership
   #                    ] }
   #   let(:collection_type) { create(:collection_lw_type, settings) }
-  #   let(:collection) { build(:collection_lw, collection_type_gid: collection_type.gid) }
```
jeremyf, line 155: Hmm…this documentation looks like it might be nice to have.

app/controllers/hyrax/contact_form_controller.rb
laritakr: I don't know... I'm just putting back what was there before that I removed accidentally.

spec/factories/collections.rb
laritakr: Oh, I removed it by overwriting with current factory code, but missed copying the current documentation. I'll add that back in.

laritakr: which documentation? This portion of the code was basically a copy of Hyrax's so I copied over their current documentation with their current factory code.

laritakr: Basically, the nesting logic and documentation was removed.

laritakr:
293 failing specs left.

kirkkwang (approved):
Looks good to me!

PR #2077, opened 2023-12-19
🧹 Favor Sipity::Entity function over to_sipity_entity

jeremyf:
The Sipity conversion methods are now explicit functions (much like the `Array()` function).

- https://github.com/samvera/hyrax/blob/5aaa568d348a180add2a1337d9d794b740703df8/app/models/sipity.rb#L20

### DIFF
--- a/spec/requests/work_approval_permissions_spec.rb
+++ b/spec/requests/work_approval_permissions_spec.rb
@@ -39,11 +39,11 @@ RSpec.describe 'Work approval permissions', type: :request, singletenant: true, 
     end
 
     it 'can approve a work' do
-      expect(work.to_sipity_entity.workflow_state.name).to eq('pending_review')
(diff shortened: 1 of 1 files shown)


kirkkwang (approved):

orangewolf (commented):
to me, this is stepping backward. having as a method on the object allows the work class to handle the sipity entity in different ways depending on the objects class. if you use this method, the Sipity::Entity has to know about an unknown number to classes. Can you help me understand the api benifit of changing this?

jeremyf:
@orangewolf Hyrax removed the `#to_sipity_entity` method some where in the Hyrax 4 era.  

https://github.com/samvera/hyrax/blob/f7a3fcde5cc08a72629a3a05ba27b3c214eff590/app/models/concerns/hyrax/suppressible.rb#L26

https://github.com/samvera/hyrax/commit/9f7b2c689deaee5532b481f3a5c97ea009f95f41 removes this

PR #2090, opened 2023-12-20
🧹 Clean up actors

kirkkwang:
Just found one actor that was overriding Hyrax.  This commit will switch
it to a decorator.

Ref:
  - https://github.com/scientist-softserv/hykuup_knapsack/issues/44

### DIFF
--- a/app/actors/hyrax/environment_decorator.rb
+++ b/app/actors/hyrax/environment_decorator.rb
@@ -1,25 +1,21 @@
 # frozen_string_literal: true
 
-# OVERRIDE Hyrax 2.9 to add in import flag
+# OVERRIDE Hyrax v5.0.0rc2 to add in import flag
+
 module Hyrax
   module Actors
-    class Environment
+    module EnvironmentDecoractor
       # @param [ActiveFedora::Base] curation_concern work to operate on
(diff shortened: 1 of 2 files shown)


app/actors/hyrax/environment_decorator.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Actors
+    module EnvironmentDecoractor
+      extend ActiveSupport::Concern
+
+      class_methods do
+        # @param [ActiveFedora::Base] curation_concern work to operate on
+        # @param [Ability] current_ability the authorizations of the acting user
+        # @param [ActionController::Parameters] attributes user provided form attributes
+        def new(curation_concern, current_ability, attributes, importing = false)
+          @importing = importing
```
jeremyf, line 15: Do we need a super method here?

jeremyf: Let's make this not a class method, but instead the initialize method.

```ruby
def initialize(curation_concern, current_ability, attributes, importing = false)
  @importing = importing
  super(curation_concern, current_ability, attributes)
end
```

jeremyf (approved):

PR #2109, opened 2023-12-21
🧹 Clean up services

kirkkwang:
This commit will reconcile the services that are overrides for Hyrax with the Hyrax 5.0.0rc2 version.  There were a number of overrides that were changed to the decorator pattern.

Ref:
  - https://github.com/scientist-softserv/hykuup_knapsack/issues/46


app/services/hyrax/collection_types/create_service_decorator.rb
```diff
@@ -209,7 +118,7 @@ def self.create_user_collection_type
       #   If calling from Abilities, pass the ability.
       #   If you try to get the ability from the user, you end up in an infinite loop.
       # rubocop:disable Metrics/MethodLength
-      def self.add_default_participants(collection_type_id)
+      def add_default_participants(collection_type_id)
```
jeremyf, line 121: I'd feel better if this were in a submodule `ClassMethods`

Reading this module it would be easy to think it's an instance method.  But we prepend it to a singleton class.

app/services/hyrax/indexes_thumbnails_decorator.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax v5.0.0rc2 to make collection thumbnails uploadable
+
+module Hyrax
+  module IndexesThumbnailsDecorator
+    # Returns the value for the thumbnail path to put into the solr document
+    def thumbnail_path
+      if object.class == Collection && UploadedCollectionThumbnailPathService.uploaded_thumbnail?(object)
```
jeremyf, line 9: What about `object.collection?`. We have `Valkyrie::Collection` coming soon, and class level comparison is always dubious.

I strongly prefer `object.is_a?` over `object.class ==` due to `is_a?` working with modules and such things.

kirkkwang: That's a good idea, so instead of `object.class == Collection` we do `object.collection?` which, if I'm understanding correctly, would also catch cases of `Valkyrie::Collection`, is that correct?

app/services/hyrax/quick_classification_query_decorator.rb
```diff
@@ -0,0 +1,29 @@
...
+
+# OVERRIDE Hryax v5.0.0rc2
+
+module Hyrax
+  module QuickClassificationQueryDecorator
+    # OVERRIDE: only use work types that are enabled in the current tenant
+    # @param [::User] user the current user
+    # @param [#call] concern_name_normalizer (String#constantize) a proc that translates names to classes
+    # @param [Array<String>] models the options to display, defaults to everything.
+    def initialize(user,
```
jeremyf, line 11: I think we could do:

```ruby
def initialize(user, models: Site.instance.available_works, **kwargs)
  super(user, **kwargs.merge(models: models))
end
```

That way we don't need to know about the other implementation details.

app/services/hyrax/thumbnail_path_service_decorator.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax v5.0.0rc2 - use site defaults instead of app wide defaults
+module Hyrax
+  module ThumbnailPathServiceDecorator
+    def default_image
+      Site.instance.default_work_image&.url || super
```
jeremyf, line 7: Yummy refactor!

app/services/hyrax/thumbnail_path_service_decorator.rb
```diff
@@ -0,0 +1,12 @@
...
+    end
+  end
+end
+
+Hyrax::ThumbnailPathService.singleton_class.send(:prepend, Hyrax::ThumbnailPathServiceDecorator)
```
jeremyf, line 12:
```suggestion
Hyrax::ThumbnailPathService.singleton_class.send(:prepend, Hyrax::WorkThumbnailPathServiceDecorator)
```

app/services/hyrax/thumbnail_path_service_decorator.rb
jeremyf: It looks like this is a complete duplication of the `WorkThumbnailPathServiceDecorator`

app/services/hyrax/thumbnail_path_service_decorator.rb
jeremyf: Why the above change?  We can get rid of another module and reduce duplication of knowledge.

spec/services/hyrax/work_thumbnail_path_service_decorator_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::WorkThumbnailPathService, type: :decorator do
```
jeremyf, line 3: 😄

app/services/hyrax/thumbnail_path_service_decorator.rb
kirkkwang: Hmm, are you suggesting to remove the `app/services/hyrax/thumbnail_path_service_decorator.rb` and put `Hyrax::ThumbnailPathService.singleton_class.send(:prepend, Hyrax::WorkThumbnailPathServiceDecorator)` in `app/services/hyrax/work_thumbnail_path_service_decorator.rb`?

jeremyf: Yeah.

kirkkwang: I think my concern with that is the coupling of both of services.  In the `super` Hyrax actually has two different returns

`WorkThumbnailPathService` => `ActionController::Base.helpers.image_path 'work.png'`
and
`ThumbnailPathService` => `ActionController::Base.helpers.image_path 'default.png'`

If in an event that we want to change Hyku's thumbnail path and work thumbnail path then it would be easier to find and already set up.  But I'm not holding on to this idea too strongly.

jeremyf: I'm alright with what you have.  But consider, your tests show what `super` means and the code looks like duplication (because it is; it just overlays different things, which is the nature of polymorphism/encapsulation)

jeremyf (approved):

PR #2112, opened 2023-12-22
🤖 Attempt at enabling rerun failed tests

kirkkwang:
This is an attempt at enabling rerun failed tests in CircleCI so we don't have to rerun the entire test suite when a test fails.


kirkkwang:
Meh, not really working as I thought it would

PR #2113, opened 2023-12-22
♻️ Account for observed customizations

jeremyf:
Dear reviewer, put on your reading glasses.

This commit looks at the newly refactored
`Hyrax::IiifAv::DisplaysContentDecorator` as well as the current state of [PALS's Hyrax::IiifAv::DisplaysContentDecorator][1] and attempts to account for the variances between the two by introducing configurations.

Yes, we could port this to the hyrax-iiif_av gem, but for now that would not solve the underlying issue of how we've been handling things.

[1]: https://github.com/scientist-softserv/palni-palci/blob/8754556c0225ce9f04674c1ffac6403586fd65f4/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb

### DIFF
--- a/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb
+++ b/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb
@@ -8,6 +8,47 @@ module Hyrax
     #   request.base_url => hostname
     # also to remove #auth_service since it was not working for now
     module DisplaysContentDecorator
+      extend ActiveSupport::Concern
+
+      included do
+        ##
+        # @!group Class Attributes
+
+        ##
+        # @!attribute iiif_audio_labels_and_mime_types [r|w]
+        #   @return [Hash<String,String>] Hash of valid audio labels and their mime types.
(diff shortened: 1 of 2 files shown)


app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb
```diff
@@ -8,6 +8,47 @@ module IiifAv
...
+        # @!group Class Attributes
+
+        ##
+        # @!attribute iiif_audio_labels_and_mime_types [r|w]
+        #   @return [Hash<String,String>] Hash of valid audio labels and their mime types.
+        class_attribute :iiif_audio_labels_and_mime_types, default: { "ogg" => "audio/ogg", "mp3" => "audio/mpeg" }
+
+        ##
+        # @!attribute iiif_video_labels_and_mime_types [r|w]
+        #   @return [Hash<String,String>] Hash of valid vidoe labels and their mime types.
```
kirkkwang, line 24: just a typo `video`

kirkkwang (approved):
Cool, i think it looks good

PR #2116, opened 2023-12-22
🧹 Comment out check for reports

jeremyf:
For this spec, if `Hyrax.config.analytics?` is enabled we need a legato
user to query the analytics.  Which creates a weird dependency that we
may not want in this integration test.

(This stuff is best tested in view tests; but here we are)

I opted to not remove the spec but instead favor the
`Hyrax.config.analytics?` which, if it is true likely means we have ENV
vars that are in play.


ShanaLMoore:
Hey Jeremy, 

This one may be worth discussing. 

Kirk and I addressed the failing spec in another [way](https://github.com/samvera/hyku/pull/2118). We had noticed that our local locale file was not taking precedence over the hyrax gem's version. Correcting that issue makes the spec pass. So is adding this conditional still worth while?

PR #2147, opened 2024-01-09
Valkyrize the reindexer jobs

laritakr:
Ref https://github.com/scientist-softserv/hykuup_knapsack/issues/35

The current reindex wouldn't work for Valkyrie resources. However this change only picks up the new collection types, ignoring the ActiveFedora types.

Starting this as a draft to note the need for SOME change, even though this may not be adequate.

### DIFF
--- a/app/jobs/reindex_admin_sets_job.rb
+++ b/app/jobs/reindex_admin_sets_job.rb
@@ -2,6 +2,8 @@
 
 class ReindexAdminSetsJob < ApplicationJob
   def perform
-    AdminSet.find_each(&:update_index)
+    AdministrativeSet.find_each do |admin_set|
+      ReindexItemJob.perform_later(admin_set)
+    end
   end
 end
--- a/app/jobs/reindex_collections_job.rb
+++ b/app/jobs/reindex_collections_job.rb
@@ -2,7 +2,7 @@
 
 class ReindexCollectionsJob < ApplicationJob
   def perform
-    Collection.find_each do |collection|
+    PcdmCollection.find_each do |collection|
       ReindexItemJob.perform_later(collection)
     end
   end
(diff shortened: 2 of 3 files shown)


app/jobs/reindex_item_job.rb
```diff
@@ -2,6 +2,6 @@
 
 class ReindexItemJob < ApplicationJob
   def perform(item)
-    item.update_index
+    Hyrax.index_adapter.save(resource: item)
```
ShanaLMoore, line 5:
```suggestion
    Hyrax.query_service.find_all_of_model(model: item.class).each do |resource|
      Hyrax.index_adapter.save(resource: item)
    end
```

Should this turn into the above, based on our convo with Rob? Would this work if there are existing active fedora objects?

ShanaLMoore: I have to blow away my env to confirm ^^ checking now though...

ShanaLMoore (approved):

app/jobs/reindex_item_job.rb
ShanaLMoore: this was the error i received: 
image

app/jobs/reindex_collections_job.rb
```diff
@@ -2,7 +2,7 @@
 
 class ReindexCollectionsJob < ApplicationJob
   def perform
-    Collection.find_each do |collection|
+    PcdmCollection.find_each do |collection|
```
laritakr, line 5: This needs to be change to use the config collection class. Possibly should also reference multiple models?

app/jobs/reindex_admin_sets_job.rb
```diff
@@ -2,6 +2,8 @@
 
 class ReindexAdminSetsJob < ApplicationJob
   def perform
-    AdminSet.find_each(&:update_index)
+    AdministrativeSet.find_each do |admin_set|
```
laritakr, line 5: This needs to be change to use the config collection class. Possibly should also reference multiple models?

PR #2149, opened 2024-01-09
I35 valkyrize hyku routes and redirects

kirkkwang:
## 🧹 Add custom presenters for resource work types

1ff3036c3897c2c5c43dc84bef9e2992030fc7ab

The Hyrax generator does not generate presenters for resource work
types so I'm adding them here and ensuring the respective resource work
type controllers use them.

## 🧹 Add redirects for work types

83e7dab77e6fc2bccd5cf3a053aef1ad6f91158b

This commit will add redirects for the the work types so when viewing or
editing them, the user will be redirected to the Valkyrie (Resource)
version of the work type.  The won't work unless turbolinks is disabled.
We are opting for using an ENV variable to disable turbolinks for now so
it can be easily removed in the future.

## ⚙️ Set TURBOLINKS_FOR_VALKYRIE to false

3f37899439388fca3c7b0b66ccde449b9a4e6060

This allows for the redirects of AF works to Valkyrie works to function
correctly because turbolinks was not allowing the redirect to happen.

Ref:
  - https://github.com/scientist-softserv/hykuup_knapsack/issues/100


app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -1,7 +1,7 @@
-<%# OVERRIDE Hyrax v5.0.0rc2 Adjust permission checks and add analytics button %>
+<%# OVERRIDE Hyrax v5.0.0rc2 Adjust permission checks and add analytics button and add ENV to control turbolinks %>
 <div class="show-actions">
   <% if presenter.editor? %>
-      <%= link_to t('.edit'), edit_polymorphic_path([main_app, presenter]), class: 'btn btn-secondary' %>
+      <%= link_to t('.edit'), edit_polymorphic_path([main_app, presenter]), class: 'btn btn-secondary', data: { turbolinks: ENV.fetch('TURBOLINKS_FOR_VALKYRIE', 'true') } %>
```
jeremyf, line 4: Question: does `ENV.fetch('TURBOLINKS_FOR_VALKYRIE')` return `false`?  Or `"false"`? (string or boolean)

bkiahstroud, line 4: I think all of these `ENV.fetch` lines need to be cast as a boolean ([example](https://github.com/scientist-softserv/palni-palci/blob/b880891b053becbe73d90bbddc8e4c8baae87afb/app/controllers/application_controller.rb#L121)). I _think_ ENV vars are always strings, so `"false"` will actually be truthy

bkiahstroud (changes requested):

app/views/hyrax/base/_show_actions.html.erb
kirkkwang: it does return a string

kirkkwang: Hmmm, I'm not trying to check for truthy or falsey, I'm looking to set it to either string "true" or "false" so it would evaluate to `data: { turbolinks: 'true' }` or `data: { turbolinks: 'false' }`

.env
```diff
@@ -67,3 +67,7 @@ HYKU_RESTRICT_CREATE_AND_DESTROY_PERMISSIONS=true
 # AND comment this out
 HYRAX_ANALYTICS=false
 ##END## Enable Google Analytics
+
+# When set to false, this turns off turbolinks which allows for
+# redirects from AF works to Valyrie works in the edit and show
+TURBOLINKS_FOR_VALKYRIE=false
```
bkiahstroud, line 73: I wish this var had a more descriptive name, but I'm struggling to come up with a great alternative... 

I can imagine a dev seeing "TURBOLINKS_FOR_VALKYRIE", thinking "turbolinks makes things faster, I should set that to `true`", and unknowingly causing problems. 

Also, I think it should have a `HYKU_` prefix since that seems to be the convention.

bkiahstroud: Maybe something along the lines of 

```
HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE=true
```

kirkkwang: I can make it with the HYKU prefix, but I'm also hoping they would read my comment

kirkkwang: Sounds good to me

app/helpers/hyku_helper.rb
```diff
@@ -19,4 +19,8 @@ def admin_host?
   def admin_only_tenant_creation?
     ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ADMIN_ONLY_TENANT_CREATION', false))
   end
+
+  def turbolinks_enabled?
+    ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE', true))
```
bkiahstroud, line 24: When our ENV var is `true`, turbolinks needs to be `false`, and vice versa. Here's my suggestion on how we can do that in a (hopefully) easy-to-understand, descriptive way: 

```suggestion
  def block_valkyrie_redirect?
    !ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE', true))
```

```mermaid
flowchart TD
A1(HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE)
A2[block_valkyrie_redirect?]
A3[turbolinks: false]
A4(Page redirects)
B1(HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE)
B2[block_valkyrie_redirect?]
B3[turbolinks: true]
B4(Page does not redirects)

A1 --> |true|A2 --> |false|A3 --> A4
B1 --> |false|B2 --> |true|B3 --> B4
```

bkiahstroud (changes requested):

.env
```diff
@@ -67,3 +67,7 @@ HYKU_RESTRICT_CREATE_AND_DESTROY_PERMISSIONS=true
 # AND comment this out
 HYRAX_ANALYTICS=false
 ##END## Enable Google Analytics
+
+# When set to false, this turns off turbolinks which allows for
+# redirects from AF works to Valyrie works in the edit and show
+HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE=false
```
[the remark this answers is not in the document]
kirkkwang: Cool, made the change

app/helpers/hyku_helper.rb
kirkkwang: would it be semantically different if we used

```rb
    ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE', false))
```

kirkkwang: so `false` instead of not `true`

app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -1,7 +1,7 @@
-<%# OVERRIDE Hyrax v5.0.0rc2 Adjust permission checks and add analytics button %>
+<%# OVERRIDE Hyrax v5.0.0rc2 Adjust permission checks and add analytics button and add ENV to control turbolinks %>
 <div class="show-actions">
   <% if presenter.editor? %>
-      <%= link_to t('.edit'), edit_polymorphic_path([main_app, presenter]), class: 'btn btn-secondary' %>
+      <%= link_to t('.edit'), edit_polymorphic_path([main_app, presenter]), class: 'btn btn-secondary', data: { turbolinks: turbolinks_enabled? } %>
```
ShanaLMoore, line 4: Where is #turbolinks_enabled defined?

ShanaLMoore: Oh I see from the convo above. It needs to get renamed.

ShanaLMoore:
This may or may not be related to this work. I took this branch for a spin and the routes are working now 🎉 

But when I try to save a GenericWorkResource it fails because [@form](https://github.com/samvera/hyrax/blob/main/app/views/hyrax/base/_form.html.erb) doesn't exist.  We can dig in tomorrow if you'd like. 

image

app/helpers/hyku_helper.rb
bkiahstroud: I don't think so. The second argument of `ENV.fetch` is a fallback value:

```ruby
# If a is nil, use b
ENV.fetch(a, b)
```

So the value should be whatever we want the default to be, which in this case, I think is `true`

ShanaLMoore:
> This may or may not be related to this work. I took this branch for a spin and the routes are working now 🎉
> 
> But when I try to save a GenericWorkResource it fails because [@form](https://github.com/samvera/hyrax/blob/main/app/views/hyrax/base/_form.html.erb) doesn't exist.
> 
> Should valkyrie objects be routed to a different create? We can dig in tomorrow if you'd like.
> image

Looks like the answer is [yes](https://github.com/samvera/hyrax/wiki/Hyrax-Valkyrie-Usage-Guide#controllers)? I'll see if I can implement it to confirm.

app/views/hyrax/base/_show_actions.html.erb
kirkkwang: Ah yes you're right, good catch

app/helpers/hyku_helper.rb
kirkkwang: alright sounds good to me

.env
```diff
@@ -67,3 +67,7 @@ HYKU_RESTRICT_CREATE_AND_DESTROY_PERMISSIONS=true
 # AND comment this out
 HYRAX_ANALYTICS=false
 ##END## Enable Google Analytics
+
+# When set to false, this turns off turbolinks which allows for
+# redirects from AF works to Valyrie works in the edit and show
+HYKU_ALLOW_REDIRECT_TO_VALKYRIE_RESOURCE=true
```
jeremyf, line 73: Non-blocker, but curious why not name this `HYKU_BLOCK_VALKYRIE_REDIRECT`?  There's a method that toggles the truthiness of this value and then we used that toggled value.

kirkkwang: That sounds good to me, I can make the change

bkiahstroud (commented):
@kirkkwang why was the ENV variable removed?

bkiahstroud (commented):
Ah, I see the commit message: 

> This refactor will simply remove the ENV variable have the helper method return false with a TODO.  This should give us enough documentation in the future.

kirkkwang:
> @kirkkwang why was the ENV variable removed?

basically it introduced a lot of extra complexity that wasn't needed, instead we document why in the code comments

bkiahstroud (dismissed):
> basically it introduced a lot of extra complexity that wasn't needed, instead we document why in the code comments

It certainly increases complexity, I agree with you there. However, there are significant benefits to using an `ENV` variable in a situation like this: 

1. We can toggle it on/off in rancher without changing any code. Even if we did want to change it in the code, it would be a single line, which we could do quickly and safely. 
2. Because this is Hyku, other institutions who are mid-migration can use this. In its current form, an institution mid-migration will not be able to upgrade their Hyku once this changes is removed. 

There may be a couple other benefits, but those two are the most significant and relevant.

kirkkwang:
Cool that sounds good too.  I'm going to have to bring in @jeremyf on this because I can go either way and I'll let the difference of opinions be put forth here so we have it tracked.  I'll put this PR back into draft for now.

bkiahstroud (commented):
For the record, I agree with @jeremyf's suggested rename of the ENV variable to `HYKU_BLOCK_VALKYRIE_REDIRECT`. That way we don't have to flip the polarity

config/routes.rb
```diff
@@ -5,6 +5,15 @@
 require 'sidekiq/web' if ENV.fetch('HYRAX_ACTIVE_JOB_QUEUE', 'sidekiq') == 'sidekiq'
 
 Rails.application.routes.draw do # rubocop:disable Metrics/BlockLength
+  Hyrax.config.curation_concerns.each do |resource|
+    next if resource.to_s.match?(/.+Resource\z/)
+
+    resource = resource.to_s.underscore
+
+    get "/concern/#{resource}s/:id/edit", to: redirect("/concern/#{resource}_resources/%{id}/edit")
```
jeremyf, line 13: It's possible that we don't need to use a redirect but could instead specify the controller/action pair.  Since this route is declared earlier than Hyrax, this route will take precedence.

Why avoid using a redirect?  It saves at least one HTTP request.

kirkkwang: cool i'm not sure what that means but i'll do some research

PR #2154, opened 2024-01-16
🎁 Add listener to remove PCDM Collection from Featured

jeremyf:
With this commit I'm introducing the HyraxListener class, registering it
as part of the pub/sub ecosystem.

Then wiring in the behavior that removes the collection from the
featured collection.  This echoes the callback behavior of the
Collection model.


app/models/collection.rb
```diff
@@ -11,6 +11,6 @@ class Collection < ActiveFedora::Base
   prepend OrderAlready.for(:creator)
 
   def remove_featured
```
ShanaLMoore, line 13: How does this method get called?

jeremyf: This is called via some callbacks defined inside the class but outside the scope of this diff.

PR #2157, opened 2024-01-26
Modify Collection & Admin Set Valkyrization

laritakr:
Addresses:
- https://github.com/scientist-softserv/hykuup_knapsack/issues/35
- https://github.com/scientist-softserv/hykuup_knapsack/issues/94

Moves away from directly using Hyrax::PcdmCollection and Hyrax::AdministrativeSet instead favoring inheriting from those (with CollectionResource and AdminSetResource). This allows for inheriting from Hyrax factories and extending them; thus bringing Hyku’s testing ecosystem closer to Hyrax.

Similar work may be desired for FileSets, but this would require some base Hyrax work first.

Note: There is javascript in app.js.erb that may affect the behavior of granting of permissions in the UI... this still needs to be investigated.


jeremyf (approved):

laritakr:
Replaced by https://github.com/samvera/hyku/pull/2164

PR #2159, opened 2024-01-30
🧹 Make the appropriate link generate

kirkkwang:
This commit will add a decorator to check the `human_readable_type` of the given object is a Valkyrie migration object (which by convention ends with 'resource') and adjust the generated link accordingly.  For example, a GenericWorkResource should generate links like `/concern/generic_work_resources/...` instead of
`/concern/generic_works/...`.

The models were updated to include the Hyrax::NestedWorks so the `Attach Child` button would populate the appropriate work types.

Lastly, the `Gemfile.lock` was updated to include the latest version of Hyrax `double_combo` branch.


app/models/concerns/hyrax/solr_document_behavior_decorator.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax 5.0.0rc2 to account for Valkyrie migration object that end in "Resource"
+
+module Hyrax
+  module SolrDocumentBehaviorDecorator
+    def hydra_model(classifier: nil)
```
jeremyf, line 7: I don't like relying on the human_readable_type; that's subject to translation whims and I foresee a very quick change from `Generic Work resource` to `Generic Work`.

Consider that we have [`valkyrie_bsi` attribute on the Solr document](https://github.com/samvera/hyrax/blob/f4b8c28768314e690f9cff454184600e0ca49d29/app/models/concerns/hyrax/solr_document_behavior.rb#L80-L84); can we use that?

PR #2161, opened 2024-02-02
🧹 Fix relationship links

kirkkwang:
This commit will make sure that the set child flag transaction fires when ingesting with Bulkrax.  Also fixed a weird bug where the extra space in the query was causing the query to fail.

TODO: update gemfile.lock once the following gets merged

- [ ] https://github.com/scientist-softserv/iiif_print/pull/328
- [ ] https://github.com/samvera/bulkrax/pull/908

Issue: 
- https://github.com/scientist-softserv/hykuup_knapsack/issues/143

### DIFF
--- a/app/indexers/concerns/hyku_indexing.rb
+++ b/app/indexers/concerns/hyku_indexing.rb
@@ -10,7 +10,7 @@ module HykuIndexing
     define_method method_name do |*args, **kwargs, &block|
       super(*args, **kwargs, &block).tap do |solr_doc|
         # rubocop:disable Style/ClassCheck
-        
+
         # specs refer to object as @object
         object ||= @object || resource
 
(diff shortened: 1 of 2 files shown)


app/indexers/concerns/hyku_indexing.rb
```diff
@@ -10,14 +10,15 @@ module HykuIndexing
...
-        
+
         # specs refer to object as @object
         object ||= @object || resource
 
         solr_doc['account_cname_tesim'] = Site.instance&.account&.cname
         solr_doc['bulkrax_identifier_tesim'] = object.bulkrax_identifier if object.respond_to?(:bulkrax_identifier)
         solr_doc['account_institution_name_ssim'] = Site.instance.institution_label
         solr_doc['valkyrie_bsi'] = object.kind_of?(Valkyrie::Resource)
+        solr_doc['member_ids_ssim'] = object.member_ids.map(&:id)
```
ShanaLMoore, line 21: Note to self: We may need a valkyrie check here? valkyrie member ids returns valkyrie objects with id keys, allowing us to map (:id). Active Fedora's member_ids are just ids.

ShanaLMoore (approved):

PR #2176, opened 2024-04-03
🐛 Fix PDF.js for Valkyrie

kirkkwang:
# Story

This commit will bring back the full functionality of PDF.js for Valkyrie Hyrax.  This includes:

- toggle the PDF.js viewer
- toggle downlad pdf button

Ref:
  - https://github.com/scientist-softserv/iiif_print/issues/342

# Screenshots

image

image

### DIFF
--- a/app/listeners/hyrax_listener.rb
+++ b/app/listeners/hyrax_listener.rb
@@ -55,8 +55,9 @@ class HyraxListener
   # def on_collection_membership_update
   # end
 
-  # def on_file_characterized
-  # end
+  def on_file_characterized(event)
+    pdf_viewer_and_download_button(event)
+  end
 
   # def on_file_downloaded
   # end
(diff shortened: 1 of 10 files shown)


app/listeners/hyrax_listener.rb
```diff
@@ -55,8 +55,9 @@ def on_collection_metadata_updated(event)
   # def on_collection_membership_update
   # end
 
-  # def on_file_characterized
-  # end
+  def on_file_characterized(event)
```
ShanaLMoore, line 58: Remind me, how does this get called again?

kirkkwang: Sure!  When the `'file.characterized'` gets published like in [here](https://github.com/samvera/hyrax/blob/double_combo/app/services/hyrax/characterization/valkyrie_characterization_service.rb#L19-L22), the listeners pick it up and then we can execute logic upon trigger, here's some comments that explain it better

https://github.com/samvera/hyrax/blob/double_combo/app/services/hyrax/listeners/file_listener.rb#L8-L13

jeremyf (approved):

PR #2191, opened 2024-04-22
Update Bulkrax for Embargo/Lease import support

ShanaLMoore:
Contribute bulkrax x embargo back to hyku

ref:
- https://github.com/samvera/bulkrax/commit/855ce39e805295377eda9a7c24645fbb0ee5617c

Issue:
- https://github.com/scientist-softserv/palni-palci/issues/1008

@samvera/hyku-code-reviewers

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -125,7 +125,7 @@ GIT
 
 GIT
   remote: https://github.com/samvera/bulkrax.git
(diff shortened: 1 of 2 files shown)


laritakr (approved):
Looks good. Hopefully the build problem can be resolved soon.

Is there a migration that should be run & committed as well? I don't see a schema change for this....

Issue #2194, opened 2024-04-23
PR Feedback => tickets 


ShanaLMoore:
ref: 
- https://github.com/samvera/hyku/pull/2178/files

- [ ] app/controllers/concerns/hyrax/collections_controller_behavior_decorator.rb

<img width="533" alt="image" src="https://github.com/samvera/hyku/assets/10081604/920d62c8-9cf1-45eb-a3a6-9544869ce3d1">

- [ ] app/controllers/hyrax/dashboard/profiles_controller_decorator.rb

<img width="533" alt="image" src="https://github.com/samvera/hyku/assets/10081604/d60e841a-ca2b-4d45-9b29-c9b1e324e7c8">

- [ ] app/models/concerns/bulkrax/has_matchers_decorator.rb

<img width="613" alt="image" src="https://github.com/samvera/hyku/assets/10081604/2be89d39-74c4-4de6-9281-bf15efe64c3d">

- [ ] app/models/file_download_stat_decorator.rb

<img width="534" alt="image" src="https://github.com/samvera/hyku/assets/10081604/4eec0f9b-65da-463c-94b6-5befba4afcc6">

- [ ] app/views/catalog/_index_header_list_default.html.erb

<img width="554" alt="image" src="https://github.com/samvera/hyku/assets/10081604/212a78ce-59a4-4855-bf18-5f114260347b">

- [ ] app/views/hyrax/base/_related_items.html.erb

<img width="549" alt="image" src="https://github.com/samvera/hyku/assets/10081604/d3d9ab10-60a0-4730-afad-6534d24de4cb">

- [ ] lib/tasks/sidekiq.rake

<img width="538" alt="image" src="https://github.com/samvera/hyku/assets/10081604/89972ee4-56ce-48bf-b950-0d942d2fb763">

- [ ] spec/controllers/hyrax/homepage_controller_spec.rb

<img width="539" alt="image" src="https://github.com/samvera/hyku/assets/10081604/edcdba8c-c006-47c5-82c7-df99a76a74a5">

- [ ] spec/features/create_work_spec.rb

<img width="547" alt="image" src="https://github.com/samvera/hyku/assets/10081604/42102d72-4338-4460-8cb5-660a6c5498f4">

ShanaLMoore:
# OTHER NOTICED ISSUES

- [ ] thumbnails missing from search index pg

https://github.com/scientist-softserv/palni-palci/blob/main/app/views/catalog/_index_header_list_default.html.erb

[...]

PR #2195, opened 2024-04-24
I2194 address pr feedback and then some

ShanaLMoore:
ref feedback: 
- https://github.com/samvera/hyku/issues/2194

I will also be correcting anything else wrong that I'm noticing, while spinning up the app. 

After this gets merged, we will cut a new release of hyku! 

## FIXES INCLUDE: 
- [PR feedback](https://github.com/samvera/hyku/issues/2194)
- defaults basic auth to samvera/hyku
- sets favicon and updates view to support it
- adds raise to has_matcher_decorator to alert user of missing geonames. This should be moved to bulkrax but we're leaving it here for now. 
- removes sidekiq.rake per Rob's request. It was intentionally removed from Hyku previously. 
- adds a Hyrax bug fix to location_indexer_decorator to pass a string and avoid based_near error when nil
- Rob contributed back redis cache store setting in production.rb config
- rubocop fix to recent_document partial. 
- spec updates


kirkkwang (dismissed):

laritakr (changes requested):
I'm requesting either a change or verification that the favicon changes don't break existing ability.

app/views/_head_tag_extras.html.erb
```diff
@@ -1,25 +1,25 @@
...
 
 <!-- generics -->
-<link rel="icon" href="<%= Site&.favicon&.url(:v32) %>" sizes="32x32">
-<link rel="icon" href="<%= Site&.favicon&.url(:v57) %>" sizes="57x57">
-<link rel="icon" href="<%= Site&.favicon&.url(:v76) %>" sizes="76x76">
-<link rel="icon" href="<%= Site&.favicon&.url(:v96) %>" sizes="96x96">
-<link rel="icon" href="<%= Site&.favicon&.url(:v128) %>" sizes="128x128">
-<link rel="icon" href="<%= Site&.favicon&.url(:v128) %>" sizes="192x192">
-<link rel="icon" href="<%= Site&.favicon&.url(:v228) %>" sizes="228x228">
+<link rel="icon" href="<%= favicon(:v32) %>" sizes="32x32">
```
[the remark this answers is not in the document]
ShanaLMoore: @laritakr I was still able to update the favicon from the admin dashboard: 

It used to be the samvera tree, but I updated it on the hyku-local tenant, seen below: 

image


image

ShanaLMoore:
> I'm requesting either a change or verification that the favicon changes don't break existing ability.

@laritakr I was still able to update the favicon from the admin dashboard: 

It used to be the samvera tree, but I updated it on the hyku-local tenant, seen below. It seems like this is to set the default mainly for the proprietor page, unless the tenant specifies otherwise. 

image


image

laritakr (dismissed):

app/indexers/concerns/hyrax/location_indexer_decorator.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# OVERRRIDE HYRAX v5.0.1 to turn loc into a string for the location lookup,
```
kirkkwang, line 3:
```suggestion
# OVERRIDE HYRAX v5.0.1 to turn loc into a string for the location lookup,
```

Being a little picky, but often I do search for the `OVERRIDE` comment so this might get missed

ShanaLMoore: no totally, you caught my typo! thanks!!

ShanaLMoore: @kirkkwang would you mind re approving?

kirkkwang (approved):

PR #2197, opened 2024-04-29
🎁 Add `ActiveJob::Scheduler`

kirkkwang:
The deploy was not working because it was missing the `repeat` method which is provided by the `ActiveJob::Scheduler` gem.


ShanaLMoore:
specs are passing locally 🤔

ShanaLMoore (approved):

PR #2200, opened 2024-05-01
🎁 1019 batch email notifications

ShanaLMoore:
# Story

Users can get the notifications in the form of an email digest set to daily, weekly, or monthly frequencies.

Refs
  - https://github.com/scientist-softserv/palni-palci/issues/1019

Users only get notifications that they would check in the admin dashboard (top right area).

Users can choose a daily, weekly, or monthly email digest of notifications sent to their registered email.

# Screenshots / Video

image

# Notes


db/schema.rb
```diff
@@ -914,6 +914,7 @@
     t.string "preferred_locale"
     t.string "provider"
     t.string "uid"
+    t.string "batch_email_frequency", default: "off"
```
ShanaLMoore, line 917: Request from @laritakr could we change "off" to be "never". Something about off being a reserved keywords for translations and her UI work.

ShanaLMoore: Since you are off today, she will likely make this change herself. Please confirm on your return.

laritakr:
Requires translations to run.

app/views/hyku_mailer/summary_email.html.erb
```diff
@@ -0,0 +1,42 @@
...
+    }
+    th {
+      background-color: #f2f2f2;
+    }
+  </style>
+</head>
+<body>
+  <h2>Hello, <%= @user.name %></h2>
+  <p>You have the following notifications:</p>
+  <p>Click <a href="<%= @url %>">here</a> to view them in Hyku.</p>
```
laritakr, line 26: I'm wondering if we should use the application &/or tenant name rather than "Hyku"? Not a big deal, but many of the users may not know what "Hyku" is. If the email is too generic, wouldn't they possibly just assume it's spam? Including the actual site name, a possible site logo, etc would seem more clear.

Additionally, I can see a future request might be to use i18n for the email text... just something to consider.

laritakr (approved):
Just a comment to consider, but definitely not a stopper at this point. Looks good to me.

ShanaLMoore:
We discussed we will replace HYKU with the account settings, in the email. Maybe institution name? 

Email language support will be out of scope for this round. Have the client create another ticket for this.

laritakr (approved):

PR #2205, opened 2024-05-06
🐛 Bring back footer for splash page

kirkkwang:
This commit will bring the footer back for the splash page as well as retain the footer display on the homepage.

ref:
  - https://github.com/scientist-softserv/hykuup_knapsack/issues/211

### Logged out
image

### Logged in
image

### In a tenant
image


ShanaLMoore:
For the first screenshot - I assume this is the view when the user is logged out? After you login does a link to the accounts appear in the top left corner?

kirkkwang:
Yes that is correct, it does appear.

ShanaLMoore (approved):

PR #2208, opened 2024-05-08
⚙️ Use valkyrie and show backtrace on staging

kirkkwang:
This commit will add the HYRAX_VALKYRIE env var so Hyku will now use Valkyrie.  It will also add a HYKU_SHOW_BACKTRACE env var so that the backtrace will be shown on the staging server.


ShanaLMoore (approved):

ShanaLMoore:
We can see the files again 🎉 
image

ShanaLMoore:
and backtrace works for staging as well 🎉 
image

PR #2187, opened 2024-04-17
I343 iiif print valkyrization

laritakr:
Refs 
- https://github.com/scientist-softserv/iiif_print/issues/343
- https://github.com/scientist-softserv/iiif_print/issues/312

Changes proposed in this pull request:
* updates schema.rb
* additional tenant_config options for iiif_print
* Attempt to get Dockerfile to appropriately set up tesseract

## Notes 
- in my test, the build did not fully complete. Additional work will be needed to the build.
- An update to the IiifPrint gem is still needed

PR #2209, opened 2024-05-09
copy of i343

ShanaLMoore:
Copy of PR to see if the pipeline will behave differently.

ref:
- https://github.com/samvera/hyku/pull/2187


ShanaLMoore:
did not resolve the failing specs in the pipeline. specs are passing locally 🤔 closing

PR #2216, opened 2024-05-17
:gift: allow knapsack to override resource metadata

ShanaLMoore:
In this commit, we wrap the inclusion of metadata in a conditional so that hyku knapsack can override it, if the decorators exists.


ShanaLMoore:
found an alternative approach that doesn't require us to decorate the models, indexers and forms

PR #2219, opened 2024-05-23
🎁 Add branding persistent volumes to deploys

ShanaLMoore:
Collection thumbnail would not display, even though it referenced the correct path. Perhaps we need to add a persistent volume. I will test this branch to verify that it works. 

## BEFORE

image

image

## NOTES

This has issue has been seen and solve in other projects. 

ref: 
- https://github.com/scientist-softserv/atla-hyku/pull/187

This commit is required to resolve the issue with uploading images from the Collections' branding tab. Because the symlink wasn't established, a user would not be able to see the images they uploaded even though the path correctly exists.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,6 @@
 version: 2.1
 orbs:
   samvera: samvera/circleci-orb@1
-  browser-tools: circleci/browser-tools@1.4.6
   ruby: circleci/ruby@2
   node: circleci/node@5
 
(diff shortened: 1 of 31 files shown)


ShanaLMoore:
The file displays now, but...
image

Thumbnail is still missing: 

image

ShanaLMoore:
The thumbnail path needs to be 
image

## TODO
remove the app/samvera/ from the beginning of it

PR #2227, opened 2024-05-30
⬆️ Update WillowSword

kirkkwang:
This update will bring in a more flexible way of creating works.  Prior to this commit we were only able to add dc terms which something like `keyword` was not apart of (it is from schema.org instead).  Also, we are able to set embargo and lease visibilities.


laritakr (approved):
Glad to have permissions working!

PR #2231, opened 2024-05-30
allow use of Google Analytics 4

orangewolf:
Updates Hyku config to allow GA4. Sets GA4 up so that events reported to GA4 will go both to the site wide `GOOGLE_ANALYTICS_ID` and to a tenant specific Analytics ID if one is set.


ShanaLMoore (approved):

ShanaLMoore:
Rob, is this ready to be merged?

orangewolf:
@ShanaLMoore yes, I think so

PR #2234, opened 2024-06-03
Update bulkrax

laritakr:
Update Bulkrax

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -125,7 +125,7 @@ GIT
 
 GIT
   remote: https://github.com/samvera/bulkrax.git
-  revision: bf7bc5ca422ca3d8961607d4eb558fce55517e90
+  revision: d8f9b85a9064b352a252c5deb590a430ca4a738e
   branch: main
   specs:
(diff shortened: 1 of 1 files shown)


config/initializers/bulkrax.rb
```diff
@@ -117,10 +117,5 @@
   end
   # rubocop:enable Metrics/BlockLength
 
-  # Sidebar for hyrax 3+ support
```
kirkkwang, line 120: I think we still need this

config/initializers/bulkrax.rb
```diff
@@ -117,10 +117,5 @@
...
-  if Object.const_defined?(:Hyrax) && ::Hyrax::DashboardController&.respond_to?(:sidebar_partials)
-    Hyrax::DashboardController.sidebar_partials[:repository_content] << "hyrax/dashboard/sidebar/bulkrax_sidebar_additions"
```
kirkkwang, line 122: I think we still need this

kirkkwang (approved):

PR #2236, opened 2024-06-04
:bug: [i51] Fix for collection sharing bug

ShanaLMoore:
Before this commit, adding sharing to a collection would result in an error. The form was sending the value of "false" instead of the selected access.

Issue: 
- https://github.com/scientist-softserv/palni_palci_knapsack/issues/51

## BEFORE

Failed silently and did not add selections to the form. This is because access params was set to false instead of the actual selected value. 

`#<ActionController::Parameters {"access_grants_attributes"=>#<ActionController::Parameters {"0"=>#<ActionController::Parameters {"access"=>"false", "agent_id"=>"registered", "agent_type"=>"group"} permitted: true>} permitted: true>} permitted: true>`

https://github.com/scientist-softserv/palni_palci_knapsack/assets/84697174/d6c4b44e-c6ae-435d-9090-aa2973361bc0

## AFTER

image

Screenshot 2024-06-04 at 15-39-31 Edit User Collection collection test __ Hyku


kirkkwang (dismissed):

ShanaLMoore:
Putting back in draft to confirm if the failures are legit.

laritakr (approved):

PR #2242, opened 2024-06-12
I205 collection thumbnail rework

ShanaLMoore:
Knapsacks' branding uploads are still not displaying the files because the directories aren't symlinked. We suspect there may be a layering issue when putting it in the Dockerfile. We are moving this bit of code to happen later in the process.


bin/web
```diff
@@ -2,4 +2,9 @@
 # frozen_string_literal: true
 `echo "$GOOGLE_OAUTH_PRIVATE_KEY_VALUE" | base64 -d > prod-cred.p12` unless ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'].to_s.strip.empty?
 
+# Ensure the directory exists and create the symbolic link
+RUN mkdir -p /app/samvera/hyrax-webapp/public && \
```
ShanaLMoore, line 6:
```suggestion
mkdir -p /app/samvera/hyrax-webapp/public && \
```

kirkkwang (dismissed):

kirkkwang (approved):

PR #2244, opened 2024-06-14
Branding fixes

laritakr:
# Story
Fixes https://github.com/scientist-softserv/hykuup_knapsack/issues/205

Alt text did not work correctly for collection thumbnails and banner.
Default collection image did not get used... default work image was used instead.

- [ ] Setting Branding
    - [ ] Images and text can be set and will appear on collection's branding tab
    - [ ] Images and text can be removed and disappear from collection's branding tab
    - [ ] Branding shows appropriately on collection's public show page
    - [ ] Branding reverts to standard when removed from collection's branding tab
- [ ] Setting Default thumbnails
    - [ ] Default thumbnails and alt text are saved and appear on Settings/Appearance default images tab
    - [ ] Default thumbnails and text can be removed and disappear from Settings/Appearance default images tab
    - [ ] Setting a default thumbnail or text reindexes all works &/or collections
    - [ ] Removing a collection's default thumbnail does not affect works
    - [ ] Removing a work's default thumbnail does not affect collections
- [ ] Thumbnails show appropriately on all views:
    - [ ] catalog search results view
    - [ ] dashboard index views
    - [ ] collection public show
    - [ ] collection dashboard show
    - [ ] homepage featured collections

# Notes

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller_decorator.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller_decorator.rb
@@ -5,6 +5,7 @@
 # - ensure user is allowed to change visibility
 # - add the ability to upload a collection thumbnail
 # - add altext to collection banner
+# @TODO clean up of unnecessary methods now that we are using Valkyrie transactions to set branding
 
 module Hyrax
   module Dashboard
@@ -45,7 +46,7 @@ module Hyrax
       def edit
         form
         # Gets original filename of an uploaded thumbnail. See #update
-        return unless ::SolrDocument.find(@collection.id).thumbnail_path.include?("uploaded_collection_thumbnails") && uploaded_thumbnail?
+        return unless ::SolrDocument.find(@collection.id).thumbnail_path&.include?("uploaded_collection_thumbnails") && uploaded_thumbnail?
         @thumbnail_filename = File.basename(uploaded_thumbnail_files.reject { |f| File.basename(f).include? @collection.id }.first)
       end
 
--- a/app/controllers/hyrax/dashboard/collections_controller_decorator.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller_decorator.rb
@@ -2,9 +2,9 @@
 
 # OVERRIDE Hyrax v5.0.0rc2
 # - Fix file upload in logo and banner
-# - Use work titles for collection thumbnail select & to add an option to reset to the default thumbnail
-
-# OVERRIDE Hyrax v5.0.0 to add the ability to upload a collection thumbnail
+# - ensure user is allowed to change visibility
+# - add the ability to upload a collection thumbnail
+# - add altext to collection banner
 
 module Hyrax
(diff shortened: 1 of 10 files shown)


app/controllers/hyrax/dashboard/collections_controller_decorator.rb
```diff
@@ -19,21 +20,14 @@ def show
         super
       end
 
-      # OVERRIDE Hyrax v5.0.0 to add the ability to upload a collection thumbnail - START
```
laritakr, line 22: I moved this down since it shouldn't be getting used any more and didn't need as prominent of location in the module.

app/controllers/hyrax/dashboard/collections_controller_decorator.rb
```diff
@@ -19,21 +20,14 @@ def show
...
-      end
-
       # rubocop:disable Metrics/AbcSize
       def update_valkyrie_collection
         return after_update_errors(form_err_msg(form)) unless form.validate(collection_params)
 
         result = transactions['change_set.update_collection']
                  .with_step_args(
                           'collection_resource.save_collection_banner' => { update_banner_file_ids: params["banner_files"],
-                                                                            banner_unchanged_indicator: params["banner_unchanged"] },
```
laritakr, line 36: banner_unchanged was unneeded, because we need the ability to save text even if the banner doesn't change.

app/views/hyrax/collections/show.html.erb
```diff
@@ -36,16 +39,15 @@
       <% unless @presenter.total_viewable_items.blank? %>
           <div class="hyc-bugs">
             <div class="hyc-item-count">
-              <b><%= @presenter.total_viewable_items %></b>
```
laritakr, line 39: This resulted in the count showing twice... the line below already includes it.

ShanaLMoore:
spec/tasks/rake_spec.rb:96 is failing for me locally. I'll see if I can sort it out.

laritakr:
> spec/tasks/rake_spec.rb:96 is failing for me locally. I'll see if I can sort it out.

Yeah my attempted fix didn't fix. But it is still a random failure the majority of the time as is... Wondering if we should just comment those out?

ShanaLMoore:
> > spec/tasks/rake_spec.rb:96 is failing for me locally. I'll see if I can sort it out.
> 
> Yeah my attempted fix didn't fix. But it is still a random failure the majority of the time as is... Wondering if we should just comment those out?

I reverted the attempted fix and it passed for me locally. I'll still merge it if it still fails on the CI. Perhaps we can comment them out but I'd want to dig into it a little more since I don't remember rake tasks specs failing before.

laritakr:
The rake task has been flaky for me with almost every PR (it's a failure on main as well). It passes for me locally as well, so I don't know how to fix it.

laritakr:
That large files job is also a flaky spec. :/

PR #2248, opened 2024-06-20
🐛 Wrap quotes around ENV var

ShanaLMoore:
Not doing so treats the ENV var as a class, and thus throws an error about that class not existing.

## BEFORE
The site was unable to render

image

## AFTER

image


ShanaLMoore:
todo: update pals' submodule once this gets merged

laritakr (approved):

PR #2250, opened 2024-06-20
Update gems

ShanaLMoore:
Pulls in fixes from Hyrax and IiifPrint to support display options within the form block of the metadata yaml files


laritakr:
I assume we also want to update iiif_print to hide the iiif_Print terms?

laritakr (approved):

PR #2252, opened 2024-06-21
🐛 Fix Add Child link

kirkkwang:
The links under the Add Child dropdown was not working because the HTML structure was not correct.  This commit will close the button tag and then open the ul tag which allows the links to be clickable.

Ref:
  - https://github.com/scientist-softserv/palni_palci_knapsack/issues/78

pals-38


ShanaLMoore:
Have you tried this in palni palci yet? I thought the issue was that CDL was missing from the drop down.

ShanaLMoore (approved):

ShanaLMoore:
> @ShanaLMoore Ah I uploaded the wrong screen cap, that was in Hyrax where the problem stemmed from. I've updated it now and it should have Work, Image, ETD, and OER. I will investigate the other half of the problem of why PALs is only getting two work types when this is updated in PALs. The focus of this PR is to make sure the Add Child link is navigating to the correct place.

That makes sense. This is a hyku PR after all. Thanks!

PR #2255, opened 2024-06-24
:broom: Update iiif print

ShanaLMoore:
pulls in fix for undefined method
ref: 
- https://github.com/scientist-softserv/iiif_print/pull/360


ShanaLMoore:
closed due to accidental commits of unrelated files. I'll fix and reopen then.

PR #2267, opened 2024-07-14
Ci build docker images for macos 14

bkiahstroud:


bkiahstroud:
This was a brief test to see if this would be an easy win, but since `docker` isn't installed in the MacOS runner, it likely isn't. Plus [it seems like others have had difficulties](https://github.com/jupyter/docker-stacks/issues/2088#issuecomment-1925317748) achieving the same. 

I think the next best option is using the Arm-based Linux runners when they become available (theoretically later this year): 
- https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/

PR #2279, opened 2024-08-12
✅ renames namespace due to naming conflicts

ShanaLMoore:
Adventist override was not taking precedence because Hyku had an override with the same name. This PR changes the name space to specify Hyku, so that when Adventist wants to override Hyrax it truly overrides Hyrax.

### DIFF
--- a/spec/presenters/hyrax/collection_presenter_decorator_spec.rb
+++ b/spec/presenters/hyrax/collection_presenter_decorator_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Hyrax::CollectionPresenter, type: :decorator do
         def collection_type_badge
           "<span>"
         end
-        prepend Hyrax::CollectionPresenterDecorator
+        prepend Hyku::CollectionPresenterDecorator
       end
     end
     let(:presenter) { base_class.new }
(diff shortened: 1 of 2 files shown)


spec/presenters/hyrax/collection_presenter_decorator_spec.rb
```diff
@@ -21,7 +21,7 @@
         def collection_type_badge
           "<span>"
         end
-        prepend Hyrax::CollectionPresenterDecorator
+        prepend Hyku::CollectionPresenterDecorator
```
kirkkwang, line 24: Just being a stickler but should the spec's path also be moved?

`spec/presenters/hyku/collection_presenter_decorator_spec.rb`

ShanaLMoore: I debated that but the spec is describing Hyrax::CollectionPresenter so I decided against it. 🤔

ShanaLMoore: I suppose an alternative approach is the change adventist's class name instead.

kirkkwang: I have no strong opinions!  The only reason why I was thinking that was because my switch to spec extension would probably not be able to find this because it sticks to convention so it would expect the same path pattern.

kirkkwang (approved):

PR #2284, opened 2024-08-14
Themes logic refactor

kirkkwang:
# Summary

## ♻️ Add `Hyku::HomePageThemesBehavior`

2ba8794749b0dff8fb3ad04104ab66d2f07e02b6

This commit will move the #inject_theme_views method and the
around_action that calls it into a concern.

## 🐛 Use unshift instead of push

94d495664bffa79804aa4c6c75c59671829454f5

In the HykuKnapsack, we want the themes view path to be prioritized over
the Hyku application's themes view path.


laritakr (approved):
great work!

PR #2288, opened 2024-08-15
simplify per tenant recurring jobs, make sure they stay running by checking them every night

orangewolf:


laritakr (dismissed):
Code looks good, pending the ability to actually build & run specs.

PR #2295, opened 2024-08-19
Subdomains to exclude

orangewolf:
A refiling of https://github.com/samvera/hyku/pull/1654


cjcolvar (approved):
Still looks good to me!

PR #2302, opened 2024-08-22
Batch Email Notifications

laritakr:
Refs https://github.com/scientist-softserv/palni_palci_knapsack/issues/121

`last_emailed_at` was previously added to the User table. However User table is not tenant-specific, and this date needed to be unique to each tenant. This moves the `last_emailed_at` date to its own table and modifies the use to match the new location.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -407,12 +407,6 @@ ActiveRecord::Schema.define(version: 2024_08_20_200440) do
     t.datetime "updated_at", null: false
   end
 
-  create_table "hyrax_flexible_schemas", force: :cascade do |t|
-    t.text "profile"
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-  end
-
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -407,6 +407,12 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "hyrax_flexible_schemas", force: :cascade do |t|
```
laritakr, line 410: I'm not sure why this came in... there must have been a prior migration that hadn't been run and saved?

sjproctor (approved):
Looks good to me!

db/schema.rb
orangewolf: It needs to be manually removed or it will break future migrations on new apps. Just deleting these lines will do the trick.

orangewolf (changes requested):
This is good to go other than the db/schema.rb change.

kirkkwang (approved):

PR #2311, opened 2024-09-03
:broom: [i1020] - ga4 follow up

ShanaLMoore:
# Summary

This contains the following work needed to support google analytics: 

- [x] update Hyrax to pull in typo changes that prevented download tracking
  - [x] https://github.com/samvera/hyrax/pull/6897
- [x] add fallback for when application name is blank
- [x] don't send email if 0 new counts 
- [x] all download buttons needs 'file_download' as id for tracking 
  - missing from: 

Image

Issue: 
- https://github.com/scientist-softserv/palni-palci/issues/1020

# TODO: 

- [ ] update palni palci knapsack's submodule after this gets merged

## change download-pdf-button => file_download

54168b7ccf1f7e849a1a1194f2a881af2ae4b4f5

ID must be changed so to be consistent with google analytics tracking.

Issue:
- https://github.com/scientist-softserv/palni-palci/issues/1020

## Add fallback for when application_name is nil

fd8bdb741643826e6f80ab84c455fdb4d24e6349

This should avoid sending emails with blank names in the title.

## Skip if new_work_views or new_file_downloads is 0

608faf9b56a21c616a09ecc6eaadbd06d7e30d54

## update hyrax

7c392159ae2be1a3c5255ca08551cf64662b9deb

pulls in ga4 typo fix
- https://github.com/samvera/hyrax/pull/6897

### DIFF
--- a/app/jobs/depositor_email_notification_job.rb
+++ b/app/jobs/depositor_email_notification_job.rb
@@ -8,7 +8,6 @@ class DepositorEmailNotificationJob < ApplicationJob
       statistics = user.statistics_for
       next if statistics.nil?
 
-      # Skip if new_work_views and new_file_downloads is 0
       next if statistics[:new_work_views].zero? && statistics[:new_file_downloads].zero?
 
       HykuMailer.depositor_email(user, statistics, current_account).deliver_now
--- a/app/jobs/depositor_email_notification_job.rb
+++ b/app/jobs/depositor_email_notification_job.rb
@@ -8,6 +8,9 @@ class DepositorEmailNotificationJob < ApplicationJob
       statistics = user.statistics_for
       next if statistics.nil?
 
+      # Skip if new_work_views or new_file_downloads is 0
+      next if statistics[:new_work_views].zero? || statistics[:new_file_downloads].zero?
+
       HykuMailer.depositor_email(user, statistics, current_account).deliver_now
     end
     DepositorEmailNotificationJob.set(wait_until: (Time.zone.now + 1.month).beginning_of_month).perform_later
--- a/app/jobs/depositor_email_notification_job.rb
+++ b/app/jobs/depositor_email_notification_job.rb
@@ -8,8 +8,8 @@ class DepositorEmailNotificationJob < ApplicationJob
       statistics = user.statistics_for
       next if statistics.nil?
 
-      # Skip if new_work_views or new_file_downloads is 0
-      next if statistics[:new_work_views].zero? || statistics[:new_file_downloads].zero?
+      # Skip if new_work_views and new_file_downloads is 0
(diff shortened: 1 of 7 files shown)


app/views/hyku_mailer/depositor_email.html.erb
```diff
@@ -1,34 +1,34 @@
...
+        color: white;
+        text-decoration: none;
+        border-radius: 5px;
+      }
+    </style>
+  </head>
+  <body>
+    <div class="email-container">
+      <p>Dear Author,</p>
+      <p>You had <%= pluralize(@statistics[:new_file_downloads], 'new download') %> in <%= Date.today.prev_month.strftime("%B %Y") %> across your <%= pluralize(@statistics[:new_work_views], 'new work') %> in <%= @application_name %>. Your current readership:</p>
```
ShanaLMoore, line 23: all that I changed here was the addition of the word "new". I'm not sure why it looks like there are so many changes.

ShanaLMoore: and actually, this message doesn't really make sense. 

image

doesn't this imply there's been a new file download from one of the 6 new works, which i don't think is how these measurements work. instead of 6 new works should it be the total work count? or should language change to say there's been 6 new work views?

app/jobs/depositor_email_notification_job.rb
```diff
@@ -8,6 +8,9 @@ def perform
       statistics = user.statistics_for
       next if statistics.nil?
 
+      # Skip if new_work_views or new_file_downloads is 0
+      next if statistics[:new_work_views].zero? || statistics[:new_file_downloads].zero?
```
laritakr, line 12: I think this should be an and rather than an or... we'd want to send as long as they have anything non-zero, right?

app/mailers/hyku_mailer.rb
```diff
@@ -11,7 +11,7 @@ def summary_email(user, messages, account)
     @messages = messages || []
     @account = account
     @url = notifications_url_for(@account)
-    @application_name = account.sites.application_name
+    @application_name = account.sites.application_name || account.cname
```
laritakr, line 14: Not a blocker but should we make cname human readable so it looks better?

app/views/hyku_mailer/depositor_email.html.erb
laritakr: I would think it should say something like "You had 1 new download and 6 new works". Assuming that's what was tracked.

laritakr (changes requested):

app/views/hyku_mailer/depositor_email.html.erb
ShanaLMoore: changed to 
image

laritakr (approved):

PR #2313, opened 2024-09-05
Adjust themes

kirkkwang:
# Story
## 💄 Adjust featured collection section

74e4e0a4f63b9219179c73e28185a967ea8b4ec5

This commit will adjust the featured colleciton section so it wraps and
doesn't clip anything else.

## 💄 Adjust Recently Uploaded section

1af17b4236bf466a097a01f810daa1a036684d67

In the Institutional Repository and Neutral Repository themes, the
Recently Uploaded section was only showing one column, this commit makes
it show two.

# Screenshots / Video

image

image

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -695,3 +695,10 @@ tr[data-feature="use-iiif-print"] {
     transition: 0.1s;
   }
 }
+
+// Featured collection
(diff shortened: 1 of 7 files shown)


laritakr (approved):
Code looks good. 

I see there's a spec failing but i've seen that to be a flaky spec PLUS that feature is actually broken and there's a ticket for it.

PR #2317, opened 2024-09-09
flesh out ability to find jobs in various background job tools

orangewolf:


ShanaLMoore (approved):
we haven't tested it but it looks good code wise

PR #2334, opened 2024-09-19
🧹 Updates version of order-already gem

sjproctor:
Ref:
- https://github.com/scientist-softserv/adventist_knapsack/issues/771

This commit updates the version of the order-already gem to 0.3.2. This version of the gem handles blank values preventing blank fields from being assigned `[""]` instead of `[]`.

@samvera/hyku-code-reviewers


orangewolf (changes requested):
this should have a lockfile update as well

ShanaLMoore (changes requested):
Please commit the Gemfile.lock as well. Without it the gem version won't actually be changed.

Sorry, disregard. I didn't noticed Rob already reviewed and commented!!

orangewolf (approved):

PR #2338, opened 2024-09-19
Support s3 file storage

orangewolf:
Need to add the following env vars to make this work:

```
REPOSITORY_S3_STORAGE=true
REPOSITORY_S3_BUCKET=
REPOSITORY_S3_REGION=us-west-2
REPOSITORY_S3_ACCESS_KEY=
REPOSITORY_S3_SECRET_KEY=
```

@samvera/hyrax-repo-managers


gleithner:
@orangewolf I'm currently implementing similar code except we use the prefix option and no access key/secret. Ideally I'd like to be able to use this code by setting shrine_s3_options in a knapsack and have that be respected in wings.rb when it configures the S3 storage adapter.

PR #2341, opened 2024-09-20
💄 Removes the multiple indicator from Solr suffix

sjproctor:
This commit improves communication of intent by removing the multiple indicator from the hyku_indexing module.

Ref:
- https://github.com/scientist-softserv/adventist_knapsack/issues/185

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -12,11 +12,11 @@ class CatalogController < ApplicationController
   before_action :enforce_show_permissions, only: :show
 
   def self.created_field
-    'date_created_ssim'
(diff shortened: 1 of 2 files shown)


ShanaLMoore:
I'm hving a change of heart about this. Won't it be a problem if the data in prod is multiple, but when we convert it to valkyrie we're asking it to be single? will we lose data that way? cc @sjproctor

I think we need to match how the index properties are defined in prod.

ShanaLMoore:
Please work with @kirkkwang  to resolve [this issue](https://github.com/scientist-softserv/adventist_knapsack/issues/822)

kirkkwang (approved):

PR #2344, opened 2024-09-20
💄 Match catalog search results match to Hyrax

kirkkwang:
# Summary

This commit will along the look of the catalog search results page to look closer to the way Hyrax looks.  Also, changing the #markup method to render #text and not #to_html because the #to_html method wraps a <p> tag around the text and threw off the spacing.

# Screenshots

## From PG Nurax
image

## Before
image

## After
image


ShanaLMoore (approved):

sjproctor (dismissed):

config/features.rb
```diff
@@ -43,6 +43,6 @@
           description: "Show General Login Link at Top Right of Page."
 
   feature :treat_some_user_inputs_as_markdown,
-          default: true,
+          default: false,
```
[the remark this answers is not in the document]
kirkkwang: this feature was brought over from PALs specifically, i don't think it should be a default for ALL hyku

spec/helpers/application_helper_spec.rb
```diff
@@ -6,6 +6,8 @@
     let(:bold) { '*bold*' }
 
     it 'renders markdown into html' do
+      allow(Flipflop).to receive(:treat_some_user_inputs_as_markdown?).and_return(true)
```
ShanaLMoore, line 9: Could you add a context for when treat_some_user_inputs_as_markdown? is set to false?

kirkkwang: Good idea, I can do that!

ShanaLMoore (approved):

PR #2362, opened 2024-10-21
Fix snippets

ShanaLMoore:
Issue: 
- https://github.com/scientist-softserv/adventist_knapsack/issues/769

## TODO: 

- [x] uv search params should not search if no highlights exists ✅ 
- [x] #render_ocr_snippets should check for highlights and not snippets.blank? ✅ 
- [x] When there are no snippets the "keyword matches" label still appears. ✅  (M3) **[SHANA]**

<details> 

### when there's a match

image

### when there's no matches

image

image

</details> 

- [ ] When there is a match, it should highlight the first one. ❌ (prod issue - M3)
- [ ] When there is a match, it also finds more matches than it should. ❌ (prod issue - M3)
- [x] Thumbnail link does not perform a UV search ✅  (M1) **[LARITA]**
- [ ] Update Adventist knapsack and make any necessary additional changes (M1)

### DIFF
--- a/app/views/catalog/_index_header_list_default.html.erb
+++ b/app/views/catalog/_index_header_list_default.html.erb
@@ -1,7 +1,7 @@
 <%# OVERRIDE Hyrax v5.0.1 to collection badge and markdown of links %>
 <div class="search-results-title-row col-12 d-flex flex-row align-items-center pb-2">
   <h4 class="search-result-title">
-    <%= link_to document.title_or_label, generate_work_url(document, request, params) %>
+    <%= link_to markdown(document.title_or_label), generate_work_url(document, request, params) %>
   </h4>
 
   <% if document.hydra_model == Hyrax.config.collection_class || document.hydra_model < Hyrax.config.collection_class %>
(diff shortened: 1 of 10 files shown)


app/views/catalog/_index_header_list_default.html.erb
```diff
@@ -1,6 +1,9 @@
 <%# OVERRIDE Hyrax v5.0.1 to collection badge and markdown of links %>
 <div class="search-results-title-row col-12 d-flex flex-row align-items-center pb-2">
-  <h4 class="search-result-title"><%= link_to markdown(document.title_or_label), generate_work_url(document, request) %></h4>
+  <h4 class="search-result-title">
+    <%= link_to document.title_or_label, generate_work_url(document, request, params) %>
```
orangewolf, line 4: Isn’t markdown an added feature? Why are we removing it?

orangewolf (commented):
Just one burning question. The rest looks good.

app/views/catalog/_index_header_list_default.html.erb
laritakr: I think this was an accidental change between several revisions of this line. It should definitely go back. Good catch.

orangewolf (approved):

PR #2363, opened 2024-10-25
make initialization of wings simpler and make it easier to override

orangewolf:
By setting `WINGS_CONCERNS` and using it we can make it simpler to set Wings and we can then use that set up to look them back up in reverse for the valkyrie adapter.

### DIFF
--- a/config/initializers/wings.rb
+++ b/config/initializers/wings.rb
@@ -8,20 +8,15 @@ end
 
 # rubocop:disable Metrics/BlockLength
 Rails.application.config.after_initialize do
-  [
-    GenericWork,
-    Image,
-    Etd,
-    Oer
-  ].each do |klass|
+  # Add all concerns that are migrating from ActiveFedora here
+  WINGS_CONCERNS ||= [AdminSet, Collection, Etd, GenericWork, Image, Oer].freeze
+
+  WINGS_CONCERNS.each do |klass|
     Wings::ModelRegistry.register("#{klass}Resource".constantize, klass)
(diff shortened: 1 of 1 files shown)


config/initializers/wings.rb
```diff
@@ -110,34 +105,8 @@
...
-      Hyrax::AccessControl
-    elsif 'FileSet' == klass_name
-      Hyrax::FileSet
-    elsif 'Hydra::AccessControls::Embargo' == klass_name
-      Hyrax::Embargo
-    elsif 'Hydra::AccessControls::Lease' == klass_name
-      Hyrax::Lease
-    elsif 'Hydra::PCDM::File' == klass_name
-      Hyrax::FileMetadata
-    else
```
laritakr, line 138: I'm a bit concerned with the removal of the above mappings. Where does Hydra::AccessControls::Embargo get mapped to Hyrax::Embargo, etc? All of these were added due to prior errors, and I want to be sure we aren't breaking prior fixes.

orangewolf: They are initialized in Hyrax wings adapter. If you open the Hyku console and do Wings::ModelRegistry.reverse_lookup(Hydra::AccessControls::Embargo) you will see the right result

orangewolf: Or at least I did when I tried them all one by one

laritakr (approved):

config/initializers/wings.rb
laritakr: I tried a few and they worked so I approved.

PR #2380, opened 2024-11-12
New account setting to override Bulkrax's default split pattern

bkiahstroud:
Ref:
- https://github.com/samvera/bulkrax/pull/984

# Summary

Add a new Account setting that will override Bulkrax's default multi-value split pattern. This effectively allows each tenant to have its own default split pattern

This change relies on this Bulkrax commit to work:
- https://github.com/samvera/bulkrax/pull/984/commits/9268bbb

image

# Considerations

The default split pattern is only used when a field mapping specifies `split: true`:

```ruby
config.field_mappings['Bulkrax::CsvParser'] = {
  'contributor' => { from: %w[contributor], split: true },
```

If a field declares a specific split pattern, **it will be used for all tenants**:

```ruby
config.field_mappings['Bulkrax::CsvParser'] = {
  'contributor' => { from: %w[contributor], split: /\s*[|]\s*/ },
```

I think this is okay; in my experience, most installations use a single pattern for the vast majority of their fields. If I'm reading [Bulkrax's `#process_parse` method](https://github.com/samvera/bulkrax/blob/77528bf27940441206beae95fc6902168707115f/app/matchers/bulkrax/application_matcher.rb#L43) correctly, that could be used in conjunction with `split: false` to handle one-off fields with unique split patterns:

```ruby
config.field_mappings['Bulkrax::CsvParser'] = {
  'contributor' => { from: %w[contributor], split: false },

---

def parse_contributor(src)
  src.split(/custom pattern/)
end
```

That being said, if you can think of a better way to accomplish this, I'm open to suggestions 

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -23,6 +23,7 @@ module AccountSettings
     setting :allow_downloads, type: 'boolean', default: true
     setting :allow_signup, type: 'boolean', default: true
     setting :analytics_provider, type: 'string'
+    setting :bulkrax_split_pattern, type: 'string', default: Bulkrax.multi_value_element_split_on.source
     setting :bulkrax_validations, type: 'boolean', disabled: true
     setting :cache_api, type: 'boolean', default: false
     setting :contact_email, type: 'string', default: 'change-me-in-settings@example.com'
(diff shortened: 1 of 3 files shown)


orangewolf (changes requested):
most of the time I feel like the default split pattern isn't as useful. its better to list the split pattern per field. And if we do that, then this wont help much. lets touch base tomorrow and see if this really solves the problem its pointed at

bkiahstroud:
This PR has been superseded by:
- #2384

PR #2381, opened 2024-11-13
[i886] - update logic to fix collection thumbnail bug

ShanaLMoore:
The default work thumbnail was displaying even though the default collection thumbnail was set.

Issue:
- https://github.com/scientist-softserv/adventist_knapsack/issues/886

# When No Collection Branding is Set

## When Default Collection Thumbnail is Set
- [ ] Collections should render the default thumbnail

Screenshot 2024-11-13 at 13-49-38 Show Appearance __ Hyku

Screenshot 2024-11-13 at 13-50-18 Collections

## When Default Collection Thumbnail is NOT set
- [ ] Collections should render Hyrax's default thumbnail

Screenshot 2024-11-13 at 13-50-54 Show Appearance __ Hyku

Screenshot 2024-11-13 at 13-51-05 Collections

# When Collection Branding is Set

## When Default Collection Thumbnail is Set
- [ ] Collections should render the Branding thumbnail

Screenshot 2024-11-13 at 13-50-37 Edit User Collection es __ Hyku

Screenshot 2024-11-13 at 15-40-53 Collections

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,7 +50,7 @@ Rails.application.configure do
   config.force_ssl = true
   config.ssl_options = {
     redirect: {
-      exclude: -> request { request.path == '/healthz' }
+      exclude: ->(request) { request.path == '/healthz' }
     }
   }
   # Use the lowest log level to ensure availability of diagnostic information
--- a/app/services/hyrax/thumbnail_path_service_decorator.rb
+++ b/app/services/hyrax/thumbnail_path_service_decorator.rb
@@ -10,10 +10,14 @@ module Hyrax
       collection_thumbnail = CollectionBrandingInfo.where(collection_id: object.id.to_s, role: "thumbnail").first
       return collection_thumbnail.local_path.gsub(Hyrax.config.branding_path.to_s, '/branding') if collection_thumbnail
 
-      return default_image if object.try(:thumbnail_id).blank?
+      if object.try(:thumbnail_id).blank?
+        return default_collection_image if object.try(:collection?)
+        return default_image
+      end
 
       thumb = fetch_thumbnail(object)
       return default_collection_image unless thumb
+
       return call(thumb) unless thumb.file_set?
       if audio?(thumb)
         audio_image
--- a/spec/services/hyrax/thumbnail_path_service_spec.rb
+++ b/spec/services/hyrax/thumbnail_path_service_spec.rb
@@ -18,4 +18,26 @@ RSpec.describe Hyrax::ThumbnailPathService, type: :decorator do
       end
     end
   end
+
+  describe '.default_collection_image' do
+    context 'when the site has a default collection image' do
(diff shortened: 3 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -18,13 +18,15 @@ def group_navigation_presenter
   #  - fallback to Hyrax's default image
   def collection_thumbnail(document, _image_options = {}, url_options = {})
     view_class = url_options[:class]
-    # The correct thumbnail SHOULD be indexed on the object
-    return image_tag(document['thumbnail_path_ss'], class: view_class, alt: alttext_for(document)) if document['thumbnail_path_ss'].present?
+    # Use collection branding thumbnail if it exists
+    if document['thumbnail_path_ss'].present? && !document['thumbnail_path_ss'].include?('default_work_images')
```
laritakr, line 22: This indicates that the wrong thumbnail is indexed... should we do something about this situation? I don't know the answer, but it seems like we'd want to remedy it somehow...

config/environments/production.rb
```diff
@@ -50,7 +50,7 @@
   config.force_ssl = true
   config.ssl_options = {
     redirect: {
-      exclude: -> request { request.path == '/healthz' }
+      exclude: ->(request) { request.path == '/healthz' }
```
ShanaLMoore, line 53: rubocop fix

app/helpers/application_helper.rb
ShanaLMoore: @laritakr yes, that appears to be what's happening. it gets the work default image because of[ this line](https://github.com/samvera/hyku/blob/main/app/services/hyrax/thumbnail_path_service_decorator.rb#L13). 

And you're right, instead of a band aid let me try to prevent it from doing that for collections.

laritakr (dismissed):

app/helpers/application_helper.rb
ShanaLMoore: Actually, I think all i need to do is check if it's a collection and called #default_collection_image instead 🤔

ShanaLMoore:
The failing specs gets resolved by updating Hyrax. ref: https://github.com/samvera/hyku/pull/2382

laritakr (approved):

ShanaLMoore:
This was tested in adventist knapsack as well as hyku demo, and it appears to be working. 

As an aside, locally I was getting this error when trying to set a thumbnail in Adventist knapsack, but the same error happened when using main (of both adventist knapsack and hyku). This error does not occur in staging or prod so I'm not sure what this is about. 

To reproduce, spin up main Adventist Knapsack. In the submodule check out this branch. Spin the project up, create a collection. Click the branding tab. Upload and save a thumbnail file. 🤔 

image

ShanaLMoore:
the spec and lint failure are coming from main so I'm going to go ahead and merge this

PR #2384, opened 2024-11-20
Add per-tenant Bulkrax field mappings

bkiahstroud:
## Summary

In multi-tenant Hyku applications, different tenants can have different requirements for field mappings. This is usually fine for the `:from` setting; if one tenant has a column named `title_alpha` and another `title_beta`, both can be added to the `title`'s field mapping `:from` setting. 

However, conflicts arise when tenants have different requirements for other settings, such as `:split` or deciding which field should be considered the `source_identifier`. In those cases, having a single, static field mapping config doesn't work.

This PR introduces a new `Account` settings as well as an override to Bulkrax to allow for configuring field mappings completely independently on a per-tenant basis. 

## Screenshots

<details><summary>New account default settings</summary>

image

</details>

<details><summary>Override possible by admins within tenant</summary>

image

</details>

<details><summary>JSON validation</summary>

image

</details>

<details><summary>Importer uses tenant field mappings</summary>

image

</details>

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -164,6 +164,8 @@ module AccountSettings
 
   def validate_json
     json_editor_settings.each do |key|
+      next unless settings[key].present?
+
       begin
         JSON.parse(settings[key])
       rescue JSON::ParserError => e
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -8,7 +8,7 @@ module AccountSettings
   extend ActiveSupport::Concern
   # rubocop:disable Metrics/BlockLength
   included do
-    cattr_accessor :array_settings, :boolean_settings, :hash_settings, :string_settings, :private_settings do
+    cattr_accessor :array_settings, :boolean_settings, :hash_settings, :json_editor_settings, :string_settings, :private_settings do
       []
     end
     cattr_accessor :all_settings do
@@ -23,6 +23,7 @@ module AccountSettings
     setting :allow_downloads, type: 'boolean', default: true
     setting :allow_signup, type: 'boolean', default: true
     setting :analytics_provider, type: 'string'
+    setting :bulkrax_field_mappings, type: 'json_editor', default: Bulkrax.field_mappings.to_json
     setting :bulkrax_validations, type: 'boolean', disabled: true
     setting :cache_api, type: 'boolean', default: false
     setting :contact_email, type: 'string', default: 'change-me-in-settings@example.com'
@@ -58,7 +59,7 @@ module AccountSettings
     validates :contact_email, :oai_admin_email,
               format: { with: URI::MailTo::EMAIL_REGEXP },
               allow_blank: true
-    validate :validate_email_format, :validate_contact_emails
+    validate :validate_email_format, :validate_contact_emails, :validate_json
     validates :google_analytics_id,
               format: { with: /((UA|YT|MO)-\d+-\d+|G-[A-Z0-9]{10})/i },
               allow_blank: true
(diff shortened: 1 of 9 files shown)


app/views/proprietor/accounts/edit.html.erb
```diff
@@ -34,7 +34,7 @@
           </div>
 
           <h3><%= t(".account_settings") %></h3>
-          <% current_account.live_settings.each do |key, value| %>
+          <% @account.live_settings.each do |key, value| %>
```
K8Sewell, line 37: Should this change be parked until question in FIXME in app/views/shared/_settings.html.erb is resolved?  I am not sure if the tests will continue to pass if this is changed back to `current_account` so if that is the case then please ignore this comment.

app/models/concerns/account_settings.rb
```diff
@@ -136,6 +137,12 @@ def set_type(value, to_type)
       value.is_a?(String) ? JSON.parse(value) : value
     when 'string'
       value.to_s
+    when 'json_editor'
+      begin
+        JSON.pretty_generate(JSON.parse(value))
+      rescue JSON::ParserError
+        value
```
K8Sewell, line 144: I am relatively sure `pretty_generate` or `parse` do not mutate the value so if that is true may I suggest setting it to a variable and returning that?  I could be misunderstanding the `begin` and `rescue` block so if that's true then feel free to ignore this.

Should the json be cleaned of any `"_proto_"` references before being returned?  Not sure if that is overkill for here or if that is done elsewhere.  Found some JS that does this.

```
const cleaner = (key, value) => key === "__proto__" ? undefined : value;

JSON.parse(json, cleaner)

```

app/views/proprietor/accounts/edit.html.erb
bkiahstroud: This change is okay to stay. `current_account` was incorrect here too and changing it to `@account` doesn't break anything. 

`current_account` was essentially returning `Account.new`, while `@account` is the actual Account the form is editing

app/models/concerns/account_settings.rb
bkiahstroud: You're correct that neither `parse` nor `pretty_generate` mutate the value. If you call `JSON.parse` on a string that is not valid JSON, it throws a `JSON::ParserError`. Since we're dealing with a value that comes from user input, we can't trust it to always be valid JSON. Because of that, we plan for a `JSON::ParserError` to be thrown and, when it is, we `rescue` and simply give the user back what they entered so they can correct it. 

[This recording](https://share.zight.com/jku090b2) shows what I'm trying to describe. This method simply returns `value` if it's invalid JSON, which is how it shows up in the text box (otherwise the text box would be empty if we returned nothing, forcing the user to start over from scratch). `#validate_json` is what handles adding the error message if `value` is not valid JSON. 

> Should the json be cleaned of any `"_proto_"` references before being returned? ...

I don't think so. I'm honestly not familiar with `_proto_` in the context of JSON. Since this JSON specifically is meant to represent Bulkrax field mappings (which don't use `_proto_`), I don't think we need to plan for that possibility

app/views/proprietor/accounts/edit.html.erb
K8Sewell: Well, then that should absolutely stay.  Thanks for the clarification!

app/models/concerns/account_settings.rb
K8Sewell: Ahh, that makes much more sense.  I had a feeling I was missing something about that rescue block.  Appreciate the explanation.

For the `"_proto_"` I was just doing some JSON sanitizing best practices and stumbled upon a [stackoverflow article](https://stackoverflow.com/questions/63926663/how-should-untrusted-json-be-sanitized-before-using-json-parse) about preventing a regular user from using that syntax to make themselves an admin like the code snippet below.  However, I absolutely understand if that is not a rational concern given the way that JSON is used.

```
json = '{"user":{"__proto__":{"admin": true}}}';
```

K8Sewell (approved):

PR #2391, opened 2024-11-26
Prevent field mappings set value from bleeding into other specs

bkiahstroud:
Since `Hyku.default_bulkrax_field_mappings=` sets an instance variable, it can persist in scope for other specs, causing unexpected / flaky failures


laritakr (approved):
Thanks for finding the solution so quickly! :)

PR #2393, opened 2024-11-28
Fix incorrectly colored save button

sephirothkod:
Simpleform injects css classes into its button tags and those classes conflict with what is used in Hyku. The forms on the Pages page were using the button tag instead of the submit tag and causing the classes to be duplicated and conflict with each other. By using the submit tag it doesn't inject classes and only applies the classes specified in the tag.

Button color for "Save changes" button was white instead of blue like the rest of Hyku. This fixes that issue as well as issues in the future stemming from this "class injection".

@samvera/hyku-code-reviewers

### DIFF
--- a/app/views/hyrax/pages/_form.html.erb
+++ b/app/views/hyrax/pages/_form.html.erb
@@ -1,4 +1,4 @@
-<%# Copied from Hyrax v5.0.1 to correct button coloring since simpleform overrides hyku classes. %>
+<%# OVERRIDE from Hyrax v5.0.1 to correct button coloring since simpleform overrides hyku classes. %>
 <%= render "shared/nav_safety_modal" %>
 <div class="card tabs">
 
@@ -52,7 +52,7 @@
             </div>
           </div>
           <div class="card-footer d-flex justify-content-end">
-            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
             <%= link_to t(:'hyrax.pages.cancel'), hyrax.admin_admin_sets_path, class: 'btn btn-light' %>
           </div>
         <% end %>
@@ -68,7 +68,7 @@
             </div>
           </div>
           <div class="card-footer d-flex justify-content-end">
-            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
             <%= link_to t(:'hyrax.pages.cancel'), hyrax.admin_admin_sets_path, class: 'btn btn-light' %>
           </div>
         <% end %>
@@ -84,7 +84,7 @@
             </div>
           </div>
           <div class="card-footer d-flex justify-content-end">
-            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
             <%= link_to t(:'hyrax.pages.cancel'), hyrax.admin_admin_sets_path, class: 'btn btn-light' %>
           </div>
         <% end %>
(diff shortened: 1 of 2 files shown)


bkiahstroud (commented):
Thanks for the PR! 

The way you've described the issue makes me wonder if this is happening elsewhere in Hyku as well... if the fundamental issue is a conflict between Hyku theming and simpleform, I think changing the class names Hyku uses would make more sense. Would something like prefixing the classes Hyku uses with `hyku_` be enough to prevent the conflict?

sephirothkod:
> Thanks for the PR!
> 
> The way you've described the issue makes me wonder if this is happening elsewhere in Hyku as well... if the fundamental issue is a conflict between Hyku theming and simpleform, I think changing the class names Hyku uses would make more sense. Would something like prefixing the classes Hyku uses with `hyku_` be enough to prevent the conflict?

The classes that I am referring to are the bootstrap classes so it is part of a library that we can't really change.

app/views/hyrax/pages/_form.html.erb
```diff
@@ -0,0 +1,96 @@
+<%# Copied from Hyrax v5.0.1 to correct button coloring since simpleform overrides hyku classes. %>
```
bkiahstroud, line 1:
```suggestion
<%# OVERRIDE from Hyrax v5.0.1 to correct button coloring since simpleform overrides hyku classes. %>
```

bkiahstroud (commented):
That makes sense to me. In that case, would you just mind adding some `OVERRIDE` comments around the lines that changed? We often `grep` for `OVERRIDE` when doing Hyrax upgrades

app/views/hyrax/pages/_form.html.erb
```diff
@@ -0,0 +1,96 @@
...
+      <div class="card labels">
+        <%= simple_form_for ContentBlock.for(:terms), url: hyrax.page_path(ContentBlock.for(:terms)), html: { class: 'nav-safety' } do |f| %>
+          <div class="card-body">
+            <div class="field form-group">
+              <%= f.label :terms %><br />
+              <%= f.text_area :terms, value: f.object.value, class: 'form-control tinymce', rows: 20, cols: 120 %>
+            </div>
+          </div>
+          <div class="card-footer d-flex justify-content-end">
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
```
sephirothkod, line 87:
```suggestion
            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
```

app/views/hyrax/pages/_form.html.erb
```diff
@@ -0,0 +1,96 @@
...
+      <div class="card labels">
+        <%= simple_form_for ContentBlock.for(:agreement), url: hyrax.page_path(ContentBlock.for(:agreement)), html: { class: 'nav-safety' } do |f| %>
+          <div class="card-body">
+            <div class="field form-group">
+              <%= f.label :agreement %><br />
+              <%= f.text_area :agreement, value: f.object.value, class: 'form-control tinymce', rows: 20, cols: 120 %>
+            </div>
+          </div>
+          <div class="card-footer d-flex justify-content-end">
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
```
sephirothkod, line 71:
```suggestion
            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
```

app/views/hyrax/pages/_form.html.erb
```diff
@@ -0,0 +1,96 @@
...
+      <div class="card labels">
+        <%= simple_form_for ContentBlock.for(:help), url: hyrax.page_path(ContentBlock.for(:help)), html: { class: 'nav-safety' } do |f| %>
+          <div class="card-body">
+            <div class="field form-group">
+              <%= f.label :help %><br />
+              <%= f.text_area :help, value: f.object.value, class: 'form-control tinymce', rows: 20, cols: 120 %>
+            </div>
+          </div>
+          <div class="card-footer d-flex justify-content-end">
+            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' %>
```
sephirothkod, line 55:
```suggestion
            <%= f.submit t(:'hyrax.pages.update'), class: 'btn btn-primary text-white mr-2' # OVERRIDE button to submit %>
```

bkiahstroud (approved):

PR #2395, opened 2024-12-05
fix guest lookup to be unscoped

orangewolf:
guests fail to to find the existing guest account because of the default scope.


ShanaLMoore (dismissed):

app/controllers/devise_guest_controllers_helpers_decorator.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
```
ShanaLMoore, line 2: An override comment could be useful.

orangewolf: I swear there was one. but yes I 100% agree with you

PR #2398, opened 2024-12-09
Fix 'basic theme' Appearance customization colors

laritakr:
Fixes color customizations for default views. Additional work is needed for themes.

ref https://github.com/scientist-softserv/palni_palci_knapsack/issues/176

A number of class codes changed, between blacklight, bootstrap, and hyrax. This PR adjusts the behavior of the appearance color customizations and/or view partials to restore the color customization behaviors.

* fixes "restore all defaults" options
* fixes some ui errors
* color customizations now function in the default views (themes still need work)

@samvera/hyku-code-reviewers


kirkkwang (approved):
I think it looks good!

PR #2405, opened 2025-01-10
:broom: renames scientst-softserv => notch8

ShanaLMoore:
renames all instances of scientist-softserv to notch8 due to company transition changes


laritakr (approved):
Thanks! looks great.

PR #2410, opened 2025-01-14
✅ Fix flaky work_show_presenter_spec

kirkkwang:
This commit will ensure that the state of the test_strategy tenant is returned to its default after the test runs because we were seeing this interfere with other tests it was left set on non-default.


ShanaLMoore:
Thanks for digging into this!

ShanaLMoore (approved):

PR #2420, opened 2025-01-16
🧹 Fix catalog gallery view

kirkkwang:
Since the Hyrax 5 update the catalog gallery index page had a grid layout of four columns while in default Hyrax has three.  Also removed some extra spacing from top row of the items.  There was also an extra border that was around the facets that were removed.  The description facet on each result has a truncate of 200 characters.

image


sjproctor (approved):
Looks good!

PR #2422, opened 2025-01-21
⚙️ Add default for simple form wrappers inline

kirkkwang:
## Story

This commit will adjust the wrappers (inline) default styling to make the elements look like they're not squished.  Which was particularly affecting the lease forms.

## Before:

image

image

## After:
image

image


ShanaLMoore (approved):
This is cool! I've never seen this config before 👀 How did you discover it? It looks so much better

PR #2424, opened 2025-01-21
Update test action to remove node_modules for CI issue

orangewolf:
CI is periodically failing because the node_modules directory is running around. make sure it is removed before starting the containers,


ShanaLMoore:
cc @kirkkwang reminding us that this exists. If you're on Pals this week can you close the loop on this work this week?

ShanaLMoore:
closing because this doesn't seem to be an issue anymore

PR #2426, opened 2025-01-22
Ensure there are UserStat records before running

ShanaLMoore:
Sidekiq is flooded with UserStatCollectionJob even when UserStats.any? => false. 

Let's not run the job is there are no user stats

### DIFF
--- a/app/jobs/user_stat_collection_job.rb
+++ b/app/jobs/user_stat_collection_job.rb
@@ -3,7 +3,8 @@ class UserStatCollectionJob < ApplicationJob
   queue_as :default
 
   def perform(*_args)
-    # Do something later
+    return unless UserStat.exists?
+
     importer = Hyrax::UserStatImporter.new(verbose: true, logging: true)
     importer.import
     UserStatCollectionJob.set(wait_until: Date.tomorrow.midnight).perform_later


kirkkwang (approved):

spec/jobs/user_stat_collection_job_spec.rb
```diff
@@ -13,10 +13,23 @@
...
-      switch!(account)
-      expect { UserStatCollectionJob.perform_now }.to have_enqueued_job(UserStatCollectionJob)
+  describe '#perform' do
+    context 'when UserStat records exist' do
+      before do
+        FactoryBot.create(:user_stat) # Ensure a UserStat record exists
+      end
+
+      it 'enqueues UserStatCollectionJob after perform' do
+        switch!(account)
```
kirkkwang, line 23: My only hesitation is that sometimes when you're switching to a different tenant in a test, it can leaky into other tests making them flaky.  I added this in another test:
https://github.com/samvera/hyku/blob/acaaf6a5af1127377cb687773f3c308a9b2a6779/spec/services/create_account_spec.rb#L11

spec/jobs/user_stat_collection_job_spec.rb
```diff
@@ -13,10 +13,23 @@
...
-  describe '#reenqueue' do
-    it 'Enques an TenantMaintenanceJob after perform' do
-      switch!(account)
-      expect { UserStatCollectionJob.perform_now }.to have_enqueued_job(UserStatCollectionJob)
+  describe '#perform' do
+    context 'when UserStat records exist' do
+      before do
+        FactoryBot.create(:user_stat) # Ensure a UserStat record exists
+      end
+
```
ShanaLMoore, line 21:
```suggestion
 after do 
   # Ensure we reset to the default tenant after each test 
   Apartment::Tenant.switch!(Apartment.default_tenant) 
 end 
```

kirkkwang (approved):

PR #2429, opened 2025-01-24
:gift: adds batch email, depositor email, and user stat to Account Settings

ShanaLMoore:
Issue: 
- https://github.com/notch8/palni_palci_knapsack/issues/209

Feature flags are now visible in  ADMIN DASHBOARD > SETTINGS > ACCOUNT. When enabled, they will trigger the relevant background jobs.

image

Additionally, if disabled a user will not see batch email frequency in their profiles. It also removed their ability to set it: 

### PROFILE SHOW PG

<details> 

#### WHEN DISABLED

image

 
image

#### WHEN ENABLED

image

image

</details> 

### PROFILE EDIT PG

<details> 

#### WHEN DISABLED

image

#### WHEN ENABLED

image

</details> 

</details> 

## :gift: add dynamic toggling for job scheduling using environment variables

abf36be366b1c7b5468645d87255b9d178b48a30

- Updated `find_or_schedule_jobs` to dynamically include jobs based on environment variables.
- Introduced the following environment flags to control job scheduling:
  - `ENABLE_BATCH_EMAIL` for `BatchEmailNotificationJob`
  - `ENABLE_DEPOSITOR_EMAIL` for `DepositorEmailNotificationJob`
  - `ENABLE_USER_STAT` for `UserStatCollectionJob`
- This allows buggy jobs to be disabled temporarily without modifying the codebase.
- Default behavior ensures only non-buggy jobs are scheduled when flags are not set.

## ♻️ Refactor to remove ENV vars in favor of account settings pattern

52e90c5687e55f92f2ebad8010a579c0ff8268a6

- Adds batch_email_notifications, depositor_email_notifications, and user_analystics to account settings
- only displays batch email frequency if batch email notifications are enabled from account settings

## 🧹 Automatically schedule jobs when account settings change

06afd3e8b3980a98b0b4825cd4b2471b21f8dd94

When certain account settings (user_analytics, batch_email_notifications,
or depositor_email_notifications) are updated, the system will now
automatically schedule the corresponding background jobs. This eliminates
the need for a server restart to activate these features.

- Added after_save callback to detect settings changes
- Added logic to compare relevant settings before and after save
- Triggers find_or_schedule_jobs when settings change

This ensures features like user analytics start working immediately
after being enabled through the settings interface.

## ♻️ Refactor account settings change detection for better readability

50040bc16d1f7252429f81aa1a8b56cc6faa0dcf

Simplifies the #schedule_jobs_if_settings_changed method to be more
intuitive and easier to maintain:

- Replace database column check with direct settings check
- Add descriptive variable names and comments
- Break complex logic into clear, distinct steps
- Improve code organization with early returns

The functionality remains the same, but the code is now more
self-documenting and easier to understand.

## ✅ update schedule_recurring_jobs specs

30e144c936af57de9eab369b5534d86000a62bda

- Add test contexts for enabled and disabled account settings
- Verify that EmbargoAutoExpiryJob and LeaseAutoExpiryJob always run
- Verify that BatchEmailNotificationJob, DepositorEmailNotificationJob,
  and UserStatCollectionJob only run when their respective settings
  are enabled

## ✅ update account settings specs

20b1c9718a1e284a435ca0c91387767ee58c2bc8

## ✅ Update create_account_spec.rb

b2794c511863f9b9be0f3e33c495989a241660a5

## ✅ Mock find_or_schedule_jobs in account endpoints spec

fe02a778cf7385cfc2e254ce9233657f36e3cf04

The account endpoints spec was failing because it attempted to
schedule jobs before the tenant schema was created. By mocking
find_or_schedule_jobs in this specific test, we avoid the schema
error while preserving the actual job scheduling behavior in other
tests.

### DIFF
--- a/spec/features/accounts_spec.rb
+++ b/spec/features/accounts_spec.rb
@@ -16,6 +16,7 @@ RSpec.describe 'Accounts administration', multitenant: true do
       allow(Apartment::Tenant).to receive(:switch).with(account.tenant) do |&block|
         block.call
       end
+      allow_any_instance_of(Account).to receive(:find_or_schedule_jobs)
     end
 
     around do |example|
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -47,6 +47,8 @@ class Account < ApplicationRecord
                      format: { with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\z/ },
                      unless: proc { |a| a.tenant == 'public' || a.tenant == 'single' }
 
+  after_save :schedule_jobs_if_settings_changed
+
   def self.admin_host
     host = ENV.fetch('HYKU_ADMIN_HOST', nil)
     host ||= ENV['HOST']
@@ -182,7 +184,6 @@ class Account < ApplicationRecord
       LeaseAutoExpiryJob
     ]
 
-    # Add conditional checks based on a config or environment variable
     jobs_to_schedule << BatchEmailNotificationJob if batch_email_notifications
     jobs_to_schedule << DepositorEmailNotificationJob if depositor_email_notifications
     jobs_to_schedule << UserStatCollectionJob if user_analytics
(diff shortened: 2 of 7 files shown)


orangewolf (changes requested):
The concept here is fine, but the implementation needs rework. “ENABLED” is implied. The flag is either on or off and that word is just extra. Also, all settings in Hyku should use AccountSettings. The automatically handles the Boolean issue, makes them settable per tenant or globally and provides consistency.

orangewolf:
I recommend BATCH_EMAIL_NOTIFICATION for the env var name

ShanaLMoore:
> The concept here is fine, but the implementation needs rework. “ENABLED” is implied. The flag is either on or off and that word is just extra. Also, all settings in Hyku should use AccountSettings. The automatically handles the Boolean issue, makes them settable per tenant or globally and provides consistency.

Thanks for the reminder. That's definitely easier to manage. I'll make the updates as suggested.

spec/features/accounts_spec.rb
```diff
@@ -16,6 +16,7 @@
       allow(Apartment::Tenant).to receive(:switch).with(account.tenant) do |&block|
         block.call
       end
+      allow_any_instance_of(Account).to receive(:find_or_schedule_jobs)
```
ShanaLMoore, line 19: Thoughts about this? This spec was failing because of the after_save addition. cc @orangewolf

PR #2430, opened 2025-01-24
🧹 Clean up featured collection UI

kirkkwang:
This commit will add a few more html tags to the featured collections, both from the admin end and public facing end, to make it look more like the featured works.  This will fix custom font issues where the titles of the collections were not getting the correct font that was set.

### Admin view:
image

### Public view:
image

Ref:
- https://github.com/notch8/palni_palci_knapsack/issues/178

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -519,6 +519,15 @@ body.public-facing {
   border: 1px solid #ccc;
 }
 
+.new_featured_collection_list {
+  width: 100%;
+}
+
+// align featured collections with featured works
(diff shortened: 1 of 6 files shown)


ShanaLMoore (commented):
Could you add issue numbers to your PRs so that github can link them? It'll make things much easier to track for our future selves. Besides that it looks good to me.

kirkkwang:
> Could you add issue numbers to your PRs so that github can link them? It'll make things much easier to track for our future selves. Besides that it looks good to me.

Sounds good!

ShanaLMoore (approved):

PR #2440, opened 2025-01-31
Add member methods on AdminSetResource

kirkkwang:
This commit will introduce two instance methods on the AdminSetResource:
- #member_of
- #member_collection_ids

These methods will make it possible for users to query for resources are on the given AdminSetResource.  This is particularly useful for the WillowSword gem.

Ref:
- https://github.com/notch8/palni-palci/issues/1071


laritakr (commented):
Other than the lint error, looks good.

laritakr (approved):

PR #2444, opened 2025-02-04
Pitt account settings follow up

ShanaLMoore:
Disable Pitt feature that sends user stats and emails, until these feature bugs get fixed: 

:batch_email_notifications
:depositor_email_notifications
:user_analytics

The settings are no longer visible on the account form: 

image


laritakr (approved):
I didn't realize that settings could be disabled so easily. This is great!

PR #2452, opened 2025-02-10
Change to released Hyrax version

laritakr:
Updates the hyrax version to the 5.0.4 release (the most recent pre-rails-7.2 version) which encompasses all bugfixes to date.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,7 +53,7 @@ gem 'good_job', '~> 2.99'
 gem 'googleauth', '~> 1.9.0'
 gem 'google-protobuf', force_ruby_platform: true # required because google-protobuf is not compatible with Alpine linux
 gem 'grpc', force_ruby_platform: true # required because google-protobuf is not compatible with Alpine linux
-gem 'hyrax', '~> 5.0.4' # lock hyrax to the pre-rails 7.2 version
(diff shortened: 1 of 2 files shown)


laritakr:
I don't know what is wrong, but I can't use hyrax v5.0.4 in the Gemfile. When I go back to the hash prior to the version bump, everything builds correctly. 

 I am leaving Hyrax at the prior hash for now.

laritakr:
I see that this causes the specs to run, but why do you have to use a github ref? Is that normal for hyrax? I don't do that for other gems.

ShanaLMoore:
When you use gem 'hyrax', '~> 5.0.4', Bundler fetches the published gem from RubyGems.org. This version has been packaged according to the gemspec, which excludes the spec directory content as we saw in the gemspec file.
However, when you use gem 'hyrax', '5.0.4', github: 'samvera/hyrax', ref: 'v5.0.4', Bundler fetches directly from the GitHub repository. When fetching from GitHub, you get the entire repository content at that ref/tag, including the spec directory and all test files, not just what's packaged in the gem.
In other words:

RubyGems version = packaged gem with spec files excluded ([due to gemspec rules](https://github.com/samvera/hyrax/blob/main/hyrax.gemspec#L21))
GitHub version = full repository including all files

This explains why using the GitHub ref works - you're getting the complete source code including test support files, rather than the packaged gem which intentionally excludes them.

laritakr:
Closing PR - we can just continue to use the `main_before_rails_72` branch directly.

PR #2453, opened 2025-02-11
I194 google analytics

kirkkwang:
## 🧹 Revert account_settings and spec

b58501579a435b6e0c5f14e76620f6f8933290ee

This commit will turn back on the batch email, depositor email, and
google analytics settings for the tenants (accounts).

## 🧹 Remove early return and over add #sorted_users

4d91549f1aee05d8427a09bec9cd2bfdcc951da2

This commit will remove the early return in the user_stat_collection_job
since it was too aggressive and will have a condition where it will
never run.  Also, adding an override to #sorted_users to acocunt for
multitenancy in Hyku.

## Update Hyrax to bring in bug fixes for GA4

df8a46edc3ae373210ee0a766e7c61e1f09a1ac7

This commit will update the Hyrax revision to bring in the bug fixes for
Google Analytics 4 and also update the logic in Hyku's ga4 partial to
account for multitenancy more accurately.

image

image


app/jobs/user_stat_collection_job.rb
```diff
@@ -3,8 +3,6 @@ class UserStatCollectionJob < ApplicationJob
   queue_as :default
 
   def perform(*_args)
-    return unless UserStat.exists?
```
ShanaLMoore, line 6: Not blocking, but I think we wanted to move this down to after importer.import instead of completely removing it.

spec/jobs/user_stat_collection_job_spec.rb
```diff
@@ -28,12 +28,5 @@
         expect { UserStatCollectionJob.perform_now }.to have_enqueued_job(UserStatCollectionJob)
       end
     end
-
```
ShanaLMoore, line 31: see comment from above.

ShanaLMoore: I'll make the changes and commit

ShanaLMoore (approved):

PR #2456, opened 2025-02-13
Allow for theme color defaults

laritakr:
## Summary 
Creates a simple way to define and set default colors for themes in the admin UI. 

The colors to override are defined by attaching them in the `home_themes.yml`. The button to set the colors to the default for the theme will only be displayed if the theme defines custom colors.

<details>
<summary>Video</summary>

[recorded (1).webm](https://github.com/user-attachments/assets/9ea4bc9c-b683-4d7b-85fb-98d3a1f60452)

</details>


orangewolf (approved):
yep, I 100% dig this approach. Great work @laritakr!

PR #2457, opened 2025-02-14
Switch from Sidekiq to GoodJob as the Default Worker

ShanaLMoore:
Replaced Sidekiq with GoodJob as the default to simplify infrastructure and improve job persistence by using PostgreSQL instead of Redis. Sidekiq is still supported as a configuration option.

### Why?
- No Redis required—jobs are stored in PostgreSQL. For Sidekiq, if we lose the jobs we lose them forever 😭 
- Jobs persist in the database, making them easier to inspect and recover (`GoodJob::Execution.all` in Rails console)
- Ensures transactional consistency with database commits
- Built-in concurrency controls and better observability

Developers should update their environments accordingly and use `./bin/worker` to kick off the worker since it is smart about launching the correct worker type.

Goodjobs dashboard is available at the same location as the sidekiq one. Both are available at `/jobs`

image

Success ✅ 
image


laritakr (commented):
Shouldn't you run the migration & save the schema changes along with this?

orangewolf:
> Shouldn't you run the migration & save the schema changes along with this?

The good jobs migrations are already part of Hyku. Empty tables are cheap and the logic to make them conditional was too complicated. So we just let them run no matter which worker you used.

sjproctor (approved):

PR #2464, opened 2025-02-19
Fix file permissions for branding and uploads directories

ShanaLMoore:
Changes podSecurityContext to align with EFS filesystem permissions (50002). Previously files were being created with overly restrictive permissions (600) preventing web access to uploaded logos and branding images.

- Updates runAsUser, runAsGroup, and fsGroup to 50002 to match EFS POSIX user/group
- Sets fsGroupChangePolicy to OnRootMismatch for efficient permission handling
- Applied to both main pod and worker deployments

NOTE: I'm trying web change first before applying the same to worker


ShanaLMoore:
not needed. the files weren't loading because File ACL wasn't checked on the tenant's account setting.

PR #2466, opened 2025-02-20
🐛 Rework home page featured collections layout

sjproctor:
# Story: [i240] Featured Collections Home Page Layout

Ref:
- https://github.com/notch8/palni_palci_knapsack/issues/240

## Expected Behavior Before Changes

Previous changes to the Featured Collections section on the home page layout of the Neutral Theme for logged out users created layout issues in the Featured Collections section on the home page layout of the Default Theme

## Expected Behavior After Changes

A conditional statement will allow theme with Featured Collections in a grid to have different styling applied than the themes with a table layout

## Screenshots / Video

<details>
<summary>Photo: theme with a grid layout</summary>

Screenshot 2025-02-19 at 6 15 11 PM

</details>

<details>
<summary>Photo: theme with a table layout</summary>

image

</details>

@samvera/hyku-code-reviewers

### DIFF
--- a/app/views/hyrax/homepage/_featured_collection_section.html.erb
+++ b/app/views/hyrax/homepage/_featured_collection_section.html.erb
@@ -1,8 +1,8 @@
 <div id="featured_collections">
   <% if @featured_collection_list.empty? %>
-    <p id='no-collections'><%= t('hyrax.homepage.featured_collections.no_collections') %></p>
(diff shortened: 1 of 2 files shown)


laritakr (changes requested):
I don't think we should condition the views with the hardcoded theme name. Each theme has a theme view directory where you should be able to make a copy of the partial which should work for that theme. 

Is there a reason this didn't work for this partial?

PR #2471, opened 2025-02-24
Fix sipity with migrations

laritakr:
## Summary

Refs
- https://github.com/samvera/hyrax/issues/7028
- https://github.com/notch8/palni_palci_knapsack/issues/247

Workflow management is broken between pre-valkyrie works and valkyrie resources. Due to how the `proxy_for_global_id` was being built, the Entity could not be found, resulting in no valid workflows being found either. This resulted in numerous errors. 

f3feb43733df47d4667f50a24e505d373dd75838 begins to work on the fix, by reworking how we find the entity for a work/resource. There may still be remaining issues. 

This work should be moved into Hyrax once it is fully fixed.


orangewolf (approved):
I don't think this covers all possible cases, but it definitely covers some cases. going to approve it so I don't hold up the train, but I'll revisit it with a fresh brain tomorrow.

Issue #2485, opened 2025-03-12
Video embed feature is broken


laritakr:
## Summary

The video embed feature no longer works. 

When a new work is added with a video embedded, nothing happens: [link to example in hyku demo site](https://demo.hyku-demo.notch8.cloud/concern/generic_works/aae22d81-7438-437d-996a-acd43f1d92c8?locale=en)

When an existing work prior to the bug has a video embed link, the UI is very bad, overlaying the metadata: [link to a work in adventist's repo](https://adl.b2.adventistdigitallibrary.org/concern/generic_works/23000121__inasmuch)

PR #2486, opened 2025-03-12
Fixes video embed viewer

laritakr:
Viewer uses a different indexed term than the yaml specified. Changes video_embed_tesion to video_embed_tesim.

Fixes https://github.com/samvera/hyku/issues/2485


kirkkwang (approved):
Good find on this, though i do wonder now if it can have multiple video embeds

laritakr:
> Good find on this, though i do wonder now if it can have multiple video embeds

No, multiple is specified at both the main level and the form level.

Issue #2495, opened 2025-03-18
Update Hyrax v5.0.4 Helm Chart to Support Custom Init Container Commands for Hyku (off main)


aprilrieger:
The change in Hyku requiring init scripts to run from ./bin/ means we need to update the Helm deploy for Hyrax (v5.0.4) so that we can pass a custom command while also setting a default command in the values.yaml file.

### Modify Helm Chart Templates in Hyrax - v5.0.4

- [ ] Update the deployment templates to allow passing a custom init container command for `load-solr-config`
- [ ] Update the deployment templates to allow passing a custom init container command for `db-setup`
- [ ] Update the worker-deployment templates to allow passing a custom init container command for `service-wait`
- [ ] Update Default values.yaml File in Hyrax, add the default command for Hyrax.
- [ ] Update Default values.yaml File in Hyku, add the default command for Hyku.
- [ ] Publish a New Chart Version in Hyrax
- [ ] Update Hyku to use that chart

orangewolf:
@aprilrieger would it be easier to fix the Dockerfile to set the PATH correctly and allow running them w/o the `bin/` prefix?

aprilrieger:
@orangewolf yes!

aprilrieger:
@orangewolf the solution we worked on together last week is failing. I noted my findings in the PR after checking out locally.

PR #2498, opened 2025-03-21
DRAFT: Throw it in the bin aprils' version

aprilrieger:
fixed https://github.com/samvera/hyku/issues/2495


docker-compose.yml
```diff
@@ -195,6 +195,10 @@ services:
         - ghcr.io/samvera/hyku/base:latest
         - ghcr.io/samvera/hyku/web:latest
         - ghcr.io/samvera/hyku/worker:latest
+    entrypoint: ["sh", "-c"]
+    command:
+    - >
+      echo 'running migrations'; ./bin/db-migrate-seed.sh
```
aprilrieger, line 201: Since we change the path we aren't able to run this in the web container. The deployment doesn't reference the db-migrate-seed.sh script so I think it's safe to pull out. Open to feedback.

docker-compose.yml
```diff
@@ -229,14 +233,13 @@ services:
     <<: *app
     environment:
       - CONFDIR=/app/samvera/hyrax-webapp/solr/conf
+      - PATH=/app/samvera/hyrax-webapp/bin:${PATH}
```
aprilrieger, line 236: For some reason I couldn't get the echo to set it correctly in the Dockerfile. 🤷🏼‍♀️

docker-compose.yml
aprilrieger: welp, i forgot about the db-setup init container on the web helm deploy :( So I do need to solve this in the web service. Taking this out.

aprilrieger:
To get the deployment working, I had to manually intervene by setting the correct path in each init container, ensuring they executed bundle properly without attempting to run it through Ruby.

aprilrieger:
Notes:
On deployment

I need to be able to execute the initialize scripts with the altered path to grab them.

Web: https://github.com/samvera/hyrax/blob/v5.0.4/chart/hyrax/templates/deployment.yaml

load-solr-config: `PATH=/app/samvera/hyrax-webapp/bin:/usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

For this command to find the correct script in the app/samvera/hyrax-webapp/bin dir:
```
command:
            - sh
            - -c
            - >
              solrcloud-upload-configset.sh /app/samvera/hyrax-webapp/solr/conf &&
              solrcloud-assign-configset.sh
```

db-setup: `PATH=/app/samvera/hyrax-webapp/bin:/usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
```
command:
            - sh
            - -c
            - db-migrate-seed.sh
```
BUT once the db-migrate-seed.sh script is run I need the `PATH` back to `PATH = /usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

OR it hits: /app/samvera/hyrax-webapp/bin/bundle

```
#!/usr/bin/env ruby
ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
load Gem.bin_path('bundler', 'bundle')
```
And when the bundle commands execute(ex: bundle exec rails db:migrate) from the script it errors with:
```
/app/samvera/hyrax-webapp/bin/bundle: 4: Syntax error: "(" unexpected
```
Then when we bring up the actual container I need the `PATH` set back to `PATH = /usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` so that when devs get back in the console they can bundle exec rails whatever and it not be all janky and unexpected.

hyrax: PATH = /usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

Same can be seen during the worker deployment. Also note that the chart still uses db-wait and we are moving to service-wait.

Worker: https://github.com/samvera/hyrax/blob/v5.0.4/chart/hyrax/templates/deployment-worker.yaml

db-wait: `PATH=/app/samvera/hyrax-webapp/bin:/usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
```
command:
            - sh
            - -c
            - service-wait.sh "$REDIS_HOST:6379"
```
Then when we bring up the actual container I need the `PATH` set back to `PATH = /usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`.

hyrax-worker: `PATH=/app/samvera/hyrax-webapp/bin:/usr/local/bundle/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

PR #2502, opened 2025-03-24
Use stack_car as the proxy going forward

orangewolf:
# TODO

- [ ] Update readme

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,8 +18,6 @@ x-app: &app
     - assets:/app/samvera/hyrax-webapp/public/assets:cached
     - cache:/app/samvera/hyrax-webapp/tmp/cache:cached
     - .:/app/samvera/hyrax-webapp
-  networks:
-    internal:
 
 volumes:
   assets:
@@ -34,7 +32,9 @@ volumes:
   zoo:
 
 networks:
-  internal:
+  default:
+    name: network.direct
+    external: true
 
 services:
   zoo:
@@ -46,8 +46,6 @@ services:
     volumes:
       - zoo:/data
       - zk:/datalog
-    networks:
-      internal:
     healthcheck:
       test:
         [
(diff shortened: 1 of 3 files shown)


docs/getting-started.md
```diff
@@ -29,20 +28,21 @@ newgrp docker
...
+    Hyku makes heavy use of domain names to determine which tenant to serve. On MacOS/Linux, it is recommended to use [Stack Car](https://github.com/notch8/stack_car) to handle the necessary SSL certs and proxy setup.
 
-    #### Dory Installation
+    #### Stack Car Installation
 
     ```bash
-    gem install dory
-    dory up
+    gem install stack_car
+    sc proxy cert # Only need this once per stack_car version
```
ShanaLMoore, line 37: I got this message when I ran this command. Where's the password? 

```ruby
Enter the password found https://github.com/Upinel/localhost.direct?tab=readme-ov-file#a-non-public-ca-certificate-if-you-have-admin-right-on-your-development-environment-you-can-use-the-following-10-years-long-pre-generated-self-signed-certificate
```

ShanaLMoore: nevermind, i followed the link to where it said the password was. Could we add that to the readme?

ShanaLMoore:
I'll miss the hyku.test pattern but confirming that this works for me.

docker-compose.yml
```diff
@@ -252,8 +284,6 @@ services:
     command: redis-server
     volumes:
       - redis:/data
-    networks:
-      internal:
 
   chrome:
```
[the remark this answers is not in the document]
orangewolf: I think it is already. we left the VIRTUAL_HOST in place so that existing dory will still work at least fo now

docs/getting-started.md
orangewolf: added more info to getting-started.md

PR #2519, opened 2025-04-09
Hyrax update for queued indexing

orangewolf:


laritakr (commented):
Needs lint fix. And I saw a PR sitting in Hyrax that I wanted to make sure we don't need.... https://github.com/samvera/hyrax/pull/7052

Also need to update bulkrax so we can get rid of the other hyku PR https://github.com/samvera/hyku/pull/2517

PR #2528, opened 2025-04-17
Add custom query for IiifPrint

laritakr:
## Summary

This query had previously been defined in Bulkrax. Bulkrax needed a change to be able to find metadata without the model, but the original query is still needed in IiifPrint.

Ideally this should be defined in IiifPrint directly, but the current setup has the custom query for bulkrax being registered here. To avoid confusion, this query is being added to Hyku since it is registered in Hyku.

If we determine a better way to register custom queries in gems, this can be moved to IiifPrint.

refs: https://github.com/samvera/bulkrax/pull/1025

### DIFF
--- a/app/services/hyrax/custom_queries/find_by_model_and_property_value.rb
+++ b/app/services/hyrax/custom_queries/find_by_model_and_property_value.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+##
+# A custom query used by IiifPrint to find a resource by its model and title.
+# Ideally should be defined in IiifPrint, but currently gem custom queries are registered in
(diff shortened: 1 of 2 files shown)


ShanaLMoore (approved):

app/services/hyrax/custom_queries/find_by_model_and_property_value.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
```
kirkkwang, line 1: would be nice to have IIIF Print generator create this file for future applications but maybe that's a pipe dream

kirkkwang (approved):

PR #2530, opened 2025-04-19
Cache money

orangewolf:
Adds rack-attack to throttle bots and scrappers, redis based sessions for speed and allows caching catalog#index and concern#show and concern#manifest routes if user is not logged in.

Tosses in a staging environment, because there should just be one even if its the same as prod as rails keys off it.


app/controllers/application_controller.rb
```diff
@@ -26,12 +26,27 @@ class ApplicationController < ActionController::Base
...
+    return if current_user || !cached_route?(request.path)
+
+    if response.headers["Cache-Control"].blank? || !(response.headers["Cache-Control"].include?('no-store') || response.headers["Cache-Control"].include?('no-cache'))
+      # this skips sending a session cookie # (a session cookie will cause cloudflare to avoid caching it)
+      request.session_options[:skip] = true
+    end
+  end
+
+  def cached_route?(path)
+    path.start_with?('/concern', '/catalog')
```
bkiahstroud, line 47: Form routes start with `/concern` (e.g. `/concern/generic_works/new`), but I think this should be fine since you also have `if current_user` in `#clear_session_cookies`

bkiahstroud (approved):
Not blocking, just curious: how does show page caching affect the UV and PDF.js? Is it a non-factor since they're `iframe`s?

app/controllers/application_controller.rb
orangewolf: also the "new page" isn't an issue because its just a blank form. the POST and PUT aren't ever cached

bkiahstroud (approved):

PR #2534, opened 2025-04-23
Only throttle `/catalog` requests

bkiahstroud:
The throttle limit previously had to be increased to 500 by default because each thumbnail a page loads counts as a request. Bots primarily scrape `/catalog` pages, so we're narrowing our focus to specifically throttle those requests more aggressively

@samvera/hyku-code-reviewers

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -7,7 +7,9 @@ throttle_off = ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ATTACK_RATE_T
 
 unless throttle_off
   Rack::Attack.throttle('throttle catalog requests by ip', limit: limit, period: period) do |req|
-    req.ip if req.path.starts_with?('/catalog')
+    if req.path.starts_with?('/catalog')
+      req.get_header('HTTP_X_ORIGINAL_FORWARDED_FOR') || req.ip
+    end
   end
 
   FileUtils.mkdir_p(Rails.root.join('log', 'rack_attack'))
@@ -21,10 +23,11 @@ unless throttle_off
 
   ActiveSupport::Notifications.subscribe('throttle.rack_attack') do |_name, _start, _finish, _request_id, payload|
     req = payload[:request]
+    origin_ip = req.get_header('HTTP_X_ORIGINAL_FORWARDED_FOR') || req.ip
 
     throttle_logger.info(
-      "#{req.ip} " \
       "#{req.env['rack.attack.match_type']} " \
+      "#{origin_ip} " \
       "#{req.env['rack.attack.match_data'][:count]}/#{req.env['rack.attack.match_data'][:limit]} " \
       "#{req.env['rack.attack.match_data'][:period]}s " \
       "#{req.request_method} " \
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -10,6 +10,7 @@ unless throttle_off
     req.ip if req.path.starts_with?('/catalog')
   end
 
+  FileUtils.mkdir_p(Rails.root.join('log', 'rack_attack'))
   throttle_logger = ActiveSupport::Logger.new(
     Rails.root.join('log', 'rack_attack', 'throttled_requests.log'),
     'daily'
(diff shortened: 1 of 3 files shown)


ops/staging-deploy.tmpl.yaml
```diff
@@ -12,6 +12,8 @@ derivativesVolume:
 uploadsVolume:
   storageClass: aws-efs
   size: 200Gi
+logs:
+  storageClass: efs-sc
```
bkiahstroud, line 16: @orangewolf bringing extra attention to this addition

orangewolf: you don't have enough parts here I don't think. you'd need to do something to the hyrax chart to make `logs:` a valid option. as a less cumbersome alternative, see the suggestion above

ops/demo-deploy.tmpl.yaml
```diff
@@ -35,6 +37,9 @@ extraVolumeMounts: &volMounts
   - name: uploads
     mountPath: /app/samvera/hyrax-webapp/storage/files
     subPath: storage-files
+  - name: logs
```
orangewolf, line 40:
```suggestion
  - name: uploads
```
I suggest you just use the uploads mount and not try to make a new mount

orangewolf (changes requested):

config/initializers/rack_attack.rb
```diff
@@ -1,6 +1,34 @@
...
+
+# No more that 20 requests per minute per ip
+limit = ENV.fetch('HYKU_ATTACK_RATE_LIMIT', 20).to_i
 period = ENV.fetch('HYKU_ATTACK_RATE_PERIOD', 60).to_i
 throttle_off = ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYKU_ATTACK_RATE_THROTTLE_OFF', Rails.env.development? || Rails.env.test?))
-Rack::Attack.throttle("requests by ip", limit: limit, period: period, &:ip) unless throttle_off
+
+unless throttle_off
+  Rack::Attack.throttle('throttle catalog requests by ip', limit: limit, period: period) do |req|
+    req.ip if req.path.starts_with?('/catalog')
```
bkiahstroud, line 10: If behind a load balancer/proxy, `req.ip` is returning the load balancer IP, which is not what we want. We should explore using `req.get_header('HTTP_X_ORIGINAL_FORWARDED_FOR')` and falling back on `req.ip` if `HTTP_X_ORIGINAL_FORWARDED_FOR` is blank

orangewolf: or we should figure out which deployments this is happening for and fix them ;-)

orangewolf (approved):

config/initializers/rack_attack.rb
bkiahstroud: Can you elaborate? I'm hesitant to merge this before addressing the issue; with a request limit of 20, and `req.ip` being the LBs, I think a busy site could effectively DoS itself for all users

orangewolf: which apps have the request.ip as the LBs. those apps are broken and should be fixed. if you grab a list, I can help you solve them

bkiahstroud: Hyku staging for sure, that's where I've been testing

config/initializers/rack_attack.rb
```diff
@@ -1,6 +1,37 @@
 # frozen_string_literal: true
-# No more that 500 requersts per minitue per ip
-limit = ENV.fetch('HYKU_ATTACK_RATE_LIMIT', 500).to_i
+
+# No more that 20 requests per minute per ip
+limit = ENV.fetch('HYKU_ATTACK_RATE_LIMIT', 20).to_i
```
aprilrieger, line 4: I like how these are env vars we can update live if needed.

aprilrieger (approved):

config/initializers/rack_attack.rb
bkiahstroud: Agreed, credit for these goes to @orangewolf. It's a pattern I'd love to see more often :grin:

PR #2537, opened 2025-04-25
Add Account setting to exclude chosen fields from search results page

bkiahstroud:
### Before

<details><summary>search result</summary>

image

</details>

### Update setting

<details><summary>setting</summary>

image

</details>

### After

<details><summary>search result</summary>

image

</details>

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -44,6 +44,7 @@ module AccountSettings
     setting :google_scholarly_work_types, type: 'array', disabled: true
     setting :geonames_username, type: 'string', default: ''
     setting :gtm_id, type: 'string'
+    setting :hidden_index_fields, type: 'string', default: 'title'
     setting :locale_name, type: 'string', disabled: true
     setting :monthly_email_list, type: 'array', disabled: true
     setting :oai_admin_email, type: 'string', default: 'changeme@example.com'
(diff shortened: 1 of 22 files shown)


app/models/concerns/account_settings.rb
```diff
@@ -44,6 +44,7 @@ module AccountSettings
     setting :google_scholarly_work_types, type: 'array', disabled: true
     setting :geonames_username, type: 'string', default: ''
     setting :gtm_id, type: 'string'
+    setting :hidden_index_fields, type: 'string', default: 'title'
```
orangewolf, line 47: is title really a good one to hide by default?

orangewolf (commented):
code looks good. I question hiding `title` by default in new tenants.

app/models/concerns/account_settings.rb
bkiahstroud: Hiding title mirrors the state before this change:
```ruby
config.add_index_field 'title_tesim', label: "Title", itemprop: 'name', if: false
```
Otherwise, title would show up twice in the search results; once as the header for the result, and once in its list of metadata

orangewolf: great, this is gtg then

orangewolf (approved):

PR #2540, opened 2025-04-30
🐛 Fix video_embed

kirkkwang:
The root of the problem is that we changed `video_embed_tesim` to `video_embed_tesi` which makes sense since it's not a multiple field.

Ref:
- http://github.com/notch8/wvu_knapsack/issues/44

image


ShanaLMoore:
This was already working for hyku demo. Does this solve the reason why it didn't work for flexible metadata? ie if a user sets this property in their metadata profile?

ShanaLMoore (approved):

PR #2573, opened 2025-06-11
no-method-error-bulkrax-identifier-tesim

aprilrieger:
Fix `NoMethodError` for `bulkrax_identifier_tesim` on GenericWorks seen during Bulkrax::DeleteWorkJob

Add presenter method to safely access `bulkrax_identifier` from Solr

During Bulkrax jobs (e.g., `Bulkrax::DeleteWorkJob`), a `NoMethodError` was being raised due to dynamic attempts to call `bulkrax_identifier_tesim` on a `GenericWork`. This method does not exist on the model, as it is a Solr-only field.

To resolve this, we:
- Defined a `#bulkrax_identifier` method in `Hyrax::GenericWorkPresenter` that returns the first value from `solr_document['bulkrax_identifier_tesim']`
- Added a spec to ensure the presenter method behaves correctly

This prevents both UI and background job failures caused by incorrect assumptions about model method availability.

### Example stack trace:
```ruby
NoMethodError: undefined method `bulkrax_identifier_tesim' for #<GenericWork ...>
```

### Changes proposed in this pull request:
* Add `#bulkrax_identifier` to `GenericWorkPresenter`
* Return first item from Solr field `bulkrax_identifier_tesim`
* Add presenter spec to confirm expected output

Fixes: https://github.com/notch8/palni_palci_knapsack/issues/348

### DIFF
--- a/app/presenters/hyrax/generic_work_presenter.rb
+++ b/app/presenters/hyrax/generic_work_presenter.rb
@@ -5,5 +5,9 @@
 module Hyrax
   class GenericWorkPresenter < Hyku::WorkShowPresenter
     delegate :abstract, to: :solr_document
+
+    def bulkrax_identifier
+      Array(solr_document['bulkrax_identifier_tesim']).first
+    end
   end
 end
--- a/spec/presenters/hyrax/generic_work_presenter_spec.rb
+++ b/spec/presenters/hyrax/generic_work_presenter_spec.rb
@@ -5,6 +5,21 @@
 require 'rails_helper'
 
 RSpec.describe Hyrax::GenericWorkPresenter do
+  let(:solr_document) do
+    SolrDocument.new(
+      id: '123abc',
+      'bulkrax_identifier_tesim' => ['abc123']
+    )
+  end
+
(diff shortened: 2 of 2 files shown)


laritakr:
I really appreciate how you try to proactively solve issues for our clients. This has been a major problem for a number of the Pals tenants, and Adventist also had to fight this problem. I'm sure everyone would like a solution to this pain point. 

I've seen this coming up frequently as I work on the Pitt problems so I did a little digging last week as well, and I'm glad to pair if you dig into it further. The problem goes beyond Generic Works... from what I have seen, it will happen any time any ActiveFedora work is edited or deleted through Bulkrax once the app has been Valkyrized. 

I'm not sure how many other Bulkrax bugs exist for this migration situation, but the fix is probably a lot more involved than this, and any fix really needs to be addressed in Bulkrax. I'm afraid that updating Hyku with a workaround could lead to confusion or unintended side effects. Hopefully we can find a Bulkrax solution eventually.

At this point, the client has been notified to not attempt to edit or delete works until the tenants have been migrated.

PR #2608, opened 2025-07-01
I960 controlled vocab flexible metadata

ShanaLMoore:

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

@samvera/hyku-code-reviewers

### DIFF
--- a/app/helpers/hyrax/form_helper_behavior.rb
+++ b/app/helpers/hyrax/form_helper_behavior.rb
@@ -16,32 +16,48 @@ module Hyrax
 
     def remote_authority_config_for(source_name)
       remote_authorities = {
-        'loc_subjects' => {
+        'loc/subjects' => {
           url: "/authorities/search/loc/subjects",
           type: 'autocomplete'
         },
-        'loc_names' => {
+        'loc/names' => {
           url: "/authorities/search/loc/names", 
           type: 'autocomplete'
         },
-        'loc_genre_forms' => {
+        'loc/genre_forms' => {
           url: "/authorities/search/loc/genreForms",
+          type: 'autocomplete'
+        },
+        'loc/countries' => {
+          url: "/authorities/search/loc/countries",
           type: 'autocomplete'
         },
         'mesh' => {
           url: "/authorities/search/mesh",
           type: 'autocomplete'
         },
-        'aat' => {
+        'getty/aat' => {
           url: "/authorities/search/getty/aat",
+          type: 'autocomplete'
+        },
+        'getty/tgn' => {
+          url: "/authorities/search/getty/tgn",
+          type: 'autocomplete'
+        },
+        'getty/ulan' => {
+          url: "/authorities/search/getty/ulan",
           type: 'autocomplete'
         },
         'geonames' => {
           url: "/authorities/search/geonames",
           type: 'autocomplete'
         },
-        'fast_topics' => {
-          url: "/authorities/search/oclc_fast/topic",
+        'fast' => {
+          url: "/authorities/search/fast",
+          type: 'autocomplete'
+        },
+        'discogs' => {
+          url: "/authorities/search/discogs",
           type: 'autocomplete'
         }
       }
--- a/app/helpers/hyrax/form_helper_behavior.rb
+++ b/app/helpers/hyrax/form_helper_behavior.rb
@@ -28,6 +28,14 @@ module Hyrax
           url: "/authorities/search/loc/genreForms",
           type: 'autocomplete'
         },
+        'mesh' => {
+          url: "/authorities/search/mesh",
+          type: 'autocomplete'
+        },
+        'aat' => {
+          url: "/authorities/search/getty/aat",
+          type: 'autocomplete'
+        },
         'geonames' => {
           url: "/authorities/search/geonames",
           type: 'autocomplete'
--- a/app/helpers/hyrax/form_helper_behavior.rb
+++ b/app/helpers/hyrax/form_helper_behavior.rb
@@ -32,10 +32,6 @@ module Hyrax
           url: "/authorities/search/loc/countries",
           type: 'autocomplete'
         },
-        'mesh' => {
-          url: "/authorities/search/mesh",
-          type: 'autocomplete'
-        },
         'getty/aat' => {
           url: "/authorities/search/getty/aat",
           type: 'autocomplete'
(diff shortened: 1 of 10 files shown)


app/helpers/hyrax/form_helper_behavior.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
```
ShanaLMoore, line 1: Does this belong in hyrax?

app/views/records/edit_fields/_default.html.erb
```diff
@@ -4,14 +4,78 @@
 <% return if f.object.try(:hidden?, key) %>
 
 <% record = f.object.model %>
-<% if f.object.multiple? key %>
+<% controlled_vocab_config = controlled_vocabulary_options_for(key, record.class) %>
+
```
ShanaLMoore, line 8: Todo: check to see if there are other partials in this path to delete

app/views/records/edit_fields/_subject.html.erb
```diff
@@ -0,0 +1,43 @@
+<%# OVERRIDE: Use dynamic controlled vocabulary system for subject field %>
```
ShanaLMoore: OVERRIDE HYRAX - this is required because it's defined in Hyrax, which may also means if any wants to modify behavior of properties of this directory will need overrides as well. 

image

ShanaLMoore: @orangewolf does this mean that hyrax should delete all these files? And _default partial should live in hyrax too?

ShanaLMoore:
```suggestion
<%# OVERRIDE: Hyrax v5.0.0@_subject.html.erb - Use flexible metadata controlled vocabulary %>
```

db/schema.rb
```diff
@@ -631,6 +631,22 @@
     t.index ["uri"], name: "index_qa_local_authority_entries_on_uri", unique: true
   end
 
+  create_table "qa_mesh_trees", id: :serial, force: :cascade do |t|
```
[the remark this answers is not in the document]
ShanaLMoore: this was generated by running the install command. I'm going to leave it as is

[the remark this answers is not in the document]
ShanaLMoore: this was generated by the qa install command. I'm tempted to leave it as is 🤔

app/views/records/edit_fields/_default.html.erb
ShanaLMoore: OVERRIDE HYRAX - this is required because it's defined in Hyrax, which may also means if any wants to modify behavior of properties of this directory will need overrides as well. 

@orangewolf does this mean that hyrax should delete all these files? And the dynamic _default partial should live in hyrax too? if so, is there a particular branch I should use or create my own off of 5.0-flexible?

image

ShanaLMoore: for example, license still renders even though the controlled_value.source is null

image

```
  license:
    available_on:
      class:
      - Hyrax::FileSet
      - CollectionResource
      - GenericWorkResource
      - ImageResource
      - OerResource
      - EtdResource
    cardinality:
      minimum: 0
    multi_value: true
    controlled_values:
      format: http://www.w3.org/2001/XMLSchema#string
      sources:
      - 'null'
    display_label:
      default: License
    index_documentation: displayable, searchable
    indexing:
    - license_sim
    - license_tesim
    form:
      primary: false
    property_uri: http://purl.org/dc/terms/license
    range: http://www.w3.org/2001/XMLSchema#string
    sample_values:
    - http://creativecommons.org/licenses/by/3.0/us/
    view:
      render_as: external_link
      html_dl: true
```

app/views/records/edit_fields/_subject.html.erb
```diff
@@ -0,0 +1,43 @@
+<%# OVERRIDE: Hyrax v5.0.0@_subject.html.erb - Use flexible metadata controlled vocabulary %>
+<% record = f.object.model %>
+<% controlled_vocab_config = controlled_vocabulary_options_for(key, record.class) %>
+
+<% if controlled_vocab_config.present? %>
```
ShanaLMoore: check for duplication in _default partial and render that instead of this block

app/views/records/edit_fields/_default.html.erb
ShanaLMoore: We will proceed by creating override partials for the hyrax authorities. form_helper_behavior also needs to be updated to support this change.

ShanaLMoore:
Things to consider: would we want a flag to allow users to free type?

app/views/records/edit_fields/_default.html.erb
ShanaLMoore: done ✅

config/initializers/questioning_authority.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+    
+    # Fallback to environment variables if account settings aren't available
+    if Qa::Authorities::Discogs::GenericAuthority.discogs_key.blank?
+      env_key = ENV.fetch('HYKU_DISCOGS_KEY', ENV.fetch('DISCOGS_KEY', nil))
+      env_secret = ENV.fetch('HYKU_DISCOGS_SECRET', ENV.fetch('DISCOGS_SECRET', nil))
+      
+      if env_key.present? && env_secret.present?
+        Qa::Authorities::Discogs::GenericAuthority.discogs_key = env_key
```
ShanaLMoore: we don't need this anymore. removed ability to set key and secret since it's deprecated for newer discogs

config/initializers/hyrax_controlled_vocabularies.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
```
ShanaLMoore, line 1: this will need to be revisited. eventually we want a ui where users will be allowed to create local vocab and this won't work. 

This initializer came from a refactor to resolve a rubocop error. Metrics/ModuleLength for form_helper_behavior. 

I have an idea but will put it in a different branch.

lib/tasks/discogs_setup.rake
```diff
@@ -0,0 +1,142 @@
...
+      puts "   Run: bundle exec rake discogs:setup"
+      return
+    end
+
+    # Test with a tenant that has a token
+    test_account = nil
+    Account.find_each do |account|
+      Account.switch!(account)
+      site = Site.instance
+      if site.respond_to?(:discogs_user_token) && site.discogs_user_token.present?
```
ShanaLMoore:
```suggestion
      if account.respond_to?(:discogs_user_token) && account.discogs_user_token.present?
```

ShanaLMoore:
Some controlled vocab services in Hyrax are classes, others are modules - why? Should they be normalized? cc @orangewolf

config/initializers/hyrax_controlled_vocabularies.rb
ShanaLMoore: TODO (in a later ticket): get DOT code from Rob. Checkout how that backend works and tie it in with this to may everything dynamic.

app/views/records/edit_fields/_default.html.erb
```diff
@@ -4,16 +4,84 @@
 <% return if f.object.try(:hidden?, key) %>
```
orangewolf, line 5: there are a lot of logical branches in this file. might be worth breaking it up in to multiple view partials or using some helpers

config/initializers/questioning_authority.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+# Configure Questioning Authority with dynamic credentials
+
+class DiscogsCredsConfig
```
orangewolf: one file per ruby class, one ruby class per file. pretty please =-)

ShanaLMoore: extracted to its own file ✅

app/views/records/edit_fields/_default.html.erb
ShanaLMoore: refactored ✅

PR #2617, opened 2025-07-16
I29 m3 validator

kirkkwang:
# Summary

Previously, invalid profiles could break the app because it would allow the user to save bad profiles. Users were unable to resolve the issues themselves. 

This validator is required for the hyku v6.2.0 release. 

## Ticket Number
- https://github.com/notch8/amigos/issues/29

# Screenshots / Video

image

# Expected Behavior 

Validation Rules include: 
- [x] - validate schema (houndstooth validator)
- [x] - required properties exist
  - [x] available_on/class
  - [x] range
  - [x] display_label/default 
- [ ] - check for existences of local vocab if one is being referred (will do after controlled vocab pr gets merged) 
  - [ ] https://github.com/samvera/hyku/pull/2608
  - [ ] https://github.com/notch8/hykuup_knapsack/issues/421
- [x] - check that the backend supports all the models listed in available_on.classes
- [x] - title cardinality - if no maximum caused issues
- [x] label for FileSet class to allow filesets to work
- [x] core metadata (from Hyrax) is required
- [x] prevent profile from saving (thus crashing the app) when bad data exists

# NOTES

We considered turning houndstooth into a gem, however we decided to hold off for now. There's a ticket to revisit this topic at a later date: 
- https://github.com/notch8/hykuup_knapsack/issues/409
- samvera community convo is happening [here](https://samvera.slack.com/archives/C0X9STKSP/p1752619285458939)

### DIFF
--- a/config/metadata_profiles/m3_profile.yaml
+++ b/config/metadata_profiles/m3_profile.yaml
@@ -19,28 +19,13 @@ classes:
     display_label: AdminSetResource
   CollectionResource:
     display_label: pcdmcollection
-  Hyrax::AdministrativeSet:
-    display_label: Administrative Set
   Hyrax::FileSet:
     display_label: FileSet
-  Hyrax::PcdmCollection:
-    display_label: PcdmCollection
 contexts:
   flexible_context:
     display_label: Flexible Metadata Example
   special_context:
     display_label: Special Context
-mappings:
-  blacklight:
-    name: Additional Blacklight Solr Mappings
-  metatags:
-    name: Metatags
-  mods_oai_pmh:
-    name: MODS OAI PMH
-  qualified_dc_pmh:
-    name: Qualified DC OAI PMH
-  simple_dc_pmh:
-    name: Simple DC OAI PMH
 properties:
   title:
     available_on:
@@ -52,8 +37,6 @@ properties:
         - ImageResource
         - OerResource
         - EtdResource
-    cardinality:
-      minimum: 1
     multi_value: true
     controlled_values:
       format: http://www.w3.org/2001/XMLSchema#string
@@ -65,7 +48,6 @@ properties:
         transcribe the title from the source itself.
     display_label:
       default: Title
-    index_documentation: displayable, searchable
     indexing:
       - title_sim
       - title_tesim
@@ -73,11 +55,6 @@ properties:
       required: true
       primary: true
       multi_value: true
-    mappings:
-      metatags: twitter:title, og:title
-      mods_oai_pmh: mods:titleInfo/mods:title
-      qualified_dc_pmh: dcterms:title
-      simple_dc_pmh: dc:title
     property_uri: http://purl.org/dc/terms/title
     range: http://www.w3.org/2001/XMLSchema#string
     requirement: required
@@ -98,9 +75,6 @@ properties:
         - ImageResource
         - OerResource
         - EtdResource
-    cardinality:
-      minimum: 0
-      maximum: 1
     multi_value: false
     display_label:
       default: Date Modified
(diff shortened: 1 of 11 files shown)


laritakr:
What happens if the schema has no title property? (Are there other required properties that must be included? i.e. the ones in CoreMetadata?) 

I'll have to review tomorrow, but just a few comments from my notes:

Required on a property to not break the app
- available_on
- display_label + default to not break the homepage
- form + any one sub key to not break the form (required, primary, or multi_value)
- range 
- label for FileSet class

ShanaLMoore:
> What happens if the schema has no title property? (Are there other required properties that must be included? i.e. the ones in CoreMetadata?)
> 
> I'll have to review tomorrow, but just a few comments from my notes:
> 
> Required on every property to not break the app
> 
>     * available_on
> 
>     * display_label + default to not break the homepage
> 
>     * form + any one sub key to not break the form (required, primary, multi_value... I think you can also put display: false to not show it on the form)
> 
>     * range
> 
>     * label for FileSet class to allow filesets to work

Thanks @laritakr We definitely forgot the label and form validations. And we were wondering - does it make sense to allow users to list any core metadata changes at all? Would it be ok if they don't include it in their profile (but it'll always be there on the backend), since it's "core" aka "not flexible"? Thoughts? For the meantime maybe we can include it in the validation but just thinking ... 🤔

ShanaLMoore:
@laritakr can you help me understand this statement more? 

form + any one sub key to not break the form (required, primary, multi_value... I think you can also put display: false to not show it on the form)

What exactly would the rule be? 

If form exists if must also have require, primary, multi_value...? 

I just imported a profile with form: and nothing else - nothing broke

config/metadata_profiles/m3_profile.yaml
ShanaLMoore, line 1: why is this deleted?

ShanaLMoore: it's not 🙃 I misread the diff

db/schema.rb
```diff
@@ -961,6 +961,20 @@
     t.datetime "updated_at"
   end
 
+  create_table "work_authorizations", force: :cascade do |t|
```
laritakr: Just curious where this came from?

laritakr (approved):

db/schema.rb
ShanaLMoore: thanks for the catch! we didn't make a migration so I have no idea why this is here 🤔 I'll revert it for now

laritakr (approved):

config/metadata_profiles/m3_profile.yaml
```diff
@@ -19,28 +19,13 @@ classes:
...
   Hyrax::FileSet:
     display_label: FileSet
-  Hyrax::PcdmCollection:
-    display_label: PcdmCollection
 contexts:
   flexible_context:
     display_label: Flexible Metadata Example
   special_context:
     display_label: Special Context
-mappings:
```
orangewolf, line 33: @ShanaLMoore @kirkkwang what are mappings removed here?

config/metadata_profiles/m3_profile.yaml
```diff
@@ -128,9 +102,6 @@ properties:
         - ImageResource
         - OerResource
         - EtdResource
-    cardinality:
```
orangewolf, line 131: @ShanaLMoore @kirkkwang what are mappings removed here? seems like all the cardinalities are gone?

ShanaLMoore: That’s correct. We removed them to avoid confusion because hyrax flexible metadata code wasn’t using it. 

We’ve made a ticket to address this and we’ll add them back then.

PR #2621, opened 2025-07-18
:bug: rescue google permission denied error

ShanaLMoore:
New tenants had analytics turned on by defaut when the global env variable was set. It should be set to false on a tenant level by default regardless if of the ENV var.

<details>

## BEFORE

Error present in demo.hykudemo.org reproduced locally:

image

## AFTER

image

</details>

Additionally, when a tenant added configs that the site admin hasn't added as a service account to their Google Analytics, a Google::Cloud::PermissionDeniedError would break the site. This override will gracefully handle it by displaying an error on the dashboard.

# BEFORE

Screenshot 2025-07-18 at 1 45 19 PM

# AFTER

image

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -75,6 +75,8 @@ module AccountSettings
 
   # rubocop:disable Metrics/BlockLength
   class_methods do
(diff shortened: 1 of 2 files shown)


ShanaLMoore:
image

PR #2622, opened 2025-07-21
defaults to en locale

ShanaLMoore:
Issue: https://github.com/notch8/hykuup_knapsack/issues/423

This change will ensure that if a locale is not specified in the URL, it will default to English.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -291,6 +291,7 @@ module Hyku
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+    config.i18n.default_locale = :en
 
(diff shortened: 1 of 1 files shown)


aprilrieger (approved):

config/application.rb
```diff
@@ -291,6 +291,7 @@ def self.work_types
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+    config.i18n.default_locale = :en
```
ShanaLMoore, line 294: In Rails 6 this setting is implied so this likely is not the issue afte rall 🤔

orangewolf: given that, do you want to merge this or skip it?

PR #2625, opened 2025-07-21
🎁 (validation): Ensure core metadata applies to all classes

ShanaLMoore:
Scenario: 

Core metadata is required on the profile, however we missed a validation requiring all classes to use it. I found this bug when testing WVU custom work type. I created a work type called "Test", made "Test" a value in available_on.class and was able to successfully upload the profile. However when I went to create a Test work the app crashed because Test was missing [core_metadata properties](https://github.com/samvera/hyrax/blob/5.0-flexible/config/metadata/core_metadata.yaml). 

This will hopefully prevent such issues from happening. 

image

## 🎁 (validation): Ensure core metadata applies to all classes

d78f297bdb911b62a27e5404a3a31f480e0ddb48

## ✅ Adds spec to test #validate_property_available_on

45423d6c9fb25dad6b70856a7ac0699705491a26

## update hyrax to make profile configurable

734849360d889102fdd2b2983030a08fef307870


ShanaLMoore:
TODO: write a spec and test in base hyku

PR #2627, opened 2025-07-22
I425 validation clean up 

ShanaLMoore:
Issue: 
- https://github.com/notch8/hykuup_knapsack/issues/425

This PR 
- implements cardinality

  <details> 

  
  image

  </details>

- switches multi_value => data_type (string/array)

  <details>
  
  
  image

  </details>

- removes required. it's based on cardinality now

  <details> 

  ## creator.cardinality.minimum = 0 && creator.form.required removed (required label disappears) 

  image

  ## creator.cardinality.minimum = 1 (required label appears and is enforced) 
  
  image

  </details>

- verifies classes are available

  **note: this validation is not enough. follow up [ticket/PR](https://github.com/notch8/hykuup_knapsack/issues/429) is required**

  <details> 
  
  image

  image

  </details>

- validates presence of creator

  <details> 
  
  ## creator property removes
  
  image

  ## creator property present but not on the work type GenericWorkResource

  image

  </details>

~TODO: update hyrax after merge~ ✅ 
- https://github.com/samvera/hyrax/pull/7145


app/services/hyrax/flexible_schema_validator_service.rb
```diff
@@ -14,31 +14,72 @@ class FlexibleSchemaValidatorService
...
     end
 
     private
 
+    # Validates that the profile's work types correspond to the site's
+    # enabled work types, adding human-readable error messages for any
+    # discrepancies.
+    #
+    # @return [void]
+    def validate_enabled_work_types
```
ShanaLMoore: This will soon be replaced by the work in
- https://github.com/notch8/hykuup_knapsack/issues/429
because it doesn't prevent the user from enabling the work type after disabling it in the profile. This leads to errors if the user then tries to create that work type, so a validation is not enough.

app/views/shared/_citations.html.erb
```diff
@@ -0,0 +1,57 @@
+<% # OVERRIDE Hyraxv5.0.5 to guard against nil keyword %>
```
laritakr: Is it possible that any of the others might be nil or even undefined with flexible metadata?

laritakr (dismissed):

app/views/shared/_citations.html.erb
ShanaLMoore: I don't think so. Even when rights statement isn't filled in it doesn't break like keyword did. 🤔

ShanaLMoore:
@laritakr sorry for the ping, i merged in that other PR and it reset your approval

app/views/shared/_citations.html.erb
ShanaLMoore: fix incoming

laritakr (approved):

PR #2630, opened 2025-07-25
Handles error gracefully when worktype isn't configured

ShanaLMoore:
# Summary

I thought about putting this in bulkrax directly, however I don't think it should know about the concept of flexible metadata in Hyrax. 

When a user includes a work type that is either not in their metadata profile, or it is but it's not enabled, Bulkrax will report a more helpful error than a NoMethodError. 

## Steps to reproduce: 

Upload a profile that doesn't have GenericWorkResource defined.
Import this zip with bulkrax's csv parser

[Archive(1).zip](https://github.com/user-attachments/files/21436270/Archive.1.zip)

# BEFORE: 

image

# AFTER: 

### when not defined in profile

<details>

image

</details>

### when on the profile but not enabled in available_works

<details> 
image

image

</details>

### DIFF
--- a/app/factories/bulkrax/object_factory_interface_decorator.rb
+++ b/app/factories/bulkrax/object_factory_interface_decorator.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# TODO: remove when issue https://github.com/notch8/hykuup_knapsack/issues/387 has been addressed and updated into this repo
+
 module Bulkrax
   module ObjectFactoryInterfaceDecorator
     # Perform a work-type sanity check before the factory creates/updates the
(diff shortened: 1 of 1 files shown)


app/factories/bulkrax/object_factory_interface_decorator.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
```
ShanaLMoore, line 2:
```suggestion

# TODO: remove when issue https://github.com/notch8/hykuup_knapsack/issues/387 has been addressed and updated into this repo

```

PR #2633, opened 2025-07-27
🎁 Prevents removal of classes that have already-persisted records

ShanaLMoore:
## Acceptance Criteria

- [ ] System prevents the removal of a work type from the flexible schema if there are already records of that type in the repository.

Screenshot 2025-07-27 at 08-41-39 Metadata Profiles

## Testing Instructions: 

Start with Hyku's default profile.
Create ImageResource works then try uploading the following profile

[no-image.yaml.zip](https://github.com/user-attachments/files/21454859/no-image.yaml.zip)

### DIFF
--- a/app/services/hyrax/flexible_schema_validator_service.rb
+++ b/app/services/hyrax/flexible_schema_validator_service.rb
@@ -2,6 +2,8 @@
 
 require 'json_schemer'
 require_relative 'core_metadata_validator'
+require_relative 'flexible_schema_validators/schema_validator'
+require_relative 'flexible_schema_validators/class_validator'
 
 module Hyrax
   class FlexibleSchemaValidatorService
@@ -35,9 +37,10 @@ module Hyrax
       validate_required_classes
       validate_class_availability
       validate_available_on_classes_defined
+      validate_existing_records_classes_defined
       validate_schema
       validate_label_prop
-      CoreMetadataValidator.new(profile: profile, errors: @errors).validate!
+      validate_core_metadata
     end
 
     # The default JSON schema used when no custom schema is provided.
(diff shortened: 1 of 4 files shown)


app/services/hyrax/flexible_schema_validator_service.rb
```diff
@@ -132,6 +133,43 @@ def validate_available_on_classes_defined
...
 
+    # Prevents removal of classes that have already-persisted records.
+    #
+    # If any works (or other Hyrax resources) exist whose `internal_resource`
+    # is *not* represented in the incoming profile, we add an error so the
+    # profile cannot be saved.
+    #
+    # @return [void]
+    def validate_existing_records_classes_defined
+      existing_classes =
```
laritakr: I'm still wondering if we should be using the resource table or a solr query to group by classes. Using the resource table means that we aren't supporting other adapters beyond postgres, which feels a bit more restrictive than we should be.

ShanaLMoore: oh true, good point I didn't think about that. 🤔

app/services/hyrax/existing_classes_query_service.rb
```diff
@@ -0,0 +1,60 @@
+# app/services/hyrax/existing_classes_query_service.rb
```
laritakr: I see that you defined a custom query `find_distinct_internal_resources` but not that it is actually used. 

We shouldn't have to check what the adapter is... each adapter should support the query.

ShanaLMoore: oh darn, maybe i forgot to commit something 👀 I was totally using it but you're right i don't see it here

ShanaLMoore:
> This doesn't appear to be fully implementing the custom query.

thanks for the catch 🙈 I think I lost a commit. will figure it out shortly.

sjproctor (approved):
Works locally for me!

ShanaLMoore:
this needs to change.

`Hyrax.query_service.count_all_of_model(model:)`

PR #2650, opened 2025-08-04
:bug: System-level models are always valid

ShanaLMoore:
Previously a CollectionResource record would refuse to be created and threw this error: `"Error: StandardError - Work type 'CollectionResource' is not enabled for this tenant. Please enable it via the 'Available Work Types' setting of the Admin Dashboard."` This is a bug! The system level models should always be valid.

Issue: 
- https://github.com/notch8/hykuup_knapsack/issues/436

### SAMPLE FILE
[Archive(1).zip](https://github.com/user-attachments/files/21583792/Archive.1.zip)

## BEFORE

image

## AFTER

image

### DIFF
--- a/app/factories/bulkrax/object_factory_interface_decorator.rb
+++ b/app/factories/bulkrax/object_factory_interface_decorator.rb
@@ -19,36 +19,53 @@ module Bulkrax
     # @param work_type [Class] the Valkyrie/ActiveFedora work class
     # @raise [StandardError] when the work type is invalid for the profile or tenant
     def validate_work_type!(work_type)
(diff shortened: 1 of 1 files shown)


ops/staging-deploy.tmpl.yaml
```diff
@@ -130,7 +130,7 @@ extraEnvVars: &envVars
   - name: HYRAX_FITS_PATH
     value: /app/fits/fits.sh
   - name: HYRAX_FLEXIBLE
-    value: "true"
+    value: "false"
```
ShanaLMoore, line 133: we want to test this in a non flexible metadata environment too cc @sjproctor

PR #2660, opened 2025-08-06
update-production-dc-file-links

aprilrieger:
Update the production dc file to have the updated image links for the latest strategy -- in line with the current dev dc file

This will need backporting into v6.1.0


kirkkwang (approved):

docker-compose.production.yml
```diff
@@ -177,6 +181,7 @@ services:
         solrcloud-upload-configset.sh /app/samvera/hyrax-webapp/solr/conf &&
         solrcloud-assign-configset.sh &&
         SOLR_COLLECTION_NAME=hydra-test solrcloud-assign-configset.sh &&
+        bundle &&
```
orangewolf: the built image for production should not need to bundle

orangewolf (changes requested):
we shouldn't need to call bundle on the production images because we're not mounting local code

PR #2669, opened 2025-08-14
🥳 hyrax base now has parity with the hyku base again!

orangewolf:
remove the hyku base to cut build times and complexity


aprilrieger (approved):
Thank you for getting this solution thought through and across the line. I really appreciate you pulling me in to explain the changes so I can help support the team.

PR #2683, opened 2025-08-25
Set environment default for Hyku APP_NAME

mark-dce:
* Set the APP_NAME default in the .env file
* Remove redundant defaults in docker-compose.yml

Additional fixes:
* Add quotes to interpolated values in .env file
* Remove redundant DB_HOST key in .env file

**ISSUE**
When APP_NAME is not set in the environment (either directly before launching the application, or via .env file), internal variables such as `$HYKU_ADMIN_HOST` are set to invalid values, causing admin routing to break.

I.E. when APP_NAME is not set, HYKU_ADMIN_HOST is set to `admin-.localhost.direct`, which results in an illegal hostname (because hostnames can not start or end with dashes).

**FIX**
Set the APP_NAME default in the .env file, which is loaded by `docker compose` and interpolated into `docker-compose.yml`. This ensures the samv value is used in both docker configuration and inside the Rails application.

@samvera/hyku-code-reviewers

### DIFF
--- a/.env
+++ b/.env
@@ -1,9 +1,9 @@
+APP_NAME=hyku
 CH12N_TOOL=fits_servlet
 CHROME_HOSTNAME=chrome
 COMPOSE_DOCKER_CLI_BUILD=1
 DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true
 DB_ADAPTER=postgresql
-DB_HOST=db
 DB_HOST=db
 DB_NAME=hyku
 DB_PASSWORD=DatabaseFTW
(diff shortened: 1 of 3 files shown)


.env
```diff
@@ -1,10 +1,10 @@
+APP_NAME=hyku
```
orangewolf: We set this to the current directory when using stack_car. This ensures if you have multiple Hyku instances running they all get their own name. That's why it wasn't set before. However, I did some experimentation and it looks like we can do it this way and have it work in both cases. Does that work for you?

```suggestion
APP_NAME=${APP_NAME:-hyku}
```

orangewolf (changes requested):

.env
mark-dce: Totally - let me try it in my local environment and I'll update the PR if it works.

mark-dce: @orangewolf I'm not sure I understand the functional difference between the the original and your proposed change. If APP_NAME is set in a parent process (e.g my shell or stack_car), won't that take precedence over the value set in .env?

Here's my test:
If I run with the initial .env formulation, e.g. `APP_NAME=hyku`
```
APP_NAME=my-dir docker compose up
```
The app is accessible at https://admin-my-dir.localhost.direct/ and I get
```
irb(main):001> ENV['HYKU_ADMIN_HOST']
=> "admin-my-dir.localhost.direct"
```

if I run with the proposed change, e.g. `APP_NAME=${APP_NAME:-hyku}`
```
APP_NAME=my-dir docker compose up
```
The app is accessible at https://admin-my-dir.localhost.direct/ and I get
```
irb(main):001> ENV['HYKU_ADMIN_HOST']
=> "admin-my-dir.localhost.direct"
```

They look the same to me, so I'd like the simpler formulation in the .env file if possible. But if I'm missing something about how stack_car interacts with the stack overall, I'm happy to put the change in.

PR #2685, opened 2025-08-26
Fix analytics tenant data spillover with GA4 filtering

ShanaLMoore:
Add tenant_id filtering to all Hyrax analytics calls to prevent cross-tenant data leakage in Google Analytics dashboards.

Issue: 
- https://github.com/notch8/palni_palci_knapsack/issues/314
- https://github.com/notch8/adventist_knapsack/issues/557

### DIFF
--- a/app/assets/javascripts/hyku/tenant_analytics_events.js
+++ b/app/assets/javascripts/hyku/tenant_analytics_events.js
@@ -1,0 +1,84 @@
+// OVERRIDE Hyrax v5.0.0 analytics events to include tenant identifier for multi-tenant analytics isolation
+
+// Override the analytics event tracking to include tenant ID
+function trackAnalyticsEventsWithTenant(provider) {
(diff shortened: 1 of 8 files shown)


app/services/hyrax/analytics/ga4/events_decorator.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
ShanaLMoore, line 1: can this be contributed back to hyrax?

app/services/hyrax/analytics/ga4_decorator.rb
```diff
@@ -0,0 +1,315 @@
+# frozen_string_literal: true
+
+begin
```
ShanaLMoore, line 3: This decorator is HUGE. How can we clean this up? What can be contributed back to Hyrax?

PR #2688, opened 2025-08-27
Streamline Getting Started docs

mark-dce:
The goal of this change is to provide a clear happy path for developers to get a Hyku development environment running using Docker.

This change intentionally moves or removes advanced options that are not essential for setting up a basic sandbox environment.

Changes
* Remove references to production docker-compose setup in development section
* Remove obsolete instructions for fixing docker groups
* Update release tag for checkout to v6.2.0
* Emphasize Stack Car as happy path for DNS/TLS
* Moved all optional setup and config information to it's own section
* Added initial troubleshooting section
* [Temporary] Remove testing instructions since they fail

TODO: diagnose why local testing is not working, fix, and add appropriate instructions.

@samvera/hyku-code-reviewers

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,96 +1,119 @@
 # Getting Started
 
-- [Docker (Recommended)](#docker)
-- [Locally without Docker](#locally-without-docker)
+- [Docker Development Environment (Start Here)](#docker-development-environment)
 - [Kubernetes](#kubernetes)
 - [AWS](#aws)
+- [Troubleshooting](#troubleshooting)
 
-## Docker
-
-We distribute two `docker-compose.yml` configuration files.  The first is set up for development / running the specs. The other, `docker-compose.production.yml` is for running the Hyku stack in a production setting.
-
-*Note: You may need to add your user to the "docker" group:*
-
-```bash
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -1,96 +1,119 @@
...
 
     ```bash
     gem install stack_car
-    sc proxy cert # Only need this once per stack_car version
+    sc proxy cert # Only need this once per stack_car version, requires passwords
     sc proxy up
     ```
 
-    #### Running Without Proxy
+3) **Build the Docker images:**
```
orangewolf, line 38: this is no longer needed (any maybe never was)

orangewolf (approved):
you might want to remove the seperate build step, since `docker compose up` will just build the ones that need building. otherwise this looks good to me

docs/getting-started.md
mark-dce: Agreed. I hadn't realized that yet when I was doing this PR. I'll pull it out in my next round of updated (when I put testing back in).

mark-dce:
@ShanaLMoore - no worries. I knew I could poke you or Rob, but I wasn't blocked.

PR #2693, opened 2025-09-02
Update the Hyrax Chart

orangewolf:


bin/helm_deploy
```diff
@@ -18,7 +18,7 @@ fi
 release_name="${1}"
 namespace="${2}"
 
-CHART_VERSION="${CHART_VERSION:-3.5.1}"
+CHART_VERSION="${CHART_VERSION:-3.7.0}"
```
ShanaLMoore: I tried a [deploy](https://github.com/samvera/hyku/actions/runs/17417769877). it failed.

 
Image

should we try 3.6.0 instead as that's the highest version listed in Hyrax [packages](https://github.com/orgs/samvera/packages/container/package/charts%2Fhyrax)? Or create one for [3.7.0](https://github.com/samvera/hyrax/blob/5.0-flexible/chart/hyrax/Chart.yaml#L5) as defined on the 5.0-flexible branch? @orangewolf 

Image

PR #2694, opened 2025-09-03
[i494] - Consolidate analytics concept

ShanaLMoore:
# Summary

Consolidated analytics settings by removing the separate `analytics_reporting` setting and using a single `analytics` checkbox to control both user interaction tracking and analytics dashboard reporting. This change simplifies the analytics configuration while ensuring new tenants start with blank GA4 fields and only show analytics UI when explicitly configured.

Note: The backend (Hyrax) still needs config.analytics_reporting set. This work still configures it, but the user no longer needs to be concerned about it on the frontend.

This effort is in attempt to streamline the concept and help prevent user error/confusion that we see with the current implementation. A separate ticket has been created to improve error handling for when the configuration values are incorrect. 

## Ticket Number

- https://github.com/notch8/palni_palci_knapsack/issues/314
- https://github.com/notch8/hykuup_knapsack/issues/494

# Screenshots / Video

<details>
<summary>Before: Separate analytics settings with confusing ENV fallback behavior</summary>

- Two separate checkboxes: `analytics` and `analytics_reporting`
- New tenants showed global ENV values in GA4 fields which confused the user
- Analytics sidebar drop down appeared if global env vars was set 
- The related ga4 fields were separate on the form from analytics boxes (towards the bottom)

image

</details>

<details>
<summary>After: Consolidated analytics with tenant-specific configuration</summary>

- Single `analytics` checkbox controls all analytics functionality
- New tenants have blank GA4 fields by default, even if the global env vars are set
- Analytics sidebar dropdown only appears when tenant configures their own GA4 settings
- Clean, grouped GA4 configuration section with clear setup instructions

Zight Recording 2025-09-04 at 04 39 05 PM

Zight Recording 2025-09-04 at 04 41 24 PM

### Tenant not configured

image

### Tenant configured 

Screenshot 2025-09-04 at 16-42-10 Edit Account __ Hyku

</details>

**Update**: Analytics can be configured with Matomo so the configuration section needed to be more generalized

When missing global settings. 
image

When global settings are present:

image

When configured correctly Analytics appears on the show pages, and the sidebar.

image

# Expected Behavior

## For New Tenants:
- [ ] Analytics checkbox is unchecked by default
- [ ] Google Analytics ID and Property ID fields are blank (not showing ENV values)
- [ ] Analytics dropdown is hidden in sidebar
- [ ] Analytics buttons are hidden on work show pages

## For Tenants with GA4 Configured:
- [ ] Analytics dropdown appears in sidebar under "ACTIVITY"
- [ ] Analytics reports (Works Report, Collections Report) are accessible
- [ ] Analytics buttons appear on work show pages
- [ ] Analytics functionality works with tenant-specific GA4 settings

## For All Tenants:
- [ ] Single `analytics` checkbox controls both tracking and reporting
- [ ] GA4 fields are grouped in dedicated "Google Analytics 4 Configuration" section
- [ ] Clear hint text with GA4 setup instructions and links
- [ ] No confusing technical notes about global environment settings

### DIFF
--- a/config/locales/simple_form.de.yml
+++ b/config/locales/simple_form.de.yml
@@ -8,7 +8,7 @@ de:
       account:
         admin_emails: Geben Sie jeweils eine E-Mail-Adresse ein
         allow_signup: Benutzern erlauben, sich für Ihr Repository anzumelden
-        analytics: Aktivieren oder deaktivieren Sie Google Analytics 4, um Benutzerinteraktionen, einschließlich Klicks, zu verfolgen.
+        analytics: Aktivieren oder deaktivieren Sie Analysen, um Benutzerinteraktionen zu verfolgen und Analysedaten im Dashboard anzuzeigen.
 
         cache_api: Cache für API-Endpunkte aktivieren. Experimentell
         contact_email: Die E-Mail-Adresse, von der Systembenachrichtigungen gesendet werden. Eine zusätzliche Konfiguration ist erforderlich, um eine Adresse von anderen Domänen als der Domäne der Website hinzuzufügen
@@ -25,7 +25,8 @@ de:
         file_acl: Deaktivieren Sie dies, wenn Sie ein Dateisystem wie Samba oder NFS verwenden, das das Festlegen von Zugriffssteuerungslisten nicht unterstützt
         file_size_limit: Dies sollte mindestens auf 536870912000 eingestellt sein
         geonames_username: Registrieren Sie sich unter http://www.geonames.org/manageaccount
-        google_analytics_id: Die ID Ihres Google Analytics-Kontos
+        google_analytics_id: 'Mess-ID (z.B. G-XXXXXXXXXX) von GA4 Admin → Eigenschaft → Datenstreams → Web-Stream. <a href="https://samvera.atlassian.net/wiki/spaces/hyku/pages/3185147970/Google+Analytics+4+GA4+Support#Creating-a-New-Google-Analytics-Account" target="_blank">Einrichtungsanleitung</a>'
+        google_analytics_property_id: 'Numerische Eigenschafts-ID (z.B. 123456789) von GA4 Admin → Eigenschaftsdetails (obere rechte Ecke). <a href="https://samvera.atlassian.net/wiki/spaces/hyku/pages/3185147970/Google+Analytics+4+GA4+Support#Creating-a-New-Google-Analytics-Account" target="_blank">Einrichtungsanleitung</a>'
         gtm_id: Die ID Ihres Google Tag Manager-Kontos
         hidden_index_fields: Komma-getrennt. Snake Case. Solr-Suffixe ausschließen (z. B. „title,rights_statement“)
         is_public: Können Benutzer Ihre Seite auf der Startseite entdecken oder auf Ihre Seiten zugreifen, ohne einen speziellen Benutzernamen / Passwort?
--- a/config/locales/simple_form.es.yml
+++ b/config/locales/simple_form.es.yml
@@ -8,7 +8,7 @@ es:
       account:
         admin_emails: Introduzca una dirección de correo electrónico a la vez
         allow_signup: Permitir a los usuarios registrarse en su repositorio
-        analytics: Habilite o deshabilite Google Analytics 4 para realizar un seguimiento de las interacciones del usuario, incluidos los clics.
+        analytics: Habilite o deshabilite las analíticas para rastrear las interacciones del usuario y mostrar datos analíticos en el panel de control.
 
         cache_api: Activa la caché para puntos finales de API. Experimental
         contact_email: La dirección de correo electrónico desde la que se enviarán las notificaciones del sistema. Se requiere una configuración adicional para agregar una dirección de dominios distintos al dominio del sitio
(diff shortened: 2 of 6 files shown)


app/views/proprietor/accounts/edit.html.erb
```diff
@@ -34,7 +34,40 @@
           </div>
 
           <h3><%= t(".account_settings") %></h3>
-          <% @account.live_settings.each do |key, value| %>
+
```
ShanaLMoore, line 37: Can this be a partial to be shared with admin/account/edit.html.erb?

sjproctor (approved):

app/models/concerns/account_settings.rb
```diff
@@ -138,8 +139,49 @@ def live_settings
...
+
+  def configure_hyrax_analytics_settings(config)
+    # Global analytics work for all tenants when ENV is set, regardless of tenant-specific settings
+    # But tenant-specific settings can override global settings
+    if analytics_functionally_available?
+      config.analytics = true
+      config.analytics_reporting = true
+    else
+      config.analytics = false
+      config.analytics_reporting = false
```
orangewolf: I thought analytics_reporting was being removed in this PR?

app/models/concerns/account_settings.rb
```diff
@@ -268,20 +287,17 @@ def configure_ssl
...
 
-  def analytics_credentials_present?
-    google_analytics_id.present? &&
-      google_analytics_property_id.present? &&
-      (ENV.fetch('GOOGLE_ACCOUNT_JSON', '').present? || ENV.fetch('GOOGLE_ACCOUNT_JSON_PATH', '').present?)
-  end
-
   def reload_hyrax_analytics
+    analytics_id = google_analytics_id.presence || ENV.fetch('GOOGLE_ANALYTICS_ID', '')
+    property_id = google_analytics_property_id.presence || ENV.fetch('GOOGLE_ANALYTICS_PROPERTY_ID', '')
```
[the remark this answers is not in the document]
orangewolf: I'm with copilot on this one - isn't setting the Hyrax analytics id and property id count productive here?

app/views/admin/accounts/edit.html.erb
```diff
@@ -21,7 +21,15 @@
             <%= f.text_field :tenant, class: 'form-control', readonly: @account.persisted? %>
           </div>
           <% is_superadmin = current_ability.superadmin? %>
-          <% current_account.public_settings(is_superadmin: is_superadmin).each do |key, value| %>
+          <% settings = current_account.public_settings(is_superadmin: is_superadmin) %>
+          
+          <!-- Google Analytics Settings Section -->
+          <% if settings.key?(:analytics) || settings.key?(:google_analytics_id) || settings.key?(:google_analytics_property_id) %>
```
orangewolf, line 27: what happens if its motombo and not google analytics?

app/views/shared/_google_analytics_settings.html.erb
```diff
@@ -0,0 +1,92 @@
...
+        <div class="col-md-6">
+          <%= render 'shared/settings', f: f, key: :google_analytics_property_id, value: settings[:google_analytics_property_id] %>
+        </div>
+      <% end %>
+    </div>
+  </div>
+</div>
+
+<script>
+document.addEventListener('DOMContentLoaded', function() {
```
orangewolf: live in the view is not where javascript goes =-p

orangewolf (changes requested):
the big old JS block in a view file needs to be cleaned up. the rest of my comments are just questions. I'm curious that I don't see anything about rendering the analytics id on the page... is that just because that part was working properly?

app/models/concerns/account_settings.rb
ShanaLMoore: oops! good catch. Yes this can be removed.

app/views/admin/accounts/edit.html.erb
ShanaLMoore: aw dang, I didn't think of that 😓 🤔 👀 

Update: Currently users don't have a front end to set matomo values so what would I render? I could update up the help text to guide users towards that set up, or re title it to be more generic "analytics settings"? What would you expect to happen if its matomo?

app/views/shared/_google_analytics_settings.html.erb
ShanaLMoore: totally agree, I was just trying to get it work brute force at the time 🙃 I'll move it in my clean up iteration.

app/views/admin/accounts/edit.html.erb
ShanaLMoore: I updated the logic to consider matomo for analytics, and created another ticket for the rest of its implementation to not add scope to this one: 

https://github.com/samvera/hyku/issues/2699

The config's title is more generalized now

image

app/models/concerns/account_settings.rb
ShanaLMoore: Actually Rob, I second guessed myself. This was intentional because Hyrax checks to this in their controllers. https://github.com/samvera/hyrax/blob/main/app/controllers/hyrax/admin/analytics/collection_reports_controller.rb#L8

If I remove these, the analytics no longer displays on the work show and report pages. I need to replace them, but the goal was for the user to not have to know the difference conceptually.

PR #2710, opened 2025-09-10
Update authorities to load multiple paths

laritakr:
## Summary

Handles multiple authority location branches, which allows a knapsack to set a branch with authority files to overwrite the content in any of hyku's authority files.


ShanaLMoore (approved):
Thanks for the catch and fix LaRita! 👏🏿

PR #2730, opened 2025-09-23
loading message about account switching

orangewolf:
I just can't take it any more... I keep forgetting.

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+Rails.application.console do
+  case Account.count
+  when 0
+    puts "***** No accounts, found, please run the seeds *****"
+  when 1
+    switch!(Account.first)
+    puts "***** Only one account found, switching to it automatically *****"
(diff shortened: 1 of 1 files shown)


config/initializers/console.rb
```diff
@@ -0,0 +1,13 @@
...
+
+Rails.application.console do
+  case Account.count
+  when 0
+    Rails.logger.error "***** No accounts, found, please run the seeds *****"
+  when 1
+    switch!(Account.first)
+    Rails.logger.error "***** Only one account found, switching to it automatically *****"
+  else
+    Rails.logger.error "***** Multiple accounts found, dont' forget to switch in to one with switch!(ACCOUNT_NAME) or switch!(ACCOUNT_CNAME) *****"
```
kirkkwang:
```suggestion
    Rails.logger.error "***** Multiple accounts found, don't forget to switch into one with switch!(ACCOUNT_NAME) or switch!(ACCOUNT_CNAME) *****"
```

PR #2739, opened 2025-09-29
Ensure bulkrax_identifier is present and findable on FileSets

maxkadel:
Refs https://github.com/notch8/hykuup_knapsack/issues/454

Ensure bulkrax_identifier is present and findable on FileSets

- Uses consistent Solr field value for bulkrax_identifier `bulkrax_identifier_tesim` rather than `bulkrax_identifier_tesi`
- Makes bulkrax_identifier available on FileSet via configuration

@samvera/hyku-code-reviewers

### DIFF
--- a/config/metadata_profiles/m3_profile.yaml
+++ b/config/metadata_profiles/m3_profile.yaml
@@ -1574,6 +1574,7 @@ properties:
     available_on:
       class:
         - Hyrax::FileSet
+        - FileSet
         - CollectionResource
         - GenericWorkResource
         - ImageResource
@@ -1591,7 +1592,7 @@ properties:
       default: Bulkrax Identifier
     index_documentation: displayable, searchable
     indexing:
-      - "bulkrax_identifier_tesi"
+      - "bulkrax_identifier_tesim"
       - "bulkrax_identifier_ssi"
     form:
       display: false


laritakr (approved):

config/metadata_profiles/m3_profile.yaml
```diff
@@ -1591,7 +1592,7 @@ properties:
       default: Bulkrax Identifier
     index_documentation: displayable, searchable
     indexing:
-      - "bulkrax_identifier_tesi"
+      - "bulkrax_identifier_tesim"
```
orangewolf, line 1595: is it intentional for bulkdrax_identifier to be indexed as multiple instead of single?

laritakr: Adding my 2 cents: I approved the PR because I know that's what was used before the flexible work. I assume that's why Max changed it this way... to be consistent with prior work.

maxkadel: Note that the next line is `bulkrax_identifier_ssi`, so we're indexing it as both. I was basing this change in part on this comment: https://github.com/samvera/hyku/blob/e87bca53f04ab949530359b35eb342c36f2da97c/config/metadata/bulkrax_metadata.yaml#L12-L17 

Everywhere else I see it, it's `_tesim`, so I was just making it consistent.

PR #2744, opened 2025-10-01
:bug: Fix validation to handle inconsistent naming conventions

ShanaLMoore:
# Summary

Fix validation to handle inconsistent M3 profile class naming conventions. The validator currently only checks for `"#{work_type}Resource"` variants, missing existing records when profiles use direct model class names (e.g., `MobiusWork` instead of `MobiusWorkResource`).

## Ticket Number

Discovered in: 
- https://github.com/notch8/hykuup_knapsack/issues/525

#### BEFORE

From hykuup knapsack I am in a tenant that has a MobiusWork type created. I uploaded a profile that did not have MobiusWork defined. It should've failed validation but it passed. 

**Root Cause**
The potential_existing_classes method assumes all work types have a corresponding "#{work_type}Resource" class:
end

```
# Current (buggy) logic
Hyrax.config.registered_curation_concern_types.each do |concern_type|
  classes << "#{concern_type}Resource"  # Assumes this always exists!
end
```

This assumption breaks when:

- Work types don't follow the "Resource" suffix convention
- Profiles use actual model class names instead of profile class names
- Future developers create work types without "Resource" variants

image

#### AFTER

image

## Expected Behavior

Validation should detect existing records regardless of whether profiles use:
- `WorkTypeResource` naming (e.g., `ImageResource`)  
- `WorkType` naming (e.g., `MobiusWork`)

## Sample Files

**Bug scenario:**
```yaml
# Profile excludes MobiusWork but database has MobiusWork records
classes:
  GenericWorkResource:
    display_label: Generic Work
  # MobiusWork missing - should fail but doesn't
```

**Result:** Validation incorrectly passes, allowing silent data loss.

**Fixed behavior:**
```yaml
# Same profile
classes:
  GenericWorkResource:
    display_label: Generic Work
  # MobiusWork still missing
```

**Result:** Validation correctly fails with "Classes with existing records cannot be removed from the profile: MobiusWork."

## Notes

- Currently works in core Hyku because each work type has both `work_type.rb` and `work_type_resource.rb`
- This fix future-proofs validation for any naming convention
- No breaking changes - existing functionality unchanged
- Improves robustness of profile validation system

### DIFF
--- a/app/services/hyrax/flexible_schema_validators/class_validator.rb
+++ b/app/services/hyrax/flexible_schema_validators/class_validator.rb
@@ -56,18 +56,11 @@ module Hyrax
         classes_with_records = []
         checked_models = Set.new
 
-        classes_to_check.reject! do |class_name|
-          counterpart = if class_name.end_with?('Resource')
-                          class_name.chomp('Resource')
-                        else
-                          "#{class_name}Resource"
-                        end
-          profile_classes.include?(counterpart)
-        end
-
         classes_to_check.each do |class_name|
           model_class = resolve_model_class(class_name)
           next unless model_class
+
+          next if profile_classes.include?(model_class.to_s)
 
           model_identifier = model_class.to_s
           next if checked_models.include?(model_identifier)
--- a/app/services/hyrax/flexible_schema_validators/class_validator.rb
+++ b/app/services/hyrax/flexible_schema_validators/class_validator.rb
@@ -53,14 +53,21 @@ module Hyrax
       def validate_existing_records!
         classes_to_check = potential_existing_classes - @profile['classes'].keys
         classes_with_records = []
+        checked_models = Set.new
 
         classes_to_check.each do |class_name|
           model_class = resolve_model_class(class_name)
           next unless model_class
 
+          model_identifier = model_class.to_s
+          next if checked_models.include?(model_identifier)
+
           begin
             count = Hyrax.query_service.count_all_of_model(model: model_class)
-            classes_with_records << class_name if count.positive?
+            if count.positive?
+              classes_with_records << model_identifier
+              checked_models.add(model_identifier)
+            end
           rescue StandardError => e
             Rails.logger.error "Error checking records for #{class_name}: #{e.message}"
           end
@@ -79,7 +86,8 @@ module Hyrax
         classes = @required_classes.dup
 
         Hyrax.config.registered_curation_concern_types.each do |concern_type|
-          classes << "#{concern_type.capitalize}Resource"
+          classes << "#{concern_type}Resource"
+          classes << concern_type
         end
 
         classes.uniq
--- a/app/services/hyrax/flexible_schema_validators/class_validator.rb
+++ b/app/services/hyrax/flexible_schema_validators/class_validator.rb
@@ -4,14 +4,16 @@ module Hyrax
   module FlexibleSchemaValidators
     # Handles class-related validations for flexible metadata profiles
     class ClassValidator
+      # @param profile [Hash] M3 profile data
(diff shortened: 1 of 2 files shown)


app/services/hyrax/flexible_schema_validators/class_validator.rb
```diff
@@ -79,7 +86,8 @@ def potential_existing_classes
         classes = @required_classes.dup
 
         Hyrax.config.registered_curation_concern_types.each do |concern_type|
-          classes << "#{concern_type.capitalize}Resource"
+          classes << "#{concern_type}Resource"
```
laritakr: Still wondering about this a bit... Does this mean that MobiusWorkResource would be considered valid as well as MobiusWork? Do we need to verify that the classes are actually valid, or do we do that somewhere else?

ShanaLMoore: it happens elsewhere. 

We now add BOTH "#{concern_type}Resource" AND concern_type to the list
So for MobiusWork, we check both "MobiusWorkResource" and "MobiusWork"

The #resolve_model_class method tries to constantize each class name
- If MobiusWorkResource doesn't exist as a Ruby class, it returns nil
- If MobiusWork exists, it returns the actual MobiusWork class

Example flow:
- Check MobiusWorkResource → resolve_model_class → nil (doesn't exist) → skip
- Check MobiusWork → resolve_model_class → MobiusWork class → check for existing records ✅

ShanaLMoore:
The more I work on this, the more I start to think that this should be a knapsack override.

app/services/hyrax/flexible_schema_validators/class_validator.rb
```diff
@@ -49,61 +50,69 @@ def validate_references!
...
+          invalid_classes << klass
+          return
+        end
 
-      # Returns classes that could potentially have existing records
-      # @return [Array<String>] Class names in profile format
-      def potential_existing_classes
-        classes = @required_classes.dup
+        begin
+          expected_valkyrie_class = Valkyrie.config.resource_class_resolver.call(base_class_name)
```
ShanaLMoore, line 84: cc @laritakr 🎉

app/services/hyrax/flexible_schema_validators/existing_records_validator.rb
```diff
@@ -0,0 +1,82 @@
...
+        return if classes_with_records.empty?
+
+        @errors << "Classes with existing records cannot be removed from the profile: #{classes_with_records.uniq.join(', ')}."
+      end
+
+      private
+
+      # Determines the counterpart model name (e.g., Image -> ImageResource).
+      def counterpart_for(model_identifier)
+        if model_identifier.end_with?('Resource')
```
laritakr: This should work, but I prefer testing if Wings is defined, and then using `Wings::ModelRegistry.lookup(work_type)` 

Generally because Hyrax::FileSet doesn't use the Resource suffix on the Active Fedora FileSet model, but file sets shouldn't be getting removed from the M3 profile anyway so it shouldn't matter.

laritakr (approved):
Good job! Just left one non-blocking comment.

app/services/hyrax/flexible_schema_validators/existing_records_validator.rb
ShanaLMoore: I appreciate your guidance. I admit I do not do a good job realizing/remembering such methods exist.

ShanaLMoore:
sorry to bug you again @laritakr I need another approval. I took your suggestion and it removed your previous one.

laritakr (approved):

PR #2747, opened 2025-10-04
Explicitly set HYRAX_FLEXIBLE environment variable to false

mark-dce:
**ISSUE**
Hyku is always running with Flexible Metdata enabled.

**DIAGNOSIS**
There are two issues occuring:

1. StackCar does not appear to pass host environment variables to containers, so there is no way to override the HYRAX_FLEXIBLE value set in the container `.env` file

2. The [`dotenv` gem](https://github.com/bkeepers/dotenv?tab=readme-ov-file#variable-substitution) does not support [Docker-style environments defaults](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax), so we are not getting the expected default values for HYRAX_FLEXIBLE

**EXAMPLE**
```
mark@Marks-MacBook-Pro pitt2025 % HYRAX_FLEXIBLE=ignored sc console
         run  dotenv -f .env.development,.env docker compose exec web bundle exec rails console  from "."
. . . (unrelated console output omitted . . .
Loading development environment (Rails 6.1.7.6)
irb(main):001> Hyrax.config.flexible?
=> true
irb(main):002> ENV['HYRAX_FLEXIBLE']
=> ":-false}"
irb(main):003> ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_FLEXIBLE', false))
=> true
irb(main):004>
```

**RESOLUTION**
Explicitly set the desired default in the `.env` file.

@samvera/hyku-code-reviewers


ShanaLMoore (approved):

docker-compose.yml
```diff
@@ -1,3 +1,8 @@
+# dotenv gem doesn't always do the right thing with blank env vars, so we make sure these get set to something
+x-app-env: &app-env
```
orangewolf, line 2: @mark-dce adding it to be interpreted by the docker-compose means we can still override the value. both:
`HYRAX_FLEXIBLE=true sc up` and 

```
sc sh
HYRAX_FLEXIBLE=true ./bin/rails c
```
will work properly with this addition.

orangewolf (approved):

docker-compose.yml
mark-dce: Thanks! I tested locally and this works perfectly 👍

PR #2771, opened 2025-11-08
Align all deploy values files with each other for friends deploy envs

aprilrieger:
Persists what is currently deployed to friends cluster successfully:
- Align all deploy values files with each other for friends deploy envs
- Added the secret env vars for the hyku-iiif deployment and added everything needed to deploy to friends cluster


maxkadel (approved):
Looks good, thanks!

PR #2777, opened 2025-11-11
💄 Uniform property rendering for all work types

sjproctor:
This commit standardizes the rendering of properties across the base Hyrax views as well as ETD and OER work types to ensure properties that to not have a view key in flexible metadata profiles are not rendered in the view.

Also fixes a bug where a logged out user got an error message for works that had admin_only properties.

Ref:
- https://github.com/notch8/palni_palci_knapsack/issues/499

@samvera/hyku-code-reviewers


laritakr (approved):
I'd like to consider a way to do this via flexible metadata if it's actually needed to be shown to editor rather than admin. But this is an adequate alternative for now, to ensure that all views have consistent behavior.

sjproctor:
Yeah, that's a good point LaRita. I'll made a note to add that to the discussion around a cleaner implementation of flexible metadata.

PR #2800, opened 2025-11-22
prep for 7.0 release with updated iiif_print and bulkrax versions

orangewolf:


laritakr:
I think we want another Bulkrax release. I'll do that Monday. Also curious if we're close to moving to use hyrax main?

Issue #2775, opened 2025-11-10
🐛 Label is rendering a key when there is no value


sjproctor:
## Story

- Hyku 6.2
- flex=on 
- Work show page

Label is rendering when there is no value. The value is `[""]` which is not failing the `.blank?` check.

## Acceptance Criteria
- [x] Metadata properties don't render when there is no value

<img width="1504" height="938" alt="Screenshot 2025-11-04 at 1 16 03 PM" src="https://github.com/user-attachments/assets/fd067025-ca12-4891-bfa3-72d90a4bfc37" />

PR #2837, opened 2025-12-03
🐛 Properties rendering correctly on the show page

sjproctor:
This commit ensures that flexible metadata properties only display on the show page when they contain specific values and are not empty.

The `label` property rendering on the show page was an incorrect data type in the metadata profile. 
- `data_type: string` results in a stored value of `[""]`
- `data_type: array` fixes this issue

🚧 Waiting until https://github.com/samvera/hyrax/pull/7279 gets merged into Hyrax and can be pulled into this branch.

Ref:
- https://github.com/notch8/hyku-community-issues/issues/9
- https://github.com/samvera/hyku/issues/2775

@samvera/hyku-code-reviewers

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,7 +49,7 @@ GIT
 
 GIT
   remote: https://github.com/samvera/hyrax.git
-  revision: 8abde49b1aacd17400e5bf0fd2717ab90d0ab7c9
+  revision: 6f851840b30656a2535955b7cb2fb8796689899b
   branch: main
   specs:
     hyrax (5.2.0)


app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -4,6 +4,9 @@
 <% if Hyrax.config.flexible? %>
   <% view_options_for(presenter).each do |field, options| %>
     <% view_options = conform_options(field, options) %>
+
+    <%# can add metadata property called admin_note that is always restricted to admins %>
```
laritakr: admin_note is restricted to any editor role, not just admin, so this is a bit misleading.

sjproctor: good point! Fixed!

laritakr (dismissed):

PR #2851, opened 2025-12-10
🎁 Identity Provider refactor and bug fix 

sjproctor:
### Overview

Refactors the identity provider form to improve maintainability, internationalization, and code organization. Moves all user-facing text to translation files and adds test coverage.

### Changes to the form (app/views/identity_providers/_form.html.erb)

Code organization:
- Extracts repeated variables at the top
- Extracts long URLs to variables for easier maintenance and readability
- All external links open in a new tab
- Ensures proper aria labels for all external links

UI improvements:
- Adds section headers for Documentation and Configurations sections
- Adds background to documentation section for better visual separation

Bug fixes:
- Fixes delete button bug: changes condition from IdentityProvider.count.nonzero? to identity_provider.new_record? to prevent delete button appearing on new forms when other providers exist and causing a error when clicked
- Updates turbolinks to turbo for Rails 7

Internationalization:
- Replaces all hardcoded strings with translation keys
- Uses translation keys for documentation links and aria labels
- Adds `_html` to all keys that contain html since Rails I18n automatically marks keys ending in `_html` as safe
- Normalizes all translation file by running: `i18n-tasks normalize`
- Adds new identity provider translations by running: `i18n-tasks translate-missing --from en es zh fr it de pt-BR`

### New test file (spec/views/identity_providers/_form.html.erb_spec.rb)

Test coverage:
- New form rendering with all required fields
- Validation error display
- Existing provider editing (saved values, callback URLs, metadata links)
- Logo image display
- Options JSON display
- Delete button only appears for persisted records
- Using mocks on the test to avoid database interaction

### New helpers file (app/helpers/identity_providers_helper.rb)

Extract variables and paths:
- `external_link_attrs` for opening documentation in a new tab safety/accessibility
- Documentation links
- Callback path
- Metadata path

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

Ref:
- https://github.com/notch8/hykuup_knapsack/issues/592

@samvera/hyku-code-reviewers

### DIFF
--- a/app/views/identity_providers/_form.html.erb
+++ b/app/views/identity_providers/_form.html.erb
@@ -1,31 +1,3 @@
-<%
-    external_link_attrs = {
-    target: '_blank',
-    rel: 'noopener noreferrer'
-  }
-
-  samvera_docs_url = 'https://samvera.atlassian.net/wiki/spaces/hyku/pages/3570663437/Identity+Provider+Single+Sign-On+SSO'
-  saml_docs_url = 'https://github.com/omniauth/omniauth-saml#idp-metadata'
-
-  adapter_documentation_links = [
-    {
-      name: t('hyku.identity_provider.documentation.link.saml'),
-      url: 'https://github.com/omniauth/omniauth-saml',
-      aria_label: t('hyku.identity_provider.documentation.aria_label.saml')
-    },
-    {
-      name: t('hyku.identity_provider.documentation.link.cas'),
-      url: 'https://github.com/dlindahl/omniauth-cas',
-      aria_label: t('hyku.identity_provider.documentation.aria_label.cas')
-    },
-    {
-      name: t('hyku.identity_provider.documentation.link.openid_connect'),
-      url: 'https://github.com/omniauth/omniauth_openid_connect',
-      aria_label: t('hyku.identity_provider.documentation.aria_label.openid_connect')
-    }
-  ]
-%>
-
 <%= simple_form_for(@identity_provider) do |f| %>
   <div class="card">
     <div class="card-body">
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
ShanaLMoore, line 1: Non blocking suggestion: We should probably run translations to update and support the other languages. 

ref: https://github.com/notch8/playbook/blob/main/samvera/hyrax/how_to/generate_translations.md

ShanaLMoore:
I love the visual difference you added for the documentation section.

spec/views/identity_providers/_form.html.erb_spec.rb
ShanaLMoore, line 1: 👏🏿

config/locales/en.yml
sjproctor: I made some updates to the Notch8 translation documentation based on my experience running the commands for this PR!

app/views/identity_providers/_form.html.erb
```diff
@@ -1,56 +1,159 @@
-<%= simple_form_for(@identity_provider) do |f| %>
+<%
+  identity_provider = @identity_provider
```
ShanaLMoore: One small nit: I think you can just use @identity_provider here instead of reassigning it to a local.

I also noticed the partial has quite a bit of Ruby setup at the top now (e.g., provider_collection, external_link_attrs, adapter_documentation_links, saml_callback_path, etc.). Since you already have pretty solid view specs covering the form behavior, would you be open to moving some of that setup into helpers or a presenter? That could make the template easier to read and keep the logic in a more testable place.

Overall this looks great though; thanks for tackling this! 🙏🏿

sjproctor: good feedback. I was able to extract logic into the controller and variables into a helper.

PR #2861, opened 2025-12-17
Pull Tracy's changes from his branch to 6.2-stable

kirkkwang:
Ref:
- https://github.com/samvera/hyku/compare/feature/hide-metadata-profiles-search-only

### DIFF
--- a/app/indexers/app_indexer.rb
+++ b/app/indexers/app_indexer.rb
@@ -25,6 +25,8 @@ class AppIndexer < Hyrax::WorkIndexer
       solr_doc['account_institution_name_ssim'] = Site.instance.institution_label
       solr_doc['all_text_tsimv'] = full_text(object.file_sets.first&.id)
       add_date(solr_doc)
+      # Add this guard for schema_version
+      solr_doc['schema_version_tesim'] = object.schema_version if object.respond_to?(:schema_version)
     end
   end
   # rubocop:enable Metrics/AbcSize
--- a/app/views/hyrax/dashboard/sidebar/_metadata.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_metadata.html.erb
@@ -1,0 +1,12 @@
+<%# menu item to be inserted when using flexible metadata %>
+<% if Hyrax.config.flexible? &&
+      current_ability.can?(:manage, Hyrax::FlexibleSchema) &&
+      !Site.account&.search_only? &&
+      defined?(metadata_profiles_path) %>
+  <li class="nav-item">
+    <%= menu.nav_link(metadata_profiles_path, class: "nav-link") do %>
+      <span class="fa fa-table" aria-hidden="true"></span>
+      <span class="sidebar-action-text"><%= t('hyrax.admin.sidebar.metadata_profiles') %></span>
+    <% end %>
+  </li>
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -316,6 +319,9 @@ en:
         roles:
           remove:
             confirmation: Are you sure you want to remove the role "%{role}" from the user "%{user}"?
+  dashboard:
+      sidebar:
+        metadata_profiles: "Metadata Profiles"
```
cjcolvar, line 324: It seems like this needs to be moved to a different place in the hierarchy to match the reference in the view partial: `hyrax.admin.sidebar.metadata_profiles`.

config/locales/en.yml
```diff
@@ -286,6 +286,9 @@ en:
             In order for Hyrax::Group and User records who have either the 'Work Editor' or 'Work Depositor' Role
             to have the correct permissions for Works, we grant them Workflow Roles for all AdminSets.
             NOTE: All AdminSets must have a permission template or this will fail. Run 'Create Admin Set Accesses' first."
+          clean_review_submissions_page:
+            label: Clean Review Submissions Page
+            description: "This job will remove the empty items from the Review Submissions page and fix the counts shown."
```
cjcolvar, line 291: Is this related to this PR or can it be removed and included in a different PR?

app/views/hyrax/dashboard/sidebar/_metadata.html.erb
```diff
@@ -0,0 +1,12 @@
+<%# menu item to be inserted when using flexible metadata %>
```
cjcolvar, line 1: Is this view partial rendered somewhere?  I'm not seeing it in `app/views/hyrax/dashboard/_sidebar.html.erb`

app/indexers/app_indexer.rb
```diff
@@ -25,6 +25,8 @@ def generate_solr_document
       solr_doc['account_institution_name_ssim'] = Site.instance.institution_label
       solr_doc['all_text_tsimv'] = full_text(object.file_sets.first&.id)
       add_date(solr_doc)
+      # Add this guard for schema_version
+      solr_doc['schema_version_tesim'] = object.schema_version if object.respond_to?(:schema_version)
```
cjcolvar, line 29: Where does this solr field end up getting used?  Could you please add a test to ensure this change is working as expected?

app/views/hyrax/dashboard/sidebar/_metadata.html.erb
laritakr: This comes in and is overriding the view in Hyrax 5.0-flexible branch.

app/indexers/app_indexer.rb
laritakr: This seems like it must have brought in changes from something else, as it is unrelated to this change.

config/locales/en.yml
laritakr: This came from a change that is in hyku main, and doesn't related to this PR.

config/locales/en.yml
laritakr: I agree with Chris's comment here.

PR #2871, opened 2026-01-08
fix-rspec-zeitwerk-frozen-array-errors

aprilrieger:
# Contribute RSpec Fixes Identified During Hyku Upgrade

## Summary

This pull request contributes fixes identified during an upgrade of the **Hyku version** used by the WVU project (`wvu_knapsack`).  
It resolves two compatibility issues that prevented the RSpec suite from running:

- Zeitwerk autoloading failure for `AdvancedQueryParserDecorator`
- `FrozenError` caused by ActiveFedora modifying frozen `autoload_paths`

---

## Context

The WVU project consumes **Hyku via a git submodule**. After upgrading the Hyku version and updating the CI build to run on the default **linux/amd64** platform, the RSpec suite failed before any examples executed.

These failures surfaced underlying compatibility issues that were previously masked. The fixes in this pull request were identified while investigating those CI failures and have been generalized so they can be safely contributed upstream, benefiting other Hyku-based applications.

---

## Problem

Prior to this change, the RSpec suite failed with **13 errors before any examples ran**, including:

- **Zeitwerk::NameError**  
  _Expected constant `BlacklightAdvancedSearch::AdvancedQueryParserDecorator` not defined_

- **FrozenError**  
  _ActiveFedora attempted to modify a frozen `autoload_paths` array_

---

### Issue 1: Zeitwerk constant definition

- `AdvancedQueryParserDecorator` was conditionally defined behind a version check
- Zeitwerk requires constants to be defined whenever the file is loaded
- The conditional definition prevented Zeitwerk from resolving the constant

### Issue 2: Frozen `autoload_paths` modification

- `autoload_paths` are frozen earlier during application initialization
- ActiveFedora’s railtie attempts to append `app/models/datastreams` to `autoload_paths`
- This results in:  
  `FrozenError: can't modify frozen Array`

---

### Fix 1: Always define the decorator module

- Define `AdvancedQueryParserDecorator` unconditionally
- Apply `prepend` only when the relevant version conditions are met
- Ensures Zeitwerk can always locate the constant

### Fix 2: Safely handle frozen `autoload_paths`

- Add `app/models/datastreams` early in `config/application.rb` via a `before_initialize` block
- Introduce a defensive initializer to handle edge cases where the array may already be frozen

---

## Changes Made

- **`lib/blacklight_advanced_search/advanced_query_parser_decorator.rb`**
  - Always define the decorator module
  - Conditionally apply `prepend`

- **`config/application.rb`**
  - Add `app/models/datastreams` early during initialization

- **`config/initializers/active_fedora_frozen_array_fix.rb`**
  - New initializer to safely handle frozen `autoload_paths`

---

## Impact

- No breaking changes
- Backward compatible
- Restores full RSpec suite execution
- Improves compatibility for Hyku-based applications in CI environments

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -153,16 +153,6 @@ module Hyku
     config.autoload_lib(ignore: %w[assets tasks middleware rubocop])
     config.add_autoload_paths_to_load_path = true
 
-    # Add datastreams path early to prevent ActiveFedora from modifying frozen autoload_paths in Rails 7.2
-    config.before_initialize do
-      datastreams_path = Rails.root.join('app', 'models', 'datastreams').to_s
-      unless config.autoload_paths.include?(datastreams_path)
-        # Ensure autoload_paths is not frozen before adding the path
-        config.autoload_paths = config.autoload_paths.dup if config.autoload_paths.frozen?
-        config.autoload_paths << datastreams_path
-      end
-    end
-
     ##
     # @!group Class Attributes
     #


config/application.rb
```diff
@@ -153,6 +153,16 @@ class Application < Rails::Application
...
+    # Add datastreams path early to prevent ActiveFedora from modifying frozen autoload_paths in Rails 7.2
+    config.before_initialize do
+      datastreams_path = Rails.root.join('app', 'models', 'datastreams').to_s
+      unless config.autoload_paths.include?(datastreams_path)
+        # Ensure autoload_paths is not frozen before adding the path
+        config.autoload_paths = config.autoload_paths.dup if config.autoload_paths.frozen?
+        config.autoload_paths << datastreams_path
+      end
+    end
+
```
orangewolf: I feel like this is an unrelated change since we don't have a datastreams directory

orangewolf (changes requested):
looks like an extra change got in, but the frozen array stuff looks good.

config/application.rb
aprilrieger: Thank you for the review. I have since reverted it to test out on my branch and I am getting failures related to data streams: https://github.com/notch8/wvu_knapsack/actions/runs/20836737945/job/59863393426#step:8:394

PR #2874, opened 2026-01-13
Limits and requests for friends cluster

maxkadel:
Connected to https://github.com/notch8/dev-ops/issues/1070

### DIFF
--- a/ops/demo-deploy.tmpl.yaml
+++ b/ops/demo-deploy.tmpl.yaml
@@ -1,10 +1,10 @@
 replicaCount: 2
 resources:
   limits:
-    memory: "2Gi"
+    memory: "4Gi"
     cpu: "1000m"
   requests:
-    memory: "1Gi"
+    memory: "2Gi"
     cpu: "100m"
 livenessProbe:
   enabled: false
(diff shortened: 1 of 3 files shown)


ops/demo-deploy.tmpl.yaml
```diff
@@ -1,5 +1,11 @@
 replicaCount: 2
-
+resources:
+  limits:
+    memory: "2Gi"
```
orangewolf: this seems to be about half what I'd expect it to be, but ruby has been getting more efficient. are we using data to drive this number?

maxkadel: Yes - for more "production-like" environments I'm defaulting to higher minimum values for memory. 

So, for the demo-deploy, the p95 and p99 for memory is only 147 MiB. Generally the highest memory needs are for start-up, and then if there's not a lot of traffic the running values are lower.

In general for these changes, I'm using using:
```ruby
mem_request = mem95_mi * 1.2
mem_limit = mem99_mi * 1.3
```
Then rounding based on scale so they're easier to read / reason about, and then I have some default minimums based on environment and type of service. So, if the mem_request calculation is lower than the default minimum, we go with the default minimum.

So for staging environments right now I have for defaults for rails:
```ruby
rails_app: { cpu_request: 100, cpu_limit: 1000, mem_request: 1024, mem_limit: 2048 }
```

and for production right now I have for defaults for rails:
```ruby
rails_app: { cpu_request: 100, cpu_limit: 1000, mem_request: 2048, mem_limit: 4096 }
```

Right now the highest calculated memory limit I have for a rails app is 16Gi.

maxkadel: After deploying the app with the limits above we did see a couple OOMKilled pods. I realized that when I measured p99 it was when there wasn't a lot of activity on the application, and probably had too low of a sample rate to pick up quick spikes. 

I've adjusted the values up for the PR.

Issue #1968, opened 2023-07-19
Add links to the sidekiq dashboard


ShanaLMoore:
### Descriptive summary

Hyku 4.1.0

related PR from ESSI: 
- https://github.com/IU-Libraries-Joint-Development/essi/pull/395

### Rationale

ESSI has a nice feature where admin users are able to easily link to the sidekiq dashboard. I personally reference the dashboard often during the development process. It'd be nice to have it contributed back. 

### Expected behavior

- [ ] A user has to manually type in the url to sidekiq, to see the dashboard. 

### Actual behavior

- [ ] A user will be able to access the sidekiq link from the admin dashboard.

### Related work

- https://github.com/IU-Libraries-Joint-Development/essi/pull/395

PR #2944, opened 2026-03-02
hyku 7.x release candidate ready: Add jobs dashboard link to admin sidebar with ability-based access

aprilrieger:
Ref ticket #1968

## Summary
Adds a Jobs dashboard link to the admin sidebar with ability-based access. **Restricted to superadmins only** per review feedback.

## Changes
- **Sidebar** (`_activity.html.erb`): Link to `/jobs` when `can?(:manage, :jobs_dashboard)`; single label "Job Dashboard" (no queue-specific naming).
- **Ability** (`ability.rb`): Only superadmins get `can :manage, :jobs_dashboard` (removed from admin permissions).
- **Routes** (`config/routes.rb`): `/jobs` protected by `Ability.new(u).can?(:manage, :jobs_dashboard)`.
- **Locales**: Added `jobs_dashboard` key for the sidebar label in en, de, es, fr, it, pt-BR, zh.

## PR feedback addressed
- Use `can?` only for visibility; no queue-type filter (any future queue is expected to provide `/jobs`).
- Single "Job Dashboard" label instead of queue-specific labels.
- Jobs dashboard access limited to superadmin (avoids prior pushback on exposing to non-developer admins).

### DIFF
--- a/spec/features/admin_dashboard_spec.rb
+++ b/spec/features/admin_dashboard_spec.rb
@@ -12,16 +12,14 @@ RSpec.describe 'Admin Dashboard', type: :feature, js: true, clean: true do
     end
 
     it 'shows the admin page' do # rubocop:disable RSpec/ExampleLength
-      # Jobs dashboard link shows when queue adapter is sidekiq or good_job; stub for consistent test
-      allow(ENV).to receive(:fetch).and_call_original
-      allow(ENV).to receive(:fetch).with('HYRAX_ACTIVE_JOB_QUEUE', 'sidekiq').and_return('sidekiq')
       visit Hyrax::Engine.routes.url_helpers.dashboard_path
       within '.sidebar' do
         expect(page).to have_link('Activity Summary')
         expect(page).to have_link('System Status')
         expect(page).to have_link("Your activity")
         expect(page).to have_link('Reports')
-        expect(page).to have_link('Sidekiq Dashboard', href: '/jobs')
+        # Should not see job dashboard link (superadmin only)
+        expect(page).not_to have_link('Job Dashboard')
       end
       within '.sidebar' do
         # Need to click link to open collapsed menu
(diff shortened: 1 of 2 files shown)


laritakr (dismissed):
I love this feature!

app/views/hyrax/dashboard/sidebar/_activity.html.erb
```diff
@@ -67,6 +67,17 @@
   </li>
 <% end %>
 
+<% jobs_queue = ENV.fetch('HYRAX_ACTIVE_JOB_QUEUE', 'sidekiq') %>
+<% if can?(:manage, :jobs_dashboard) && %w[sidekiq good_job].include?(jobs_queue) %>
```
orangewolf: the can? is sufficient here. Any added job queues should include a /jobs link to a dashboard. Anything less would be a reason to reject them as queue options.

app/views/hyrax/dashboard/sidebar/_activity.html.erb
```diff
@@ -67,6 +67,17 @@
...
+  <li class="nav-item">
+    <%= menu.nav_link('/jobs',
+                      class: "nav-link",
+                      title: t("hyrax.admin.sidebar.jobs_dashboard_#{jobs_queue}")) do %>
+      <span class="fa fa-briefcase" aria-hidden="true"></span> <span class="sidebar-action-text"><%= t("hyrax.admin.sidebar.jobs_dashboard_#{jobs_queue}") %></span>
```
orangewolf: lets just call it "Job Dashboard" and skip trying to do queue adapter specific labeling. It feels unnecessarily complex.

config/routes.rb
```diff
@@ -15,7 +15,7 @@
   mount IiifPrint::Engine, at: '/'
   mount Riiif::Engine => 'images', as: :riiif if Hyrax.config.iiif_image_server?
 
-  authenticate :user, ->(u) { u.superadmin? || u.admin? } do
+  authenticate :user, ->(u) { Ability.new(u).can?(:manage, :jobs_dashboard) } do
```
orangewolf, line 18: this is a great change

app/models/ability.rb
```diff
@@ -70,6 +70,9 @@ def admin_permissions
 
     # OVERRIDE: add custom action used in WorkAbility for "Delete Selected" button on Works dashboard index views
     can :batch_delete, :works
+
+    # Allow admins to access the Jobs dashboard (linked from admin sidebar)
+    can :manage, :jobs_dashboard
```
orangewolf: if we are going to show a link, I think it the whole thing needs to be super admin only OR we need community buy in that showing the jobs dashboard to the non-programer admins is ok. There was strong push back in the past about linking to the jobs dashboard for folks in the Amanda/Nic Don skill level (technical, but not developer).  I think its the best path is to allow admins to see them, but since it was put in and then removed, I feel like we're headed for a loop if we don't get folks using Hyku to agree first.

Issue #2950, opened 2026-03-04
Hyku: FileSets should have alt text metadata


ShanaLMoore:

laritakr:
Refers to https://github.com/samvera/hyrax/issues/7236

PR #2951, opened 2026-03-05
Use alt text on views

laritakr:
## Summary

Ref https://github.com/samvera/hyku/issues/2950

Adds alt text


spec/features/collection_type_spec.rb
```diff
@@ -558,7 +558,7 @@
         # email and wait one second for the item to populate in the table
         within('#s2id_collection_type_participant_agent_id') do
           fill_in "s2id_autogen1", with: 'us'
-          sleep 1
+          sleep 3
```
laritakr, line 561: Increasing sleep time to try to minimize the times this flaky spec fails.

PR #2952, opened 2026-03-05
hyku 7.x release candidate ready: Implement adapter-agnostic runtime for Fedora-off operation

aprilrieger:
## Summary

Supports running Hyku with Wings disabled (Postgres-only, no Fedora) and keeps the test suite and CI passing in both Wings-on and Wings-off modes.

### Configuration & load order
- **ValkyrieLazyMigration**: All `ValkyrieLazyMigration.migrating` calls moved from model files into `config/initializers/wings.rb`, inside the existing `after_initialize` block when Wings is enabled. Model files no longer contain this line or any `HYRAX_SKIP_WINGS` checks.
- **No-op when Wings disabled**: `config/initializers/valkyrie_lazy_migration_noop.rb` makes `ValkyrieLazyMigration.migrating` a no-op when `Hyrax.config.disable_wings` is true, so any future call site can stay as-is without env checks in models.

### Specs & CI
- **rails_helper**: For examples with `:clean` / `:clean_repo` or `type: :feature`, cleanup uses `Hyrax::SolrService.wipe!` when Wings is disabled (or when `ENV['DISABLE_WINGS']` is set), and `ActiveFedora::Cleaner.clean!` only when Wings is enabled. Avoids Fedora connection refused (e.g. port 8986) in CI when Fedora is not present.
- **Sipity**: In `app/models/sipity.rb`, added a fallback when resolving `Entity` by GID: if the GID is a ValkyrieGlobalIdProxy and the direct lookup misses, resolve the GID and try lookups by the concrete resource class and by `internal_resource`, so entities stored with legacy/concrete model names are found. Fixes `Sipity::ConversionError` in `sipity_spec` when Wings is disabled.

### Docker / no-Fedora mode
- **.env**: Removed default Wings-skip env vars; comment points to docker-compose override for Postgres-only.
- **docker-compose.override-nofcrepo4.yml**: Defines no-Fedora env (e.g. `DISABLE_WINGS=true`) for `web` and `worker` when using the no-Fedora stack.

### Other
- Config and initializers updated to use `Hyrax.config.disable_wings` (and ENV where needed for load order or CLI) so behavior is consistent and testable in both modes.

## Testing

- CI (build-test-lint) passes with the above spec/rails_helper and Sipity changes.
- Run specs with Wings disabled: `DISABLE_WINGS=true bundle exec rspec spec` (e.g. in no-Fedora compose or with Fedora not running).
- Run specs with Wings enabled: default compose with Fedora; `bundle exec rspec spec`.


app/models/sipity.rb
```diff
@@ -40,10 +40,12 @@ def Entity(input, &block) # rubocop:disable Metrics/AbcSize
                Hyrax.logger.debug("  Searching for GID: #{gid_string}")
                Entity.find_by(proxy_for_global_id: gid_string)
              when SolrDocument
-               if Hyrax.config.valkyrie_transition? # we need the actual model, not the mapped "Resource" model
+               if Hyrax.config.disable_wings
```
orangewolf, line 47: revert me!

app/services/hyrax/statistics/works/over_time_decorator.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
orangewolf, line 1: we need an override comment here, that calls this out as a Hyrax bug and links to the issue in hyrax that describes this problem

app/services/hyrax/admin_set_create_service_decorator.rb
```diff
@@ -6,6 +6,22 @@ module Hyrax
   # Because Hyku has converted the Hyrax::Group model from a PORO to a db-backed active record object,
   # we have to query for existing Hyrax groups instead of initializing empty ones.
   module AdminSetCreateServiceDecorator
+    module ClassMethods
+      private
+
+      # In transition-off mode, avoid probing legacy default admin set IDs through
+      # Hyrax.query_service. That legacy probe can hit Wings/Fedora query paths.
+      def find_unsaved_default_admin_set
```
orangewolf: this change seems like it should not be required. the fcrepo fall back is an fcrepo 6 thing, not a wings / af thing

bin/db-migrate-seed.sh
```diff
@@ -43,14 +49,23 @@ begin
...
   db_name = ENV['DB_NAME']
   db_password = ENV['DB_PASSWORD']
 
+  puts "[db-migrate-seed] DISABLE_WINGS=#{ENV['DISABLE_WINGS'].inspect} VALKYRIE_TRANSITION=#{ENV['VALKYRIE_TRANSITION'].inspect} parsed_disable_wings=#{disable_wings}"
+  puts "[db-migrate-seed] FCREPO_HOST=#{fcrepo_host.inspect} FCREPO_PORT=#{fcrepo_port.inspect}"
+  puts "[db-migrate-seed] will_wait_for_fcrepo=#{(!disable_wings && !fcrepo_host.to_s.empty?).inspect}"
+
   service_wait("#{db_host}:#{db_port}")
-  service_wait("#{fcrepo_host}:#{fcrepo_port}") if fcrepo_host
+  service_wait("#{fcrepo_host}:#{fcrepo_port}") if !disable_wings && fcrepo_host
```
orangewolf: `if fcrepo_host`

config/initializers/wings.rb
```diff
@@ -87,10 +94,17 @@
     Hyrax.query_service.services[0].custom_queries.register_query_handler(handler)
   end
 
-  [
-    Wings::CustomQueries::FindBySourceIdentifier
-  ].each do |handler|
-    Hyrax.query_service.services[1].custom_queries.register_query_handler(handler)
+  if Hyrax.config.disable_wings
```
orangewolf, line 119: use valkyrie_transisiton everywhere. until we change it ;-)

app/models/sipity.rb
aprilrieger: By the end of our pairing we actually stated that we were moving towards disable_wings as the consistent source of truth.

config/initializers/wings.rb
aprilrieger: By the end of our pairing we actually stated that we were moving towards disable_wings as the consistent source of truth.

bin/db-migrate-seed.sh
aprilrieger: Keep it simple April 😆

Gemfile.lock
```diff
@@ -1081,6 +1049,179 @@ GEM
     openseadragon (0.9.0)
       rails (> 6.1.0)
     openssl (3.3.0)
+    opentelemetry-api (1.7.0)
```
aprilrieger: This was from the besties deploy I will remove this part before merging.

bin/db-migrate-seed.sh
aprilrieger: Had an issue with just having fcrepo_host as the guard due to the hyrax helm template helper setting the `fcrepo_host= NO_FCREPO_HOST_DEFINED' which satisfies that guard and during the deploy, it's stuck in the db-wait loop waiting for fcrepo and the deployment fails.\

New guard:
`fcrepo_host != 'NO_FCREPO_HOST_DEFINED'` added and deployment succeeds

https://new.hyku-besties.notch8.cloud

.github/workflows/deploy.yaml
```diff
@@ -5,13 +5,14 @@ on:
...
         description: "Deploy to Environment"
-        require: true
+        required: true
         default: "staging"
         type: choice
         options:
           - demo
           - iiif
           - staging
+          - besties
```
orangewolf: its weird to me for this Notch8 specific label to be in Hyku, could this be in the hykuup knapsack? Or could we call it something else?

app/models/admin_set_resource.rb
```diff
@@ -3,7 +3,10 @@
 class AdminSetResource < Hyrax::AdministrativeSet
   include Hyrax::ArResource
   include Hyrax::Permissions::Readable
-  Hyrax::ValkyrieLazyMigration.migrating(self, from: ::AdminSet)
+
+  # NOTE: Uses ENV rather than Hyrax.config.disable_wings because this line
+  # executes at class load time, before Hyrax configuration is fully initialized.
+  Hyrax::ValkyrieLazyMigration.migrating(self, from: ::AdminSet) unless ENV["HYRAX_SKIP_WINGS"] == "true"
```
orangewolf: this feels really gross because we aren't using the Hyrax configuration stuff, but I see why it has to be this way. At minimum we need to do the `        ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_SKIP_WINGS', false))` pattern to make it match right. _but_ I would say that instead of doing that on every one of these we should maybe change the `ValkyrieLazyMigration#migrating` definition to be a no-op so these can be left alone.

orangewolf: ooh, better idea - Lets move them all in to a trigger after the config is loaded. The line gets removed from here completely. In the `hyku/config/initializers/hyrax.rb` file, have 
```ruby
unless Hyrax.config.disable_wings
  AdminSetResource.class_eval { Hyrax::ValkyrieLazyMigration.migrating(self, from: ::AdminSet) }
  CollectionResource.class_eval { Hyrax::ValkyrieLazyMigration.migrating(self, from: ::Collection)}
...
end
```

then the whole thing loads in the right order.

app/views/proprietor/accounts/edit.html.erb
```diff
@@ -81,7 +81,7 @@
             <%= s.input :url, required: false %>
           <% end %>
 
-          <% unless @account.search_only? %>
+          <% if !Hyrax.config.disable_wings && !@account.search_only? %>
```
orangewolf, line 84: thank you for flipping this when adding the `&&`

app/models/admin_set_resource.rb
orangewolf: the above could also go nicely in the config/initializers/wings.rb file with all the other migration shenanigans and still get the load order benefit

.env
```diff
@@ -11,10 +11,10 @@ DB_PORT=5432
 DB_TEST_NAME=hyku_test
```
orangewolf, line 11: these changes probably shouldn't be defaults yet. You might consider adding them to your docker-compose.override.yml file so they don't get committed on accident ;-)

.github/workflows/deploy.yaml
aprilrieger: Yes, will strip out besties deploy before merge. Just needed to test deploy to a sep env.

app/models/sipity.rb
aprilrieger, line 1: @laritakr tagging you here as Rob stated you ran into this before and might have a solution already? LMK!

aprilrieger: @orangewolf just to close the loop, LaRita has a issue ticket against Hyrax for the issue, and I added it to the override comment.

orangewolf (approved):

PR #2956, opened 2026-03-06
No creator on admin set form

laritakr:
## Summary

Updates m3 profiles to remove creator from the flexible admin set form since it is not settable from the UI (setting happens via a transaction step).

### DIFF
--- a/spec/fixtures/files/m3_profile.yaml
+++ b/spec/fixtures/files/m3_profile.yaml
@@ -136,7 +136,6 @@ properties:
   creator:
     available_on:
       class:
-        - AdminSetResource
         - CollectionResource
         - Hyrax::FileSet
         - GenericWorkResource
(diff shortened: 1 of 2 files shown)


spec/fixtures/files/m3_profile.yaml
```diff
@@ -157,6 +156,27 @@ properties:
     range: http://www.w3.org/2001/XMLSchema#string
     sample_values:
       - Julie Allinson
+  creator_hidden:
```
ShanaLMoore, line 159: ohh, interesting 👀 

Will it still get set in the transaction if we remove the property?

laritakr: we didn't remove the property. creator_hidden is renamed to creator, so we can have one that isn't shown on the form.

PR #2964, opened 2026-03-12
Controlled vocabulary form modifications

laritakr:
# Summary

- Allows M3 or yaml metadata to both control whether controlled vocabularies show as singular or multiple.
- Removes some of the hardcoded handling of resource_type controlled vocabulary for Oer class's special vocabulary list.
- Adds local vocabulary support for accessibility_features and accessibility_hazards
- Removes obsolete form partials
- Adjusts default m3 profile

### DIFF
--- a/app/helpers/hyrax/form_helper_behavior.rb
+++ b/app/helpers/hyrax/form_helper_behavior.rb
@@ -10,7 +10,7 @@ module Hyrax
       Hyrax::ControlledVocabularies.remote_authorities[source_name]
     end
 
-    def controlled_vocabulary_options_for(property_name, record_class)
+    def controlled_vocabulary_options_for(property_name)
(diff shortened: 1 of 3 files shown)


app/views/records/edit_fields/_rights_statement.html.erb
ShanaLMoore, line 1: how does this field render when HYRAX_FLEXIBLE is false? Does it just fall back to hyrax behavior? And when it's true, it uses the default partial? I'm curious about seeing conditionals in the other partials but not here.

laritakr: Yes, because Hyrax has a partial for rights_statement, that is what gets used. Rights statement will never fall back to default because of Hyrax.

PR #2979, opened 2026-03-20
Adds m3 profile rake tasks

laritakr:
# Summary

Adds rake tasks for m3 profiles:

- `rake hyku:flexible_schema:initialize` adds a default m3 profile to each tenant that doesn't have one.
- `rake reset:all_flexible_schemas` removes all flexible schemas from all tenants (useful for testing but VERY dangerous in real life!)

### DIFF
--- a/lib/tasks/reset.rake
+++ b/lib/tasks/reset.rake
@@ -13,12 +13,27 @@ namespace :reset do
 
   desc 'reset work and collection data from a single tenant, any argument that works with switch!() will work here'
   task :works_and_collections, [:account] => [:environment] do |_t, args|
-    raise "You must speficy a name, cname or id of an account" if args[:account].blank?
+    raise "You must specify a name, cname or id of an account" if args[:account].blank?
 
     confirm("You are about to delete all works and collections from #{args[:account]}, this is not reversable!")
     switch!(args[:account])
     Rake::Task["hyrax:reset:works_and_collections"].reenable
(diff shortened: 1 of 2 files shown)


lib/tasks/reset.rake
```diff
@@ -13,14 +13,29 @@ namespace :reset do
...
-    raise "You must speficy a name, cname or id of an account" if args[:account].blank?
+    raise "You must specify a name, cname or id of an account" if args[:account].blank?
 
     confirm("You are about to delete all works and collections from #{args[:account]}, this is not reversable!")
     switch!(args[:account])
     Rake::Task["hyrax:reset:works_and_collections"].reenable
     Rake::Task["hyrax:reset:works_and_collections"].invoke
   end
 
+  desc 'Remove all flexible schemas across all tenants'
```
ShanaLMoore, line 24: this seems scary lol 🤔 but w a confirm hopefully it's less likely to be done without realizing the consequences.

ShanaLMoore: Could we maybe do an additive version instead of destroying them all? Just a thought - non blocking

laritakr: it is dangerous, but that same reset file also removes all works & collections (also dangerous!) So I felt it was consistent.

laritakr: This is unrelated to the adding a new one for everyone... it's more for dev use.

PR #2986, opened 2026-03-25
:bug: Fix collection title showing jQuery .text on first Add 

ShanaLMoore:
# Summary

Initialize collection Select2 in `RelationshipsControl#init` before `select2('data')` so the first “Add” shows the collection title, not minified jQuery.

## Ticket Number

https://github.com/notch8/hykuup_knapsack/issues/656

## BEFORE

image

## Expected Behavior

First “Add” on the work Relationships → Collections row shows the real collection name.

image

## Notes

**Root cause in Hyku:** Your `control.es6` is an old Hyku copy from Hyrax 2.5. Current Hyrax adds `ensureSelect2()` in `init()` so the collection `<select>` always gets Select2 before anyone reads `select2('data')`. That method was missing from the Hyku override.

ref: 
- https://github.com/samvera/hyrax/blob/main/app/assets/javascripts/hyrax/relationships/control.es6#L33

@samvera/hyku-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/relationships/control.es6
+++ b/app/assets/javascripts/hyrax/relationships/control.es6
@@ -33,8 +33,31 @@ export default class RelationshipsControl {
   }
 
   init() {
+    this.ensureSelect2();
     this.bindAddButton();
     this.displayMembers();
+  }
+
+  /**
+   * Ensure Select2 is initialized on collection dropdowns before reading selection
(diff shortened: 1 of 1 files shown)


laritakr (approved):
Seems like a reasonable attempt to fix a sporadic behavior. Hopefully it resolves it!

PR #3002, opened 2026-04-08
Fix SaveCollectionThumbnail removing thumbnail metadata when unchanged

aprilrieger:
Honor thumbnail_unchanged_indicator before remove_thumbnail; skip blank alt updates that would wipe stored text. Add RSpec regression coverage and consolidate duplicate spec file.

Ref ticket # https://github.com/notch8/hykuup_knapsack/issues/619


maxkadel (approved):
Thanks April! And thanks for collapsing the specs!

PR #3007, opened 2026-04-10
Log a warning if someone may have an invalid Solr password

maxkadel:
- Refactor generation of solr_url_parts to be more idiomatic
- Checks Solr credential encoding to make sure it's url escaped, if not, logs a warning.

@samvera/hyku-code-reviewers


aprilrieger (approved):
Fantastic! Thanks Max!

PR #3009, opened 2026-04-10
🚧 Include bulkrax guided metrics 🚧 

laritakr:
# Summary

Use the guided metrics branch of Bulkrax.

Update the schema to include the new table, and add the new feature flag and ability.

Refs https://github.com/samvera/bulkrax/pull/1172


orangewolf:
@laritakr is this still needed?

laritakr:
@orangewolf yes this was the last piece of the guided import work that ran out of funding and never made it all the way in. But we don't want to lose the work. 

@SixtyGrit do we have additional plans regarding this?

PR #3011, opened 2026-04-13
Solr replication size bug

maxkadel:
Adds `SOLR_COLLECTION_REPLICAS` as default value for `SOLR_COLLECTION_REPLICAS`. Already present in the `bin/solrcloud-assign-configset.sh` script, it was unexpected that it was not used elsewhere.

New tenants will receive the new default, either `SOLR_COLLECTION_REPLICAS` if set, or `1` (also Solr's default).

Old tenants will keep whatever value they have saved in the database - usually nil, which is then resolved by Solr to 1.

@samvera/hyku-code-reviewers


aprilrieger (approved):
Thanks Max!!

PR #3013, opened 2026-04-14
Remove per-field Solr spellcheck.dictionary from catalog search fields

aprilrieger:
## Summary

Ref ticket # https://github.com/notch8/palni_palci_knapsack/issues/592

Removes `spellcheck.dictionary` from Blacklight catalog search field configs so Solr is not asked for named spellcheck dictionaries that Hyku's default Solr configset does not define.

## Problem

Fielded catalog searches (e.g. work-show links with `search_field=creator` and a `q` value) sent `spellcheck.dictionary=creator` (and similar for other fields). Solr responded with errors such as "Specified dictionaries do not exist," leading to RSolr HTTP errors, `Blacklight::Exceptions::InvalidRequest`, and the "Sorry, I don't understand your search" behavior while results were wrong or unscoped.

## Solution

Drop `field.solr_parameters = { "spellcheck.dictionary": … }` from the shared catalog search fields in `hyrax-webapp/app/controllers/catalog_controller.rb` and from the knapsack `catalog_controller_decorator` loop (including `format`). Blacklight still sets `spellcheck.q` where appropriate; only the invalid per-field dictionary names are removed.

## Testing

- [ ] `/catalog?search_field=creator&q=…` (and another field, e.g. keyword) returns 200 and scoped results, no Solr spellcheck dictionary errors in logs.

## Notes

Sites that configure matching named spellcheck dictionaries in Solr can reintroduce `spellcheck.dictionary` in a local override if they want field-specific spellcheck.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -247,10 +247,12 @@ class CatalogController < ApplicationController
     # of Solr search fields.
     # creator, title, description, publisher, date_created,
     # subject, language, resource_type, format, identifier, based_near,
+    #
+    # Hyku solr/conf/solrconfig.xml only registers spellcheck dictionaries named
+    # default, author, subject, and title. Omit per-field spellcheck.dictionary here;
+    # unknown names cause Solr errors and Blacklight::Exceptions::InvalidRequest.
     config.add_search_field('contributor') do |field|
       # solr_parameters hash are sent to Solr as ordinary url query params.
-      field.solr_parameters = { "spellcheck.dictionary": "contributor" }
-
       # :solr_local_parameters will be sent using Solr LocalParams
       # syntax, as eg {! qf=$title_qf }. This is neccesary to use
       # Solr parameter de-referencing like $title_qf.
@@ -263,7 +265,6 @@ class CatalogController < ApplicationController
     end
 
     config.add_search_field('creator') do |field|
-      field.solr_parameters = { "spellcheck.dictionary": "creator" }
       solr_name = 'creator_tesim'
       field.solr_local_parameters = {
         qf: solr_name,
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -247,10 +247,12 @@ def self.uploaded_field
     # of Solr search fields.
     # creator, title, description, publisher, date_created,
     # subject, language, resource_type, format, identifier, based_near,
+    #
+    # Hyku solr/conf/solrconfig.xml only registers spellcheck dictionaries named
+    # default, author, subject, and title. Omit per-field spellcheck.dictionary here;
+    # unknown names cause Solr errors and Blacklight::Exceptions::InvalidRequest.
```
maxkadel, line 253: Question - should we leave one commented-out one as an example? I think that matches the pattern in the rest of the catalog controller.

maxkadel (dismissed):
Thanks, looks good! 

I have one non-blocking question, happy to have it merged after you read it.

app/controllers/catalog_controller.rb
aprilrieger: Can you give me an example of what you are talking about, happy to include....

What i have interpretted is the to leave `# field.solr_parameters = { "spellcheck.dictionary": "contributor" }` commented out?

maxkadel, line 253: Something like this?

```suggestion
    #
    # Hyku solr/conf/solrconfig.xml only registers spellcheck dictionaries named
    # default, author, subject, and title. Omit per-field spellcheck.dictionary here, unless it is also added to the solrconfig.xml;
    # unknown names cause Solr errors and Blacklight::Exceptions::InvalidRequest.
    # To add per-field spellcheck, add e.g.
    # field.solr_parameters = { "spellcheck.dictionary": "contributor" }
    # inside the add_search_field block
```

aprilrieger: Thanks Max!

maxkadel (approved):
Thanks again! Looks good

PR #3023, opened 2026-04-17
Resolve incorrect routes on edit

laritakr:
# Summary

Ref: https://github.com/samvera/hyku/issues/3024

# Note: 

I have reverted a first attempt at a fix that was inadequate. I added a failing spec, and am sharing Claude's plan in [the issue ticket](https://github.com/samvera/hyku/issues/3024), containing several options to patch the bug.


orangewolf:
@laritakr can this be closed?

laritakr:
@orangewolf I'm pretty sure the situation still exists. This PR has a failing spec, and any fix would need to make the spec pass.

PR #3034, opened 2026-04-30
Improve search results performance - reduce N+1 queries on search results page

maxkadel:
In investigating performance issues with a local application, we found some queries that were increasing the number of calls to the database significantly for each additional search result.

This fixes one of those places, the GroupAwareRoleChecker. It memoizes:
- The current user's Hyrax groups, based on the site instance as a memo key
- An overall role memo, using the combination of the role name and the site instance ID as a memo key, to ensure we are not over-memo-izing.

- With memoization, ability checks should not scale linearly with result count.
- Without GroupAwareRoleChecker memoization (main): 462 queries for 5 works + 3 collections (admin user)
- With memoization (this branch): ~390 queries for the same set

@samvera/hyku-code-reviewers

## Query count benchmarks (total / uncached)
From the test - spec/features/catalog_query_count_spec.rb
| Branch                  | User  | 1 result        | 8 results       | Total delta | Uncached delta | 1 result delta from main | 8 results delta from main |
|-------------------------|-------|-----------------|-----------------|-------------|----------------|--------------------------|---------------------------|
| main                    | admin | 263 (53 unc)    | 462 (54 unc)    | 199         | 1              | n/a                      | n/a                       |
| n_plus_one_exploration  | admin | 197 (53 unc)    | 390 (54 unc)    | 193         | 1              | 66                       | 72                        |
| n_plus_two              | admin | 223 (29 unc)    | 406 (30 unc)    | 183         | 1              | 40                       | 56                        |
| combined_branch         | admin | 157 (29 unc)    | 334 (30 unc)    | 177         | 1              | 106                      | 128                       |
|                         |       |                 |                 |             |                |                          |                           |
| main                    | anon  | 170 (46 unc)    | 339 (47 unc)    | 169         | 1              | n/a                      | n/a                       |
| n_plus_one_exploration  | anon  | 170 (46 unc)    | 339 (47 unc)    | 169         | 1              | 0                        | 0                         |
| n_plus_two              | anon  | 92  (22 unc)    | 283 (23 unc)    | 191         | 1              | 78                       | 56                        |
| combined_branch         | anon  | 92  (22 unc)    | 283 (23 unc)    | 191         | 1              | 78                       | 56                        |

### DIFF
--- a/app/services/group_aware_role_checker.rb
+++ b/app/services/group_aware_role_checker.rb
@@ -12,16 +12,25 @@ module GroupAwareRoleChecker
 
   private
 
+  def current_user_hyrax_groups(site_instance)
+    @current_user_hyrax_groups ||= {}
+    @current_user_hyrax_groups[site_instance.id] ||= current_user.hyrax_groups
+  end
+
   # Check for the presence of the passed role_name in the User's Roles and
   # the User's Hyrax::Group's Roles.
   def group_aware_role?(role_name)
     return false if current_user.new_record?
-    return true if current_user.has_role?(role_name, Site.instance)
 
-    current_user.hyrax_groups.each do |group|
-      return true if group.site_role?(role_name)
-    end
+    @group_aware_role_cache ||= {}
 
-    false
+    site_instance = Site.instance
+
+    cache_key = [role_name, site_instance.id]
+    return @group_aware_role_cache[cache_key] if @group_aware_role_cache.key?(cache_key)
+
+    @group_aware_role_cache[cache_key] =
+      current_user.has_role?(role_name, site_instance) ||
+      current_user_hyrax_groups(site_instance).any? { |group| group.site_role?(role_name) }
   end
 end
--- a/app/services/group_aware_role_checker.rb
+++ b/app/services/group_aware_role_checker.rb
@@ -27,7 +27,7 @@ module GroupAwareRoleChecker
 
     site_instance = Site.instance
 
-    memo_key = [role_name, site_instance.id]
+    memo_key = [role_name, site_instance.id, current_user.id || current_user.object_id]
     return @group_role_memo[memo_key] if @group_role_memo.key?(memo_key)
 
     @group_role_memo[memo_key] =
--- a/app/services/group_aware_role_checker.rb
+++ b/app/services/group_aware_role_checker.rb
@@ -13,8 +13,8 @@ module GroupAwareRoleChecker
   private
 
   def current_user_hyrax_groups(site_instance)
-    @current_user_hyrax_groups ||= {}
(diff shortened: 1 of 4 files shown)


aprilrieger (dismissed):
What an exciting find! Thanks Max!

app/services/group_aware_role_checker.rb
```diff
@@ -12,16 +12,25 @@ module GroupAwareRoleChecker
 
   private
 
+  def current_user_hyrax_groups(site_instance)
+    @current_user_hyrax_groups ||= {}
+    @current_user_hyrax_groups[site_instance.id] ||= current_user.hyrax_groups
+  end
+
```
orangewolf, line 20: I believe this is a problem. Let's say we have user 1 and user 2 from site 1. Then we have user 3 from site 2. If they manage to hit the same process then you'd get

request 1 (user 1)
@current_user_hyrax_groups[1] = user_1_groups
request 2 (user 3)
@current_user_hyrax_groups[2] = user_3_groups
request 3 (user 2)
@current_user_hyrax_groups[1] == user_1_groups

thus user 2 would get user 1s group permissions

app/services/group_aware_role_checker.rb
```diff
@@ -12,16 +12,25 @@ module GroupAwareRoleChecker
...
-    return true if current_user.has_role?(role_name, Site.instance)
 
-    current_user.hyrax_groups.each do |group|
-      return true if group.site_role?(role_name)
-    end
+    @group_aware_role_cache ||= {}
+
+    site_instance = Site.instance
+
+    cache_key = [role_name, site_instance.id]
```
orangewolf: this has the same problem as above

orangewolf (changes requested):

app/services/group_aware_role_checker.rb
maxkadel: I've re-named things to make it clearer that this is memoization, not a true cache.

The memoization only lasts for the life of the object, which is re-created with each Ability / each call to `.can?`, so this should not bleed between sites or users. I've tried to add tests to prove this to both of us.

Let me know if you're still concerned about the memoization, or if you have ideas about how to prove (or disprove) this possibility more definitively.

app/services/group_aware_role_checker.rb
maxkadel: same as above

app/services/group_aware_role_checker.rb
orangewolf: thank you, I see that now more clearly. I should have opened the whole class file to see that current user isn't set w/ in the context of this module and that ability only allows it to be set on instantiation, current user can't be changed in an ability instance w/o significant shenanigans.

orangewolf (dismissed):

orangewolf (approved):

PR #3037, opened 2026-05-01
Scope GroupAwareRoleChecker memo by user and site

aprilrieger:
Stacked on #3034: narrows memo keys so hyrax groups (and optionally role memo) cannot mix users on the same site if one checker instance is reused, matching @orangewolf’s review.

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -52,6 +52,10 @@ RSpec.describe CatalogController do
     end
   end
   describe 'memoization isolation' do
+    after do
+      allow(Site).to receive(:instance).and_call_original
+    end
+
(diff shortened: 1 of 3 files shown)


orangewolf (approved):
I like this - it costs almost nothing in terms of performance and keeps this code safe if either the implementation of Ability changes or if the module gets reused down the road

spec/controllers/catalog_controller_spec.rb
```diff
@@ -52,6 +52,10 @@
     end
   end
   describe 'memoization isolation' do
+    after do
```
maxkadel, line 55: Not a blocker, but I would not expect this after block to be needed - the "allow x to_receive" should only live for the life of the block without a cleanup step.

maxkadel (approved):
Thanks April! Makes a lot of sense for future proofing.

PR #3039, opened 2026-05-05
Fix flaky account spec

maxkadel:
We have to cache the default options and reset them in an around block, so that even if a test fails other tests won't be impacted.


laritakr (approved):
Looks good!

PR #3041, opened 2026-05-06
Adds redirects feature to Hyku

laritakr:
## Summary

- Brings the URL redirects feature (recently merged into Hyrax) into Hyku, with Hyku-specific wiring so adopters can preserve legacy URLs from previous repository systems on a per-tenant basis.
- Refs notch8/palni_palci_knapsack#600 (epic), notch8/palni_palci_knapsack#621 (Bulkrax round-trip), and notch8/palni_palci_knapsack#630 (Hyku consolidation).

### What this delivers

URL redirects let admins register arbitrary paths as 301 redirects to a work or collection's permanent URL. The feature is the migration safety net for institutions moving from DSpace, CONTENTdm, Islandora, bepress, etc. — legacy URLs in citations, search engine indexes, and bookmarks keep resolving after the move.

Hyrax shipped the core feature (data model, validator, indexer, resolver, form tab on works). This PR adds the Hyku-side wiring needed to make it usable end-to-end: catch-all resolver route, Hyku's own reserved-prefix list, the redirects tab on the collection form, the Bulkrax mapping for CSV import/export of redirects, and the migration that creates the uniqueness-tracking table.

### Feature gating

Off by default. Both switches must be on for the resolver to engage:

1. **App config**: `HYRAX_REDIRECTS_ENABLED=true` (or `config.redirects_enabled = true`). Controls whether the schema, attribute, route, and Flipflop are registered at all.
2. **Flipflop**: toggle `:redirects` under experimental_features. Controls runtime behavior; can be scoped per-tenant via Hyku's Flipflop strategy chain.

When both are off, Hyku behaves as if the feature didn't exist. Turn on for local dev via [docker-compose.override.yml](https://github.com/user-attachments/files/27456998/docker-compose.override.yml).

### Bulkrax integration

CSV import/export uses Bulkrax's `nested_attributes: true` field-mapping flag (introduced in samvera/bulkrax#1195, released in v9.5.0). Each redirect maps to three numbered columns: `redirect_path_<n>`, `redirect_canonical_<n>`, `redirect_sequence_<n>`. The columns appear in the downloadable CSV template and round-trip through import and export.

The Hyku Gemfile pins `bulkrax ~> 9.5`.

### Reserved-prefix list

Hyku-specific tenant-host paths get added to `Hyrax.config.reserved_redirect_prefixes` so admins can't accidentally claim them as redirect aliases: `/authorities`, `/bookmarks`, `/browse`, `/exporters`, `/identity_providers`, `/importers`, `/jobs`, `/single_signon`, `/status`, `/sword`. Admin-host-only routes (`/account`, `/proprietor`) are intentionally excluded — they're host-scoped via route constraints and never reach a tenant host, so reserving them would prevent legitimate redirect targets on tenants.

### Other changes in this PR

- `CollectionUpdateDecorator` is refactored to derive its step list from `Hyrax::Transactions::CollectionUpdate::DEFAULT_STEPS` instead of hardcoding. Same end behavior for thumbnails; automatically inherits the new `sync_redirect_paths` step from upstream.
- `Hyku::FileRenameable` (the favicon uploader) forces UTF-8 encoding on the generated filename, and the favicon-uploader spec wraps the file `store!` inside its `freeze_time` block. Both were drive-by fixes during testing of this branch — the spec was a long-standing flake exposed by encoding behavior, and the encoding fix prevents a real-world inconsistency between filename strings produced at write vs. read time.

### Reference

See [vendor/gems/hyrax/documentation/redirects.md](vendor/gems/hyrax/documentation/redirects.md) for the full feature reference, including the new "Migration playbook (Bulkrax)" section that walks through the CSV column format, the reserved-prefix behavior, the reindex command, and common-error troubleshooting.

## Testing

A reviewer should manually confirm the following with the feature turned on (config + Flipflop) under both `HYRAX_FLEXIBLE` modes:

- [ ] Two-layer gating: with config off, no Redirects tab on forms and the Flipflop UI doesn't show the feature; with config on but Flipflop off, the tab appears but the resolver returns 404 for registered paths; with both on, registered paths return 301.
- [ ] Form on a work and a collection: add a redirect, mark it canonical, save, reload — the entry persists and the canonical badge shows on the right entry.
- [ ] Path normalization: pasting a full URL (e.g. `https://old.example.edu/handle/123?utm=email`) is silently accepted as `/handle/123`. Trailing slashes are stripped.
- [ ] Validation: a row with two canonical entries fails; a row whose path is already in use on a different record fails; a path matching one of Hyku's reserved prefixes (e.g. `/single_signon`) is rejected, but a non-reserved lookalike (`/single_signon_admin`) is accepted.
- [ ] Resolver returns 301 for registered paths (works → `/concern/<plural>/<id>`, collections → `/collections/<id>`) and 404 for unregistered paths. Real Hyku routes (`/dashboard`, `/catalog`, etc.) are unaffected by the catch-all.
- [ ] Per-tenant Flipflop scoping: the feature can be enabled on tenant A and left off on tenant B; the resolver responds accordingly. The same path registered on two different tenants resolves to each tenant's own record (cache-key isolation).
- [ ] Bulkrax: download the CSV template and confirm the `redirect_path`, `redirect_canonical`, `redirect_sequence` columns appear (no numbered suffixes in the template). Import a CSV with `redirect_path_1=/foo,redirect_canonical_1=true,redirect_path_2=/bar` and confirm the resulting work has both redirects and `/foo` resolves after Solr reindex. Re-export the work and confirm the columns round-trip.
- [ ] Collection update regression: after the decorator refactor, thumbnail, banner, and logo uploads on collections still work.
- [ ] Catch-all interactions: OAI-PMH endpoints, Bulkrax importer/exporter URLs, SSO, and `/status` are all unaffected by the catch-all redirect route.


laritakr:
A few more fixes are coming, but there are some dependency issues between hyrax and hyku affecting another pull request, so it's important to get this in and then update hyrax again to bring in the other fixes.

db/migrate/20240606205215_create_hyrax_flexible_schemas.hyrax.rb
laritakr, line 1: The migration was originally added without the .hyrax.rb suffix that rails hyrax:install:migrations uses to detect already-installed migrations, so a recent Hyrax bump tried to copy it in a second time and triggered DuplicateMigrationNameError. Adding the suffix matches the installer's naming convention, and because Rails tracks applied migrations by timestamp prefix only (which is unchanged), the existing schema_migrations entry still matches.

orangewolf (approved):
Looks good.

PR #3059, opened 2026-05-14
Fix accessibility issue on Show Pages: improve breadcrumb and thumbna…

### LINKED ISSUE #3054: Ensure that text and background colors have enough contrast on Show Pages
vpat.hykudemo.org
Version 7.1

### Descriptive summary, including screenshots, if any
Color contrast on this type of page does not meet the minimum accessibility standard.
Increase the color contrast to 4.5:1 or higher.
Currently, the contrast is 4.16 : 1
Color details:
- Text color: #2e74b2 / RGB (46, 116, 178)
- Background color: #e9ecef / RGB (233, 236, 239)
Specific areas to be fixed:
- Breadcrumbs: Home, Dashboard, Works, <title of the work>
- Thumbnail title 

### Rationale
Text and background colors that conflict or have too little contrast between each other can cause readability issues for users who are color blind or have low vision (vision loss equivalent to approx. 20/40 vision).

### Expected behavior
Color contrast is at least 4.5:1

### Actual behavior
Color contrast is 4.16:1.

Use SiteImprove on this type of page to see the error.

### Related work (other tickets, possibly the same issue in Hyrax)

Image

Image

trmccormick:
Fixes #3054

Improve breadcrumb link contrast on Show Pages

Breadcrumb links on show pages (works, file sets, etc.) were using the tenant-configured link color (#2e74b2) against the Bootstrap breadcrumb background (#e9ecef), resulting in a contrast ratio of 4.16:1 — below the WCAG AA minimum of 4.5:1. This impacts readability for users with color blindness or low vision.

Root cause: The tenant link color is injected via an inline <style> block generated by _appearance_styles.html.erb (body.public-facing a { color: <tenant-color>; }). Because inline styles load after compiled CSS, SCSS-only fixes were overridden by the tenant appearance settings.

Fix: Added nav.breadcrumb a { color: #000000 !important; } at the end of the inline <style> block in _appearance_styles.html.erb, after the dynamic link color rule, ensuring it takes effect regardless of tenant appearance configuration.

Changes:

_appearance_styles.html.erb — Override breadcrumb link color to #000000 for WCAG AA compliance
hyku.scss — Remove earlier ineffective .breadcrumb-item a rule (wrong selector, wrong location)
Verified: Rule confirmed active in browser dev tools with computed color rgb(0, 0, 0) on the breadcrumb links.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -766,3 +766,8 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
+
+// Accessibility fix for contrast on Show Pages
+.breadcrumb-item a {
+    color: #000000;
+}
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -766,8 +766,3 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
-
-// Accessibility fix for contrast on Show Pages
-.breadcrumb-item a {
-    color: #000000;
-}
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -766,3 +766,14 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
+
+/* Accessibility fix for contrast on Show Pages */
+/* Breadcrumb links: ensure #000000 overrides tenant link color for WCAG AA contrast */
+nav.breadcrumb a {
+    color: #000000 !important;
+}
+
+/* Document title links in gallery/list views: ensure #000000 overrides tenant link color for WCAG AA contrast */
+.document-title-heading a {
+    color: #000000 !important;
+}
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -769,7 +769,7 @@ dd {
 
 /* Accessibility fix for contrast on Show Pages */
 /* Breadcrumb links: ensure #000000 overrides tenant link color for WCAG AA contrast */
-nav.breadcrumb a {
+body.public-facing nav.breadcrumb a {
     color: #000000 !important;
 }
 
--- a/app/views/shared/_appearance_styles.html.erb
+++ b/app/views/shared/_appearance_styles.html.erb
@@ -229,4 +229,9 @@ body.community .community-explore-collections:hover .card-title a {
   color: <%= appearance.primary_button_text_color %>;
 }
 
+/* Accessibility fix for contrast on Show Pages */
+nav.breadcrumb a {
+  color: #000000 !important;
+}
+
 </style>
--- a/app/views/shared/_appearance_styles.html.erb
+++ b/app/views/shared/_appearance_styles.html.erb
@@ -229,9 +229,4 @@ body.community .community-explore-collections:hover .card-title a {
   color: <%= appearance.primary_button_text_color %>;
 }
 
-/* Accessibility fix for contrast on Show Pages */
-nav.breadcrumb a {
-  color: #000000 !important;
-}
-
 </style>


laritakr (changes requested):
Appearance styles doesn't seem the right place for this change. Normally things like this would be in the static css, not in the code that allows a tenant to customize their colors. 

If the contrast fix is site-wide, it should probably go into hyku.scss. If it's theme-specific, there are special css files per theme. 

If there's a reason it needs to be in the appearance_styles, feel free to offer additional detail.

trmccormick:
I moved it to hyku.scss based on your recommendation.

app/assets/stylesheets/hyku.scss
```diff
@@ -766,3 +766,14 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
+
+/* Accessibility fix for contrast on Show Pages */
+/* Breadcrumb links: ensure #000000 overrides tenant link color for WCAG AA contrast */
+nav.breadcrumb a {
+    color: #000000 !important;
```
laritakr, line 773: I'm curious why we need to override the breadcrumb color for links? Do we not need to change all links if it's a contrast issue?

app/assets/stylesheets/hyku.scss
```diff
@@ -766,3 +766,14 @@ dd {
...
+}
+
+/* Document title links in gallery/list views: ensure #000000 overrides tenant link color for WCAG AA contrast */
+.document-title-heading a {
```
laritakr, line 777: And on this one, is it ONLY the gallery view color? because this doesn't affect the list view.

laritakr (changes requested):
I'd like more information on what the goal is here, because changing a few random links so they don't respond to the appearance menu colors feels wrong. Why not change ALL links to have appropriate color by changing the overall default color used for links?

I'd be glad to chat briefly on the accessibility issues we need to resolve so we can get a solution that's more aligned with the full app's functionality. Come by office hour some week to look at it in more detail. 

I just think that the one-off overrides aren't the right way, and there are so many ways to control the look of the app that understanding the big picture would help figure out the best way.

app/assets/stylesheets/hyku.scss
laritakr: Screenshot 2026-06-23 at 11 22 55 AM

Screenshot 2026-06-23 at 11 22 59 AM

app/assets/stylesheets/hyku.scss
laritakr: And if it IS only breadcrumbs, then the help text on the appearance menu should be changed to not say that the affects breadcrumbs.

Screenshot 2026-06-23 at 11 26 52 AM

app/assets/stylesheets/hyku.scss
```diff
@@ -766,3 +766,14 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
+
+/* Accessibility fix for contrast on Show Pages */
+/* Breadcrumb links: ensure #000000 overrides tenant link color for WCAG AA contrast */
+nav.breadcrumb a {
+    color: #000000 !important;
+}
```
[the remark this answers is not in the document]
laritakr: The question is whether it SHOULD be overriding appearance styles. Appearance styles are intended to allow custom colors. It seems better if the site defaults themselves were set appropriately rather than overriding customization abilities.

app/assets/stylesheets/hyku.scss
```diff
@@ -766,3 +766,14 @@ dd {
...
+
+/* Document title links in gallery/list views: ensure #000000 overrides tenant link color for WCAG AA contrast */
+.document-title-heading a {
+    color: #000000 !important;
+}
```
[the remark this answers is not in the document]
laritakr: This suggestion seems logical, but the document-title-heading isn't used on every title, so it wouldn't actually work.

orangewolf:
@trmccormick can you circle back here? it would be nice to get this resolved for the next release

PR #3065, opened 2026-05-20
Put field registratations behind a gate

laritakr:
## Summary

Flexible metadata allows control of catalog controller field registration for various uses, but the hardcoded registration of fields in the CatalogController prevented control via the m3 profile. This commit moves field registration inside a gate to allow the m3 profile to have the control when using flexible behavior.


ShanaLMoore (approved):
nice catch

PR #3073, opened 2026-05-29
Update uploads directory cleanups to include more directories

maxkadel:
This cleans up tenants that have since moved to S3, allows you to pass in past directories, and deletes orphaned tenant directories (ones that were created without a tenant ID)

### DIFF
--- a/app/services/uploads_cleanup_service.rb
+++ b/app/services/uploads_cleanup_service.rb
@@ -63,7 +63,7 @@ class UploadsCleanupService
       return
     end
 
-    Rails.logger.debug "Enqueueing cleanup for #{tenant} → #{uploads_path}"
+    Rails.logger.info "Enqueueing cleanup for #{tenant} → #{uploads_path}"
     CleanupUploadFilesJob.perform_later(
       delete_ingested_after_days: @delete_ingested_after_days,
       uploads_path: uploads_path,
@@ -117,7 +117,7 @@ class UploadsCleanupService
       return
     end
 
-    Rails.logger.debug "Enqueueing orphaned tenant cleanup for #{tenant_id} → #{uploads_path} " \
+    Rails.logger.info "Enqueueing orphaned tenant cleanup for #{tenant_id} → #{uploads_path} " \
          "(all files older than #{@delete_all_after_days} days)"
     CleanupUploadFilesJob.perform_later(
       delete_ingested_after_days: @delete_all_after_days,
--- a/app/services/uploads_cleanup_service.rb
+++ b/app/services/uploads_cleanup_service.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+# Orchestrates per-tenant Hyrax upload staging cleanup by enqueueing
+# CleanupUploadFilesJob for every tenant that has local staging files.
+#
+# Two scenarios are handled:
+#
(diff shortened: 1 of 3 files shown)


app/services/uploads_cleanup_service.rb
```diff
@@ -0,0 +1,130 @@
...
+  def enqueue_tenant_cleanup(uploads_path, tenant)
+    return if @queued_paths.include?(uploads_path)
+
+    staging_dir = File.join(uploads_path, CleanupUploadFilesJob::CARRIERWAVE_SUBDIR)
+    unless Dir.exist?(staging_dir)
+      Rails.logger.debug "Skipping #{tenant}: no staging directory at #{staging_dir}"
+      return
+    end
+
+    Rails.logger.debug "Enqueueing cleanup for #{tenant} → #{uploads_path}"
```
aprilrieger: Right now the service uses Rails.logger.debug for the "Enqueueing cleanup for..." messages, which means we probably won’t see them in production logs by default. Since CleanupUploadFilesJob logs at info, it might make sense to use Rails.logger.info for the enqueueing messages as well to keep things consistent. I think the "Skipping" messages are fine staying at debug.

app/services/uploads_cleanup_service.rb
```diff
@@ -0,0 +1,130 @@
...
+  def enqueue_orphaned_cleanup(uploads_path, tenant_id)
+    return if @queued_paths.include?(uploads_path)
+
+    staging_dir = File.join(uploads_path, CleanupUploadFilesJob::CARRIERWAVE_SUBDIR)
+    unless Dir.exist?(staging_dir)
+      Rails.logger.debug "Skipping orphaned tenant #{tenant_id}: no staging directory at #{staging_dir}"
+      return
+    end
+
+    Rails.logger.debug "Enqueueing orphaned tenant cleanup for #{tenant_id} → #{uploads_path} " \
```
aprilrieger: Same comment as above here.

spec/services/uploads_cleanup_service_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+        before { allow(Dir).to receive(:exist?).with('/nonexistent/path').and_return(false) }
+
+        it 'skips gracefully without raising' do
+          expect { service.run }.not_to raise_error
+        end
+
+        it 'does not enqueue jobs for the missing path' do
+          service.run
+
+          expect(CleanupUploadFilesJob).to have_received(:perform_later).once # main loop only
```
aprilrieger: Nice fix on the flaky spec, I noticed this test is still using .once, while the earlier one switched to negative expectations. Might be worth updating this one too for consistency.

aprilrieger (dismissed):
Nice work on this @maxkadel 


I left a few non-blocking suggestions in the review notes:
- Consider bumping the "Enqueueing cleanup for..." log messages to info level for production visibility
- One spec (line 194) still uses .once in a fragile way similar to the flaky test you fixed

aprilrieger (dismissed):
thank @maxkadel

spec/services/uploads_cleanup_service_spec.rb
maxkadel: thanks - fixed

aprilrieger (approved):
Thanks Max :)

PR #3095, opened 2026-06-08
Pin json ~> 2.0

ShanaLMoore:
Knapsacks gitignore Gemfile.lock and re-resolve from scratch in CI, so unpinned gems can drift to incompatible majors (e.g. json 3.x).

Pinning json at the Hyku level keeps every downstream resolving to a compatible version without each one overriding via bundler.d.

ref: https://notch8.slack.com/archives/C08EHRPQ9MY/p1780929216187279

image

@samvera/hyku-code-reviewers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ gem 'i18n-tasks', group: %i[development test]
 gem 'iiif_print', '~> 3.0.12'
 gem 'jbuilder', '~> 2.5'
 gem 'jquery-rails' # Use jquery as the JavaScript library
+gem 'json', '~> 2.0' # avoid v3 breaking changes downstream
 gem 'json-canonicalization', "0.3.1" # The maintainers yanked 0.3.2 version (see https://github.com/dryruby/json-canonicalization/issues/2)
 gem 'json_schemer' # Required for m3 schema validation
 gem 'launchy', group: %i[test]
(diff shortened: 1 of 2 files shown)


maxkadel (approved):
Thanks @ShanaLMoore!

Gemfile
```diff
@@ -22,7 +22,7 @@ gem 'addressable', '2.8.1' # remove once https://github.com/postrank-labs/postra
 gem 'aws-sdk-sqs', group: %i[aws]
 gem 'bixby', group: %i[development test]
 gem 'blacklight', '~> 7.38' # Updated for Rails 7.2 compatibility
-gem 'blacklight_advanced_search'
+gem 'blacklight_advanced_search', '~> 7.0' # v8 introduces breaking changes
```
maxkadel: Looking into this more, I think I was wrong - I think it was an individual test for that knapsack that failed with the new version, but I think v8 should be non-breaking.

maxkadel (changes requested):
Shoot, I think I was wrong about blacklight_advanced_search needing to be pinned - I think [v8 should be compatible with BL 7](https://github.com/projectblacklight/blacklight_advanced_search/blob/0e22b8e57ae90f35a57be14d9794b16aeb052a8c/blacklight_advanced_search.gemspec#L19) - it was just a specific test that failed in that one knapsack. 

I do think we need the json pin, though.

Gemfile
ShanaLMoore: closed and replaced by https://github.com/samvera/hyku/pull/3096

PR #3096, opened 2026-06-08
Pin json ~> 2.0

ShanaLMoore:
Knapsacks gitignore Gemfile.lock and re-resolve from scratch in CI, so unpinned gems can drift to incompatible majors (e.g. json 3.x).

Pinning json at the Hyku level keeps every downstream resolving to a compatible version without each one overriding via bundler.d.

ref: https://notch8.slack.com/archives/C08EHRPQ9MY/p1780929216187279

image

@samvera/hyku-code-reviewers


maxkadel (approved):
Thanks @ShanaLMoore!

PR #3097, opened 2026-06-08
add a single tenant mode docker-compose file

orangewolf:
create a ci test make sure it stays running going forward. start solr fixes, but they need more work

@samvera/hyku-code-reviewers


hippietilley (commented):
I've gotten Hyku to come up. I'm still checking it out, but I think there are places in the code that still expect a tenant. I got this error after clicking settings > account in the sidebar. 

Image

hippietilley:
Additionally, the add work and add collection buttons aren't showing.

hippietilley:
> I've gotten Hyku to come up. I'm still checking it out, but I think there are places in the code that still expect a tenant. I got this error after clicking settings > account in the sidebar.
> Image

Oh! I checked the Site table in Adminer. Looks like a site is created, but it doesn't get an account or anything associated with it.

orangewolf:
> > I've gotten Hyku to come up. I'm still checking it out, but I think there are places in the code that still expect a tenant. I got this error after clicking settings > account in the sidebar.
> > Image
> 
> Oh! I checked the Site table in Adminer. Looks like a site is created, but it doesn't get an account or anything associated with it.

that's an excellent clue =-)

orangewolf:
@hippietilley 
I checked this and made a seed file change. The initial user was not being set to be an admin on the single tenant. Once I did that, I dropped everything and reloaded it (`docker compose -f docker-compose.single.yml down -v`) and I now get an admin and can access the the account settings page w/o error. I also see the create collection and work buttons. 

Your public schema `accounts` table should have a single entry that looks like this (random key will be different):
Screenshot 2026-06-17 at 17 44 17

and the keyed schema (`8af4e423-d953-4c9c-a649-395df74da681` in my case) should have a site that points to that account id. 

If you don't want to drop and re-up, you can look at the seed file to figure out which pieces need to be created to get all the right parts loaded in the db for single tenant. I would recommend removing any extra account objects and sites objects you might have as they will confuse the issue.

hippietilley:
It's still not working right for me. I removed the containers and volumes, pulled your changes, rebuilt the container and then brought up docker-compose.single.yml. I did have to change your `expose: -3000`, etc to `ports: -3000:3000`, but that's not really a problem and likely due to OS or docker differences. In Adminer this time the site in the table had a null value for `account_id`, but if I run `Sites.all` in the Rails console inside the web container the site (there's only one) has the `account_id` of the only account. After adding the `account_id` in Adminer I can load the account settings page, but I was getting the `@account.tenant` error from above before that. I still don't have buttons to add collections or works. The only roles in the roles table are superadmin and member. The only user has both.

I'm a bit confused about the relationships between Site, Account, Tenant, and User. I understand that Tenant comes from the Apartment gem and an Account has a Tenant and a Site has an Account, but how does User factor in? Account and User are usually synonymous in my experience so it's very unclear what the difference between them is supposed to be here.

Unrelated, but something I noticed... The worker container is querying the solr container for `good_jobs` every 1-3ish seconds. Is that intended behavior?

orangewolf:
Site - The part of the tenant data record kept within the schema for the specific tenant. not really clear to me that keeping this outside the account is useful, but hasn't been a big enough problem to refactor

Account - The part of the tenant data record that is kept in the public schema

Tenant - A concept of the Apartments gem - is signified by the Account in this application. 

User - Users are global (kept in the public schema outside of any given tenant). These are unrelated to Tenant/Account/Site and represent a person who can log in to the system in some capacity. Their Roles (which govern their permissions) are kept on a per tenant bases.

orangewolf:
@hippietilley I think I found it. Your comment on the ports made me think you were accessing the site as `localhost:3000` and not `hyku.localhost.direct`. That should work but was explicitly preventing the switch from happening. I tried to figure out why we were limiting the single tenant to root, but it was added 7 years ago by a dev who is no longer around to ask. I don't see anything that is effected by it, so I'm hoping this will be the one.

reallybigmess:
> @hippietilley I think I found it. Your comment on the ports made me think you were accessing the site as `localhost:3000` and not `hyku.localhost.direct`. That should work but was explicitly preventing the switch from happening. I tried to figure out why we were limiting the single tenant to root, but it was added 7 years ago by a dev who is no longer around to ask. I don't see anything that is effected by it, so I'm hoping this will be the one.

Popping in to confirm that the latest fix works for us now and things seem to be working with the initial admin user as well, which is great.

When I was experimenting on our fork I added 2 environment variables to .env -- HYKU_SINGLE_TENANT_NAME and HYKU_SINGLE_TENANT_CNAME, which sub in for name/cname during the calls to Account.new/Account.find_by/Account.from_cname in a few places. Would it be useful if I added those changes to this PR? Happy to throw it in comments or a commit somewhere and you can make the call. Thanks.

orangewolf:
> > @hippietilley I think I found it. Your comment on the ports made me think you were accessing the site as `localhost:3000` and not `hyku.localhost.direct`. That should work but was explicitly preventing the switch from happening. I tried to figure out why we were limiting the single tenant to root, but it was added 7 years ago by a dev who is no longer around to ask. I don't see anything that is effected by it, so I'm hoping this will be the one.
> 
> Popping in to confirm that the latest fix works for us now and things seem to be working with the initial admin user as well, which is great.
> 
> When I was experimenting on our fork I added 2 environment variables to .env -- HYKU_SINGLE_TENANT_NAME and HYKU_SINGLE_TENANT_CNAME, which sub in for name/cname during the calls to Account.new/Account.find_by/Account.from_cname in a few places. Would it be useful if I added those changes to this PR? Happy to throw it in comments or a commit somewhere and you can make the call. Thanks.

yep, feel free to either put it here or we can merge this one and add it to a new PR. Either works for me.

reallybigmess:
> yep, feel free to either put it here or we can merge this one and add it to a new PR. Either works for me.

Great thanks, I made a new PR from our fork - #3110

laritakr (approved):

PR #3101, opened 2026-06-12
Add a Work Type facet to the catalog and dashboard works page

ShanaLMoore:
## What

Adds a **Work Type** facet so you can narrow to a single work type:
- on the **catalog search page** (a sidebar facet section), and
- on the **dashboard Works page** (a dropdown in the filter row next to Visibility / Status / Admin Set).

The existing "Type" facet only distinguishes Work vs Collection vs Image, and "Resource Type" is depositor-entered metadata - neither answers "show me just the GenericWorks." Installations with several work types feel this immediately; this came out of the Enact build (five work types), where we shipped it as a knapsack override first and are contributing it back.

## How

- A `has_model_ssim` facet on `CatalogController` and in the dashboard works decorator's `configure_facets` (the dashboard filter row renders every configured facet as a dropdown, so declaring it is all it takes).
- A `work_type_facet_label` helper translates class names through each installation's `activerecord.models` locale entries, so the facet reads "Generic Work" rather than "GenericWorkResource". Two details baked in from debugging the knapsack version:
  - pluralized locale entries (`one:/other:`) are not resolved by `model_name.human`, so the lookup passes an explicit count;
  - unknown values fall back to titleized class names, so stale index data never breaks the page.
- The dashboard addition is guarded in case the facet was already copied from the catalog config.

## Testing

Helper specs included. Verified functionally in a running Hyku app (the Enact instance) with the equivalent knapsack code: facet appears in both places with counts, filtering narrows results, and the constraint chip shows the human name.

image

image

image

image

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -125,6 +125,7 @@ class CatalogController < ApplicationController
     #   The ordering of the field names is the order of the display
     config.add_facet_field 'generic_type_sim', label: "Type", limit: 5
     config.add_facet_field 'resource_type_sim', label: "Resource Type", limit: 5
+    config.add_facet_field 'has_model_ssim', label: "Work Type", helper_method: :work_type_facet_label, limit: 5
     config.add_facet_field 'creator_sim', limit: 5
     config.add_facet_field 'contributor_sim', label: "Contributor", limit: 5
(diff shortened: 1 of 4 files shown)


laritakr (approved):
Looks good. Elsewhere it was done via human_readable_type_sim but I think that term is obsolete.

PR #3115, opened 2026-06-25
Adds concurrency group and cancel-in-progress

laritakr:
## Summary

Adds concurrency group and cancel-in-progress to the CI workflow. The intent is if a PR or branch is updated with new commits then any specs running will be cancelled.


maxkadel (approved):
Looks good, thanks!

PR #3122, opened 2026-07-06
Add demo tenant reset service for public demo (sandbox) tenants

SixtyGrit:
## Summary

Adds a `DemoTenantResetService` plus `hyku:demo:snapshot[tenant]` and `hyku:demo:reset[tenant]` rake tasks so that an account flagged `public_demo_tenant` (#2896) can be reset to a stored golden state on a schedule. This is the reset automation the sandbox tenant concept promised: a publicly writable demo tenant is only safe to operate when visitor changes have a bounded lifetime.

## Motivation

This work comes out of making public Hyku demo environments safe to leave unattended. The `public_demo_tenant` flag from #2896 anticipated a daily reset that was never built. With shared credentials published for a sandbox tenant, visitors can upload junk, rebrand the tenant, and create users. All of that should be swept nightly and the tenant returned to a curated golden state.

## What changed

* New `DemoTenantResetService` with two operations, both refusing to run for accounts not flagged `public_demo_tenant`:
  * `snapshot!` captures the golden state onto the account: Site appearance attributes, all content blocks, and featured work identifiers. Capturing is an explicit operator step so a vandalized tenant is never accidentally recorded as golden.
  * `reset!` destroys all works, file sets, collections, featured entries, and Bulkrax importers and exporters in the tenant; removes visitor-created users; restores the snapshot; optionally re-imports a seed corpus through the Bulk Import pipeline (Bulkrax CSV parser); restores featured works; runs an injectable health check; and stamps `accounts.last_reset_at` on success.
* Visitor-user handling respects Hyku's data model: users are global records while role grants live in each tenant schema, so the sweep targets users holding roles in this tenant (self-registered users receive a registered-group membership role on creation). Superadmins, tenant superadmins, and a configurable keep list survive. Swept users lose their role grants here and their global record is destroyed only when they hold roles in no other tenant.
* The seed source is configuration (a CSV path via `DEMO_SEED_CSV_PATH` or the service parameter), so core carries no dependency on any particular corpus. With no seed path the reset produces an empty but fully branded tenant.
* Import draining is GoodJob-aware: it waits for queue quiescence and pulls Bulkrax's deferred relationship jobs forward so collection membership is written before the reset returns. On other queue adapters the reset logs a warning that background ingest settles after it returns.
* Migration adds `accounts.last_reset_at` (datetime) and `accounts.demo_tenant_snapshot` (jsonb).

## Placement rationale

samvera/hyku core. The `public_demo_tenant` flag this builds on is core, and any institution or service provider running a public sandbox needs a reset, not only one vendor's demo. Nothing here is deployment specific: the seed corpus, kept users, import owner, and health check are all injected configuration.

## Specs

Spec-first history (the first commit contains only failing specs):

* Non-flagged accounts refuse both `snapshot!` and `reset!`.
* `reset!` refuses to run without a stored snapshot.
* Wipe happens before restore; the seed import runs only when a path is configured.
* The health check receives the account; a falsey return fails the reset; `last_reset_at` is stamped only on success.
* Import verification raises when the importer run recorded failures or recorded no run.
* A full integration example provisions a real tenant via `CreateAccount`, snapshots a golden state, vandalizes the tenant (site rename, content block edits, a new work, a new collection, a new user), resets, and proves: visitor artifacts are gone, the golden state is restored, superadmin and kept seed users survive with the visitor user destroyed, and a second reset is idempotent.
* Rake task specs cover argument validation, unknown tenants, refusal for non-demo accounts, and service delegation.

## Notes for reviewers

* Resource deletion uses `Hyrax.persister.delete` plus index deletes with a full index purge and admin set reindex as backstop. This intentionally favors the proven fast path; file storage reclamation for derivatives is listed as a follow-up rather than blocking the reset.
* `attr_readonly :public_demo_tenant` means the guard cannot be bypassed by flipping the flag on an existing account.
* Restoring CarrierWave-mounted Site columns uses `update_columns` because assigning a filename string through the setter would be treated as a new upload.

## Follow-ups

* A scheduling wrapper (nightly job per flagged tenant) and failure surfacing land separately so this PR stays reviewable.
* File storage reclamation for orphaned derivatives after repeated resets.


app/services/demo_tenant_reset_service.rb
```diff
@@ -0,0 +1,424 @@
+# frozen_string_literal: true
+
+require 'timeout'
```
orangewolf: requires do not belong here

app/services/demo_tenant_reset_service.rb
```diff
@@ -0,0 +1,424 @@
...
+  def guard_demo_tenant!
+    return if account&.public_demo_tenant?
+
+    raise NotDemoTenant, "#{account&.cname || account.inspect} is not flagged public_demo_tenant; refusing"
+  end
+
+  # AccountElevator.switch! (rather than account.switch) is required: it
+  # switches the Apartment schema so Valkyrie queries, Site.instance, and
+  # Bulkrax's multitenant paths are scoped to the tenant.
+  def within_tenant
```
orangewolf: there is already a block version of this that can be used. within_tenant implementation is unnecessary added complexity... plus it sets the tenancy two different ways 3 lines apart ;-)

app/services/demo_tenant_reset_service.rb
```diff
@@ -0,0 +1,424 @@
...
+  end
+
+  def keep_user?(user)
+    user.superadmin? || user.tenant_superadmin? || keep_emails.include?(user.email.to_s.downcase)
+  end
+
+  # Removes the users_roles join rows in this tenant's schema; the shared
+  # Role records themselves are left in place.
+  def strip_tenant_roles!(user)
+    user.roles.reload
```
orangewolf: this isn't how roles should be removed. there is a remove roles method that deletes the records instead of just leaving them in the db

app/services/demo_tenant_reset_service.rb
```diff
@@ -0,0 +1,424 @@
...
+    user.roles = []
+    log "stripped tenant roles from #{user.email}"
+  end
+
+  def destroy_user_unless_active_elsewhere!(user)
+    return if roles_in_other_tenants?(user)
```
orangewolf: this is really expensive in a large install. is it really nessisary to clean up the users in this way? might be better to skip it. its low risk. Or maybe tag the users created for a demo tenant and use that to select and remove them? either way, loading every tenant one at a time to check the roles per user is not going to scale well.

lib/tasks/demo.rake
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
```
orangewolf, line 1: this is great, it puts most of the logic in the service and keeps the rake task thin.

orangewolf (changes requested):
This is pretty solid though I think from an archtecture perspective it's short sided. It over prescribes the use case (only for demo tenants) instead of building a generic tool that people can use for various use cases and then capturing the specific use part in the recurring jobs. But I wouldn't stop this getting merged for that reason, just want to point it out as a place to open up design thinking more broadly in the future.

ShanaLMoore:
Picking this up while Nick is out. All four inline comments are addressed in 0a49c5cf, and I merged main in (263b2c04) since the branch was three weeks behind and would not bundle against a current image.

**`require 'timeout'`** removed.

**`within_tenant`** now uses the block forms rather than the hand-rolled previous/ensure. One thing worth flagging from doing it: `AccountElevator.switch!` and `Account#switch` are not two ways of doing the same thing. `AccountSwitch#switch!` only calls `Apartment::Tenant.switch!`, and `Account#switch!` only swaps the Solr, Fedora, Redis and DataCite endpoints plus the routing host. So the previous code moved the schema but left the endpoints wherever they already were, meaning the Solr purge was not provably scoped to the tenant. It is now nested, so both move and both restore themselves:

```ruby
def within_tenant(&block)
  account.switch { Apartment::Tenant.switch(account.tenant, &block) }
end
```

**Role removal** goes through `remove_role(role.name, role.resource)`, matching `User#site_roles=`, `Site`, and `Hyrax::Group`. Passing the resource covers both shapes a tenant user carries, the Site-scoped roles and the Hyrax::Group memberships.

**The visitor user cleanup is gone entirely**, per your suggestion. Users are global records while role grants live per schema, so answering "is this user active elsewhere" genuinely requires visiting every tenant, and there is no cheaper correct version. Visitors now keep their account and simply lose their grants in this tenant. That is a behavior change rather than a refactor, so: after a reset, a visitor's User record survives with no roles instead of being destroyed. An account with no grants has no access, so the practical effect is the same and the cost is bounded.

On the architecture point: agreed, and I do not think it is a small change. The shape I would want is a generic `TenantResetService` holding the mechanics, taking a policy object that supplies the guard, the keep rules, and the snapshot scope, with `DemoTenantResetPolicy` as the first implementation. Scheduling stays in the recurring job from #3123, which is the genuinely demo-specific part.

I would rather do that as a follow-up than grow this PR, since #3123 and #3124 both base on this branch and have been idle since July 13, and a rename plus extraction forces both to rebase before they can move. Filed as #3184 with the full sketch.

Specs and rubocop are green locally: 32 examples, 0 failures across `demo_tenant_reset_service_spec.rb` and `rake_spec.rb`, no offenses on the two changed files. The integration example that runs a real reset against a provisioned tenant is not stubbed, so it exercises the new `within_tenant` and the role stripping for real.

ShanaLMoore:
Verified the `within_tenant` change end to end, since the endpoint scoping was the part I was least willing to take on faith and the specs can't cover it (single tenant in the test env).

Snapshotted a throwaway `public_demo_tenant`, vandalized it, reset it, with four unrelated seeded tenants as a control:

```
BASELINE      ellison=45  harborview=45  meridian=20  sandbox=12
VANDALIZED    works=1  site="HACKED"  visitor_admin=true
RESET_RETURNED true
AFTER         works=0  site="Golden Resettest"  announcement="golden announcement"
VISITOR       present=true  roles=[]  admin=false
AFTER_CORPUS  ellison=45  harborview=45  meridian=20  sandbox=12
```

`Hyrax::SolrService.wipe!` stayed scoped to the target: the four control tenants are unchanged. Site attributes and content blocks restored from the snapshot, and the visitor's account survives with no roles, which is the behavior change from dropping the user destroy.

orangewolf (approved):
Thank you for the clean up!

PR #3123, opened 2026-07-06
Schedule nightly demo tenant resets

SixtyGrit:
## Summary

Schedules the demo tenant reset (#3122) nightly for every account flagged `public_demo_tenant`, with failure surfacing that does not disappear into the default retry handler. Built on the repo's existing recurrence idiom: a self-rescheduling ActiveJob registered through `Account#find_or_schedule_jobs`.

Stacked on #3122 (the reset service); only the last two commits are new here.

## Motivation

A reset service is only an operational answer if it runs unattended. Hyku has no cron system; recurring work is self-rescheduling jobs (EmbargoAutoExpiryJob, LeaseAutoExpiryJob, TenantMaintenanceJob) seeded per tenant by `Account#find_or_schedule_jobs`. Registering the reset job there means flagged tenants get the nightly chain at creation time, at application boot, and re-seeded by the daily TenantMaintenanceJob pass if the chain ever breaks. No new scheduling infrastructure is introduced.

## What changed

* New `DemoTenantResetJob`, shaped like EmbargoAutoExpiryJob: it runs `DemoTenantResetService#reset!` for the current tenant, then re-enqueues itself for tomorrow midnight. Accounts that lose the flag end the chain cleanly.
* `Account#find_or_schedule_jobs` appends the job when `public_demo_tenant?`, following the same gate pattern as BatchEmailNotificationJob.
* Failure surfacing: ApplicationJob's `retry_on` swallows final exceptions in an empty block. This job overrides it with three attempts, an error log line carrying the tenant identifier, and a `Sentry.capture_exception` guarded by `defined?(Sentry)` (matching the convention in the GoodJob initializer, where Sentry is supplied by a knapsack rather than this bundle). On permanent failure the tomorrow chain is deliberately not re-enqueued: the daily maintenance pass re-seeds it, and the stale `last_reset_at` (stamped only on success, from #3122) keeps the failure visible.
* Deployment configuration is environment variables (`DEMO_SEED_CSV_PATH` with a `%{tenant}` placeholder, `DEMO_KEEP_USERS`, `DEMO_IMPORT_USER`, `DEMO_HEALTH_CHECK`), so multi-tenant deployments can wire per-tenant seed corpora without code changes.

## Placement rationale

samvera/hyku core, same reasoning as #3122: the `public_demo_tenant` flag is core and any public sandbox needs its reset to run unattended. The job carries no deployment specifics.

## Specs

Spec-first history. Coverage: the job runs the service and reschedules for tomorrow midnight; non-flagged tenants end the chain without running the service; a failed reset does not chain tomorrow's run; the `%{tenant}` seed-path templating expands the account name; `find_or_schedule_jobs` registers the job for flagged accounts, skips it when one is already queued, and never registers it for regular accounts.

## Notes for reviewers

* One reviewer-visible behavior inherited from the idiom: when no reset job is queued for a flagged tenant, the next `find_or_schedule_jobs` pass enqueues one immediately (so the first reset after flagging, and the first after a deploy that lost the queue, runs right away rather than at midnight). That matches how the embargo and lease chains behave today.
* All recurring jobs fire at midnight application time; a reset with a seed re-import will compete with the other nightly jobs on single-worker deployments. Operators can size the worker or accept the serialization; the reset drains its own import jobs before returning.

## Follow-ups

* The demo banner (next PR in the series) reads `last_reset_at` to display reset recency to visitors.


orangewolf (commented):
this one looks good - it should be held for #3122 though so I'm not hitting approve yet

ShanaLMoore:
Base is updated. Picking this up while Nick is out.

#3122 now has all four of your inline comments addressed, plus a reply to the architecture note and a follow-up issue at #3184. This branch has been merged forward onto that revised base, which also brought in 40 commits from main since the old base would not bundle against a current image.

No source changes here. Nothing in the job needed them, and it calls only the public `DemoTenantResetService#reset!`, so the rename of a private method in #3122 does not reach it. Merge was clean with no conflicts.

Green locally: 89 examples, 0 failures across `demo_tenant_reset_job_spec.rb` and `account_spec.rb`, and no rubocop offenses on `demo_tenant_reset_job.rb` or `account.rb`.

One thing worth noting given your architecture comment on #3122: this job is arguably already the shape you were describing. Every demo-specific knob lives here rather than in the service, `DEMO_SEED_CSV_PATH`, `DEMO_KEEP_USERS`, `DEMO_IMPORT_USER`, and `DEMO_HEALTH_CHECK`. So the remaining generalization work is extracting a policy object out of the service, not relocating scheduling, which is what #3184 proposes.

Your hold was on #3122, so this should be unblocked now whenever you want to take another look.

orangewolf (approved):

PR #3124, opened 2026-07-06
Add a demo environment banner for public demo tenants

SixtyGrit:
## Summary

Renders a dismissible banner on every page of a `public_demo_tenant` tenant identifying it as a demo environment, showing the running Hyku version, the reset schedule with the last reset time, and an optional acceptable use link. Stacked on #3122 (which adds `last_reset_at`); only the last two commits are new here.

## SCREENSHOT

<details>
image
</details>

## Motivation

An honest public demo tells visitors what they are looking at. Peers in this space title their demo pages with the running release or banner the version; a sandbox that visitors can write to also needs to say when their changes will disappear and what use is acceptable. Version transparency doubles as demo currency pressure: a stale version string is visible to everyone.

## What changed

* New `shared/_demo_banner` partial, self-guarded on `current_account&.public_demo_tenant?`, rendered from the base `layouts/hyrax`, the three theme overrides of that layout (community, cultural repository, institutional repository), and a new override copy of the Hyrax dashboard layout so admin pages are covered too. Proprietor, homepage, and Devise pages inherit through `layouts/hyrax` automatically.
* Content: a "Demo environment" heading, "Running Hyku vX.Y.Z" from `Hyku::VERSION`, "Content and settings reset nightly" with "Last reset: {time}" (or a not-yet-reset message), and an acceptable use link when configured. All strings are i18n keys under `hyku.demo_banner`.
* Configuration: one new per-tenant setting, `demo_acceptable_use_url`, through the AccountSettings concern, which gives the admin form field and the `HYKU_DEMO_ACCEPTABLE_USE_URL` environment fallback for free. Everything else renders from the flag with zero configuration.
* Dismissal: Bootstrap 4 alert dismiss plus a small script that remembers the dismissal in a cookie for 24 hours. Dismissal is deliberately client-side only so pages remain cacheable by CDNs; the banner returns after the cookie expires, roughly matching the nightly reset cadence.
* Accessibility: the banner sits after the skip-to-content link so tab order is preserved, is a labelled region rather than a live alert, and the dismiss control carries an aria-label.

## Placement rationale

samvera/hyku core. The banner renders only for accounts flagged with the core `public_demo_tenant` flag and shows nothing anywhere else; any institution operating a public sandbox benefits. Nothing is provider specific.

## Specs

Spec-first history. View specs cover: banner presence with the running version, reset schedule text, accessible dismiss control, not-yet-reset and last-reset states, the acceptable use link present and absent, and empty rendering for regular tenants. Feature specs cover presence on the home page and the sign-in page of a flagged tenant and absence on a regular tenant.

## Notes for reviewers

* The dashboard layout override is a copy of the gem template at the Gemfile.lock-pinned Hyrax revision with one added render line, following the repo's OVERRIDE comment convention. It carries the usual drift cost of copied gem templates; happy to move it if there is a better hook for dashboard-wide content.
* Knapsack applications that override `layouts/hyrax` themselves will need to add the one-line render to pick up the banner.
* English locale only in this PR; the sibling locale files can follow.

## Follow-ups

* The acceptable use page itself is deployment content; the URL setting stays blank until one exists.


orangewolf (commented):
looks good. hold for #3122

ShanaLMoore:
Base is updated. Picking this up while Nick is out.

#3122 now has all four of your inline comments addressed, plus a reply to the architecture note and a follow-up at #3184. This branch is merged forward onto that revised base, which also brought in 40 commits from main. Clean merge, including `config/locales/en.yml` and `application.js`, which had both moved on main.

No source changes here, same as #3123. Nothing in the banner touches the reset service; it reads `last_reset_at` off the account.

Green locally: 11 examples, 0 failures across `_demo_banner.html.erb_spec.rb` and `demo_banner_spec.rb`, including the three feature specs, and no rubocop offenses.

Rendered against a real flagged tenant at 1440px:

> **Demo environment** Running Hyku v7.1.0. Content and settings reset nightly. Last reset: August 06, 2026 20:13. ✕

Worth noting that timestamp is real, not fixture data. It is the `last_reset_at` written by #3122's service during an end to end reset on the same tenant, so this is the banner reading a value the reset actually produced.

Your hold was on #3122, so this should be unblocked whenever you want another look.

One separate thought for whoever sequences these: this PR does not depend on the reset service at all, only on the `last_reset_at` column from #3122's migration. If it is useful to land the banner sooner, that two-line migration could be split into its own PR off main and this rebased onto it, taking it out of the chain entirely.

orangewolf (approved):

PR #3125, opened 2026-07-06
Repair the sample seed task: fix the completion summary crash and add a visibility argument

SixtyGrit:
## Summary

`db:seed:sample:create[tenant,type,quantity]` has two defects that make it unreliable for standing up evaluation content. First, the Valkyrie path always crashes after creating and indexing every record, because `print_completion_summary` requires an OER argument that `Sample::ValkyrieService` never passes. Second, there is no way to control the visibility of the seeded records, and on the Valkyrie path the intended open visibility never actually takes effect, so every seeded work and collection comes out restricted. This PR fixes the crash and adds an optional fourth task argument, `visibility`, that is honored on both backends. When the argument is omitted, each backend behaves exactly as it did before.

## Motivation

This came out of work to make public Hyku demo environments safe and useful to operate. A demo tenant needs seeded content that anonymous visitors can actually see. Running the seed task with `type=valkyrie` today produces a tenant full of restricted works, and the task itself exits with an `ArgumentError` after doing all of its work, which makes it look like seeding failed entirely. Both problems affect anyone using the sample seed tooling in Hyku core, not just demo installations.

## What changed

- `Sample::SharedMethods#print_completion_summary` now takes `oers:` as a keyword argument defaulting to an empty array. The ActiveFedora service passes its OERs through the keyword; the Valkyrie service, which creates no OERs, calls the method with four positional arguments and reports "Created 0 OERs". This removes the arity crash at the end of the Valkyrie run.
- `Sample::SharedMethods#initialize` accepts an optional third parameter, `visibility` (`open`, `authenticated`, or `restricted`), stored alongside the existing `tenant_name` and `quantity`. A `VALID_VISIBILITIES` constant is derived from the `Hydra::AccessControls::AccessRight` values.
- The rake task signature is now `db:seed:sample:create[tenant,type,quantity,visibility]`. The argument is optional, is validated against the allowed values with a usage error on anything else, and is threaded into both service constructors. Usage text is updated.
- On the Valkyrie path, visibility is now applied where it can actually take effect. The previous code passed a `visibility:` key into `CollectionResource.new` and the work constructors, but `Hyrax::Resource` does not declare visibility as a Valkyrie attribute, so the constructor silently discards the key and no ACL is ever created. The inert hash keys are removed, and a new `apply_visibility` step assigns visibility through the resource's visibility writer after persistence and saves the resulting ACL, before the resource is indexed. This is the same idiom Hyrax's own shared factories use (`Hyrax::VisibilityWriter` followed by `permission_manager.acl.save`).
- On the ActiveFedora path, where mass assignment of `visibility` has always worked, the three hardcoded open-visibility attributes now defer to the requested visibility and fall back to open when none is requested.

## Placement rationale

This belongs in samvera/hyku core rather than in a downstream knapsack. The sample seed services (`app/services/sample/`) and the rake task (`lib/tasks/sample_seeds.rake`) live in core, and both defects are reproducible in a stock Hyku checkout with no downstream customization. The arity crash affects every Valkyrie-type run of a core task, and the visibility gap is a consequence of how core Hyrax Valkyrie resources handle constructor attributes. A knapsack override could patch around either problem, but that would fork core tooling to fix bugs that every adopter of the core task inherits. Nothing in this change is specific to any particular deployment.

## Specs

Spec-first commit history: the first commit contains only the specs, which fail against the previous implementation; the second commit makes them pass.

- `spec/services/sample/valkyrie_service_spec.rb` (new)
  - `#create_sample_data` completes and prints the completion summary without raising, with the heavy collaborators stubbed out. This is the regression test for the arity crash: on the previous code the run ends in `ArgumentError (given 4, expected 5)`.
  - With no visibility requested, a seeded work reloaded from the query service reports `restricted` visibility, pinning the default behavior.
  - With `open` requested, a seeded work reloaded from the query service reports `open` visibility, the persisted ACL grants read access to the `public` group, and a seeded collection likewise reloads as `open`.
- `spec/tasks/rake_spec.rb` (extended)
  - The task threads tenant, quantity, and visibility into `Sample::ValkyrieService.new` and `Sample::ActiveFedoraService.new`.
  - Omitting the fourth argument passes `nil`, which preserves prior behavior.
  - An unrecognized value such as `public` prints an error and never constructs a service.

## Notes for reviewers

- The default when no visibility is passed is deliberately "whatever each backend did before" rather than a single hardcoded value. The two backends did not behave the same way: ActiveFedora seeds have always come out open because the model honors mass-assigned visibility, while Valkyrie seeds have always come out restricted because the constructor drops the key. Forcing a shared default either would have silently opened existing Valkyrie-based workflows or silently closed existing ActiveFedora-based ones. Operators who want a specific result should now say so explicitly, for example `db:seed:sample:create[demo,valkyrie,50,open]`.
- `apply_visibility` runs between `Hyrax.persister.save` and `Hyrax.index_adapter.save` so that the read access groups are present in the ACL before the resource is indexed and `read_access_group_ssim` lands in Solr.
- Valkyrie collections get their visibility through the same `apply_visibility` step. `Sample::PermissionTemplateService.create_for_valkyrie_collection` still creates the permission template and grants but does not apply ACLs, matching its existing behavior; direct visibility assignment covers the anonymous-read case this task cares about.
- The rake task output strings stay as `puts` text, matching the existing operator-facing conventions of this task. No locale changes are needed.

## Follow-ups

- Known limitation, to be filed as a separate scoped issue: the seed services swap `ActiveJob::Base.queue_adapter` to an in-process async adapter for the duration of the run and restore it immediately after the summary prints. Characterization and derivative jobs that Hyrax enqueues downstream of file attachment land on that in-process adapter and are lost when the rake process exits, so seeded file sets never get characterization metadata or derivatives. Fixing that properly (leaving the tenant's real adapter in place, or draining the queue before exit) is out of scope for this PR.
- The seed services assume the tenant has at least one user (`User.first`) and raise otherwise. A friendlier guard with a clear error message would be a small quality-of-life improvement.


app/services/sample/shared_methods.rb
```diff
@@ -124,7 +136,7 @@ def restore_job_configuration
     end
 
     # rubocop:disable Metrics/AbcSize
-    def print_completion_summary(collections, images, generic_works, oers, total)
+    def print_completion_summary(collections, images, generic_works, total, oers: [])
```
orangewolf: if OER is needed here then ETD is likely also needed. maybe more importantly it would be better to refactor to send in a work_types hash with type name and number `{generic_works: 5, images: 2, oers: 0}` so that it will work regardless of work types available in a Hyku instance. then make the whole thing drive from the available_work_types logic.

orangewolf (changes requested):

ShanaLMoore:
Picking this up while Nick is out. Took the refactor rather than just adding ETD, since your second point is the one that actually fixes it. Also merged main in, the branch was 40 commits behind.

The type list now comes from `Site.instance.available_works`, falling back to `Hyrax.config.registered_curation_concern_types`. `quantity` still means "this many of each type", and an optional `work_types` hash overrides per type. An unrecognized type raises with the available list rather than seeding nothing.

`create_images`, `create_generic_works`, and `create_oers` are gone, replaced by one loop over the resolved counts that reuses the existing `build_work`. Oer's extra required fields (`audience`, `education_level`, `learning_resource_type`, `discipline`) moved to an `extra_attributes_for` hook, which is now the only place a type is named, and it encodes a real per-model requirement rather than a hardcoded list.

Both summaries collapsed onto one hash-taking printer, so the `oers:` keyword and two `rubocop:disable Metrics/AbcSize` comments are gone.

Two latent bugs fell out of driving cleanup from the same list:

* `ActiveFedoraService#create_sample_data` created OERs, but `clean_sample_data` only cleaned Collection, Image, GenericWork, and FileSet. Seeded OERs were never removed.
* `ValkyrieService` had no `create_oers` at all, so it could not seed OERs or ETDs even where the tenant offered them.

The rake task takes per-type counts in the quantity slot, semicolon separated because rake claims the comma:

```
rake db:seed:sample:create[tenant,valkyrie,50,open]
rake db:seed:sample:create[tenant,valkyrie,'generic_work:5;image:2',open]
```

Verified against a scratch tenant that offers all four types:

```
available:  ["GenericWork", "Image", "Etd", "Oer"]
quantity=2: {"GenericWork"=>2, "Image"=>2, "Etd"=>2, "Oer"=>2}
explicit:   {"GenericWork"=>5, "Image"=>2}
unknown:    Unknown work type 'nope'. Available: GenericWork, Image, Etd, Oer
```

Ran create and clean end to end on that tenant; cleanup reported all four types and left it empty. 24 examples, 0 failures, no rubocop offenses.

One judgement call worth flagging: when per-type counts are given there is no number left for collections, so it creates one, and the usage text says so. Happy to change that if you would rather it took a `collection:N` key.

orangewolf (dismissed):
Discussed on call 8/7/2026

app/services/sample/shared_methods.rb
```diff
@@ -2,12 +2,28 @@
 
 module Sample
   module SharedMethods # rubocop:disable Metrics/ModuleLength
-    attr_accessor :admin_set, :user
-    attr_reader :tenant_name, :sample_files_dir, :sample_data, :quantity
+    VALID_VISIBILITIES = [
```
orangewolf: these should not be defined here. They are available by calling `Hyrax::VisibilityMap.instance.visibilities`

lib/tasks/sample_seeds.rake
```diff
@@ -4,30 +4,62 @@ namespace :db do
...
         type = args[:type] || 'activefedora'
+        visibility = args[:visibility]&.downcase
 
-        case type.downcase
-        when 'activefedora', 'af'
-          Sample::ActiveFedoraService.new(args[:tenant], quantity).create_sample_data
-        when 'valkyrie', 'val'
-          Sample::ValkyrieService.new(args[:tenant], quantity).create_sample_data
-        else
+        if visibility && !Sample::SharedMethods::VALID_VISIBILITIES.include?(visibility)
```
orangewolf:
```suggestion
        if visibility && !Hyrax::VisibilityMap.instance.visibilities?(visibility)
```

orangewolf (changes requested):

ShanaLMoore:
@orangewolf picking this one up from @SixtyGrit. Both remaining comments addressed, and the branch is caught up with main.

**Visibility list** (421eb7bc): `VALID_VISIBILITIES` is gone from `Sample::SharedMethods`; the rake task asks `Hyrax::VisibilityMap.instance.visibilities` instead, and its error message now lists whatever the map holds rather than three literals.

One deviation from the suggestion, flagging it rather than burying it: `VisibilityMap` has no `visibilities?` predicate on the pinned hyrax (6fac65d), only `visibilities`, so this uses `visibilities.include?`. Same behavior, and `DEFAULT_MAP`'s keys are exactly the three values the constant used to hold.

**Merge with main** (8df26386): #3239 had landed a stopgap on the same lines, an empty `oers` array passed to `print_completion_summary` only to satisfy its arity, with a comment saying as much. This branch removes the need for it by driving both creation and cleanup from `work_type_counts`, so the branch side wins those four hunks. #3239's other additions are untouched: `random_visibility`, `visibility_pool` with its ENV knobs, and the collection ACL save.

Note the specs have not been run against this; the local stack is not up. Worth a CI pass before merge.

PR #3127, opened 2026-07-06
Containment hardening for public demo tenants

SixtyGrit:
## Summary

This PR closes three containment gaps on accounts flagged as `public_demo_tenant`. First, the last site-scoped superadmin on a public demo tenant can no longer be removed, whether through the proprietor UI, the remove-admin dropdown, or a direct PUT to the accounts controller. Second, the Invite User flow on the Manage Users page now authorizes against the existing `manage :tenant_controls` ability on public demo tenants, so only tenant superadmins can invite users there; standard tenants are unchanged. Third, the proprietor account edit form now renders the immutable `public_demo_tenant` flag as read-only text with an explanatory hint instead of a form control.

## Motivation

This work arose from an effort to make public Hyku demo environments safe to operate. A public demo tenant hands the admin role to anonymous visitors so they can explore the full admin experience, and a nightly reset restores the tenant to its baseline. That model only holds if a visitor with the admin role cannot lock the operator out or reach beyond the sandbox. Three gaps undermined that:

1. `Site#superadmin_emails=` applied removals with no floor. Assigning an empty list, which the proprietor forms post as a lone hidden blank field, silently stripped the final site-scoped superadmin. A UI-level guard existed in `Proprietor::AccountPresenter#can_remove_superadmin?`, but the remove-admin form in the same dropdown posts a superadmin email list without the removed user, and a direct PUT bypasses the presenter entirely. Once the last superadmin is gone, nobody inside the tenant can administer the demo controls anymore.
2. The invite endpoint (`Hyku::InvitationsController#create`, via devise_invitable) only checked `grant_admin_role` when the requested role was admin. A plain admin on a public demo tenant, which is to say any visitor, could invite arbitrary email addresses and hand out roles, turning the sandbox into a spam vector.
3. The account edit form rendered the `public_demo_tenant` checkbox even though the attribute is `attr_readonly` and not permitted on update, which invited confusion about whether the flag is editable.

One deliberate scope choice: appearance settings intentionally remain open to plain admins on demo tenants. Visitors branding the sandbox is part of the demo experience, and the nightly reset restores it. This was considered during the hardening pass and left open on purpose.

### Last-superadmin floor (model plus controller)

* `Site#superadmin_emails=` now calls a guard before applying removals. When the site belongs to a public demo tenant and the requested email list would leave the tenant with no site-scoped superadmins (empty list, blank-only list, or replacement emails that match no existing users), the setter raises `Site::LastSuperadminRemovalError`. A raise, rather than a validation error, is required here because the setter mutates roles immediately at attribute-assignment time, before save; a validation added after the fact would report failure while the removal had already happened.
* Swapping the last superadmin for another existing user in a single assignment still works, since the guard checks the resulting set rather than the removals alone. Standard tenants keep their current behavior, including clearing all superadmins, which matches the existing presenter logic that only protects demo tenants.
* `Proprietor::AccountsController#update` rescues the error, adds a `superadmin_emails` error to the account (message via the new `activerecord.errors.messages.cannot_remove_last_superadmin` key), and re-renders the edit form (HTML) or returns 422 (JSON).
* `Proprietor::AccountPresenter#can_remove_admin?` now also consults `can_remove_superadmin?`, because removing a user's admin role posts a superadmin email list without that user. Previously the remove-admin button was enabled in exactly the case the model now refuses.

### Invite gate (controller plus view)

* `Hyku::InvitationsController` gains a `before_action` on `create` that runs `authorize! :manage, :tenant_controls` when the current account is a public demo tenant. `Hyrax::Ability::TenantControlAbility` already denies `:tenant_controls` to plain admins on flagged tenants and grants it to tenant superadmins, so the gate reuses the existing ability without new symbols. Denial follows the app-wide CanCan rescue and redirects rather than erroring.
* The invite card on `hyrax/admin/users/index` hides on public demo tenants unless the user holds `manage :tenant_controls`, matching the delete and activate buttons further down the same page.
* The gate is conditional on the demo flag rather than unconditional, on purpose: user managers on standard tenants can invite users today without holding `:tenant_controls`, and an unconditional gate would silently remove that ability. Standard-tenant behavior is fully unchanged.

### Read-only flag on the edit form

* `proprietor/accounts/edit` replaces the `public_demo_tenant` input with a read-only text display (Yes/No plus a hint that the setting is fixed at creation), with strings in `config/locales/en.yml`. The new-account form keeps its editable checkbox since the flag is settable at creation. The attribute remains `attr_readonly` and is still not permitted by `edit_account_params`, so this is a presentation fix on top of existing server-side protection.

## Placement rationale

This belongs in samvera/hyku core rather than in a downstream overlay. The `public_demo_tenant` flag, `Hyrax::Ability::TenantControlAbility`, the proprietor presenter guard, and the demo-tenant superadmin seeding in `CreateAccount` all live in core already; this PR completes invariants those features imply but did not enforce. Any institution using the existing flag to run a public sandbox is exposed to the same three gaps, so the fix is generally useful rather than deployment-specific. The changes are also inert for everyone else: with `public_demo_tenant` false or unset, the superadmin setter, the invite flow, and the users page behave exactly as before, and the only behavior change visible outside demo tenants is the edit form rendering an immutable attribute as text.

## Specs

The first commit contains only specs, which fail without the implementation; the second commit makes them pass.

* `spec/models/site_spec.rb`: new `.superadmin_emails=` coverage. On a demo tenant: raises on empty assignment, blank-only assignment, and no-matching-user replacement, each leaving the existing superadmin intact; allows a single-assignment swap to another existing user; allows removing one of several superadmins. On a standard tenant and on a site with no account: clearing all superadmins still works.
* `spec/controllers/proprietor/accounts_controller_spec.rb`: PUT #update with a blank superadmin list on a demo account re-renders edit with an error on `superadmin_emails` and leaves the superadmin in place; a replacement update succeeds end to end.
* `spec/controllers/hyku/invitations_controller_spec.rb`: on a public demo tenant, a plain admin's invite is denied (redirect, no user created) while a tenant superadmin's invite succeeds; on a standard tenant a plain admin can still invite.
* `spec/abilities/tenant_control_ability_spec.rb`: adds user-manager cases in both tenant modes, documenting that user managers never hold `:tenant_controls`, which is why the controller gate applies only on demo tenants.
* `spec/views/hyrax/admin/users/index_demo_tenant.html.erb_spec.rb` (new file): the invite card hides for a plain admin on a demo tenant and renders for a tenant superadmin; the user listing renders either way. This lives in its own file because the sibling spec renders in a top-level before block, before the demo flag could be stubbed.
* `spec/views/proprietor/accounts/edit.html.erb_spec.rb`: the edit form renders no `account[public_demo_tenant]` input and shows the read-only Yes/No text for demo and standard accounts.
* `spec/presenters/proprietor/account_presenter_spec.rb`: `can_remove_admin?` returns false for the last superadmin on a demo tenant and true for the last superadmin on a standard tenant.

## Notes for reviewers

* The raise-and-rescue shape for the superadmin floor is unusual but deliberate. The email-list setters apply role changes during `assign_attributes`, so by the time normal validations run the roles are already gone. The guard therefore has to halt the assignment itself. The controller rescue converts it into the same render-edit-with-errors flow as a validation failure.
* If a proprietor removes the admin role from a user who is also the last superadmin on a demo tenant via a hand-crafted request (the button is now disabled in the UI), the admin-role removal can apply before the superadmin guard raises, since `admin_emails` is assigned first. The security property, at least one superadmin retained, still holds; the partial admin change is a pre-existing consequence of assignment-time side effects and is noted as a follow-up.
* The guard treats "requested emails that match no existing user" as removal, because `add_superadmins_by_email` only grants roles to existing users. Without that, a typo in a swap could zero out the superadmins while appearing to succeed.
* The invite denial redirects (via the app-wide CanCan rescue) rather than rendering a 403 page, consistent with other `authorize!` failures in non-CRUD actions.

## Follow-ups

* The server-side invite path in `Site#add_admins_by_email` (`User.invite!` for unknown admin emails posted through the proprietor form) does not go through `Hyku::InvitationsController` and is not affected by this gate; it is proprietor-only today but could get the same treatment for symmetry.
* On standard tenants the invite endpoint still only requires an authenticated user, a pre-existing looseness this PR intentionally does not change to keep the diff to demo containment.
* Making the email-list setters transactional (assign at save time instead of assignment time) would remove the raise-and-rescue shape and the partial-application edge case noted above.
* The old inline hint text for the edit-form checkbox was duplicated into several non-English simple_form locale files; those entries are now unused and could be cleaned up.

### DIFF
--- a/app/controllers/proprietor/accounts_controller.rb
+++ b/app/controllers/proprietor/accounts_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Proprietor
+  # rubocop:disable Metrics/ClassLength
   class AccountsController < ProprietorController
     before_action :ensure_admin!
 
@@ -147,4 +148,5 @@ module Proprietor
       end
     end
   end
+  # rubocop:enable Metrics/ClassLength
 end
--- a/app/controllers/proprietor/accounts_controller.rb
+++ b/app/controllers/proprietor/accounts_controller.rb
@@ -75,6 +75,8 @@ module Proprietor
           format.json { render json: @account.errors, status: :unprocessable_entity }
         end
       end
+    rescue Site::LastSuperadminRemovalError
+      respond_with_last_superadmin_error
     end
 
     # DELETE /accounts/1
@@ -131,6 +133,17 @@ module Proprietor
     def deleted_or_new(hash)
       hash.detect do |_k, v|
         ActiveModel::Type::Boolean.new.cast(v["_destroy"]) == true || v["id"].blank?
+      end
+    end
+
+    # Public demo tenants must keep at least one site-scoped superadmin; the
+    # role removal happens during attribute assignment, so the model raises
+    # rather than failing validation. Surface it as a form error.
+    def respond_with_last_superadmin_error
+      @account.errors.add(:superadmin_emails, :cannot_remove_last_superadmin)
+      respond_to do |format|
+        format.html { render :edit }
+        format.json { render json: @account.errors, status: :unprocessable_entity }
       end
     end
   end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Site < ApplicationRecord
+  # Raised when a superadmin_emails assignment would leave a public demo
+  # tenant with no site-scoped superadmins.
+  class LastSuperadminRemovalError < StandardError; end
+
   resourcify
 
   validates :application_name, presence: true, allow_nil: true
@@ -56,11 +60,14 @@ class Site < ApplicationRecord
 
   # Update superadmin emails associated with this site
   # @param [Array<String>] Array of user emails
+  # @raise [LastSuperadminRemovalError] when the assignment would leave a
+  #   public demo tenant with no site-scoped superadmins
   def superadmin_emails=(emails)
     emails = Array(emails).reject(&:blank?)
     existing_superadmin_emails = superadmin_emails
     new_superadmin_emails = emails - existing_superadmin_emails
     removed_superadmin_emails = existing_superadmin_emails - emails
+    guard_last_superadmin_removal!(emails) if removed_superadmin_emails.present?
     add_superadmins_by_email(new_superadmin_emails) if new_superadmin_emails.present?
     remove_superadmins_by_email(removed_superadmin_emails) if removed_superadmin_emails.present?
   end
@@ -112,5 +119,20 @@ class Site < ApplicationRecord
     User.where(email: emails).find_each do |u|
       u.remove_role :superadmin, self
     end
+  end
+
+  # Public demo tenants must always keep at least one site-scoped superadmin.
+  # Role changes apply immediately when the attribute is assigned (not at
+  # save), so this guard must halt the assignment; a validation error added
+  # after the fact would not roll the removal back. The requested emails only
+  # produce superadmins for users that already exist, so the assignment is
+  # refused when none of them match an existing user.
+  # @param [Array<String>] requested_emails the full email list being assigned
+  # @raise [LastSuperadminRemovalError]
+  def guard_last_superadmin_removal!(requested_emails)
+    return unless account&.public_demo_tenant?
+    return if User.where(email: requested_emails).exists?
+
+    raise LastSuperadminRemovalError, I18n.t('activerecord.errors.messages.cannot_remove_last_superadmin')
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -10,6 +10,9 @@ en:
     attributes:
       site:
         institution_name_full: Full institution name
+    errors:
+      messages:
+        cannot_remove_last_superadmin: cannot remove the last tenant superadmin from a public demo tenant
   application:
     tagline: The next-generation repository solution
   blacklight:
@@ -366,6 +369,10 @@ en:
         data_cite_endpoint: DataCite Endpoint
         fcrepo_endpoint: Fedora Endpoint
         header: Editing Account
+        public_demo_tenant:
+          disabled: 'No'
+          enabled: 'Yes'
+          immutable: This setting is defined when the account is created and cannot be changed.
         solr_endpoint: Solr Endpoint
       index:
         actions: Actions
--- a/app/controllers/hyku/invitations_controller.rb
+++ b/app/controllers/hyku/invitations_controller.rb
@@ -2,6 +2,11 @@
 
 module Hyku
   class InvitationsController < Devise::InvitationsController
+    # On public demo tenants, inviting users is restricted to holders of
+    # manage :tenant_controls (i.e. tenant superadmins).
+    # @see Hyrax::Ability::TenantControlAbility
+    before_action :ensure_tenant_controls_on_demo_tenant!, only: :create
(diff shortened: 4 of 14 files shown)


app/controllers/proprietor/accounts_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Proprietor
+  # rubocop:disable Metrics/ClassLength
```
orangewolf, line 4: fix the classes length please, don't just turn it off

app/models/site.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Site < ApplicationRecord
+  # Raised when a superadmin_emails assignment would leave a public demo
+  # tenant with no site-scoped superadmins.
+  class LastSuperadminRemovalError < StandardError; end
```
orangewolf, line 6: this should not be an exception

app/models/site.rb
```diff
@@ -113,4 +120,19 @@ def remove_superadmins_by_email(emails)
...
+  # after the fact would not roll the removal back. The requested emails only
+  # produce superadmins for users that already exist, so the assignment is
+  # refused when none of them match an existing user.
+  # @param [Array<String>] requested_emails the full email list being assigned
+  # @raise [LastSuperadminRemovalError]
+  def guard_last_superadmin_removal!(requested_emails)
+    return unless account&.public_demo_tenant?
+    return if User.where(email: requested_emails).exists?
+
+    raise LastSuperadminRemovalError, I18n.t('activerecord.errors.messages.cannot_remove_last_superadmin')
```
orangewolf, line 136: render a proper error, not raise an exception please

config/locales/en.yml
```diff
@@ -10,6 +10,9 @@ en:
     attributes:
       site:
         institution_name_full: Full institution name
+    errors:
```
orangewolf, line 13: the rest of the languages need to be filled in

orangewolf (changes requested):
tell claude I said this was really 3 separate PRs ins a trench coat and to split them up by feature 🤪

ShanaLMoore:
Closing in favour of three focused PRs, per your review. Picking this up while Nick is out.

| PR | Feature | Your comments |
|---|---|---|
| #3185 | Restrict user invitations on public demo tenants | none, carried over unchanged |
| #3186 | Show `public_demo_tenant` as read-only on the proprietor form | the translations one |
| #3187 | Keep at least one superadmin on public demo tenants | the other three |

They are independent and all base on main, so there is no stacking and they can be reviewed or merged in any order.

Two things worth pulling out of #3187, since they are the substance of your review:

**The exception is gone, and removing it fixed a bug.** The original code defended the raise on the grounds that role changes apply on assignment rather than on save, so a validation could not roll the removal back. True, but it assumed the removal had to happen first. It doesn't: the removal is skipped, a flag is set, and an ordinary validation fails the save. The old code removed the role and *then* raised, so a refused assignment still stripped the superadmin. It no longer does, and there is a spec for exactly that.

**The ClassLength disable is gone without touching the threshold.** Deleting the `rescue` and its bespoke error responder took the class back under the limit on its own, since those were the lines that pushed it over.

The translations are in all seven locales, following each file's existing terms, but they are mine rather than a native speaker's, so please correct anything that reads badly.

PR #3128, opened 2026-07-06
Credential and signup hygiene: ENV-sourced passwords, configurable basic auth, and an in-app signup spam challenge

SixtyGrit:
## Summary

This PR removes three hardcoded credentials from the codebase and adds an optional, privacy-respecting spam challenge to user self-registration. The seed rake tasks no longer hardcode `testing123`, the admin "activate user" action no longer falls back to the literal `password`, the HTTP Basic gate for non-public tenants is now configurable through environment variables and fails closed when unconfigured, and tenants can enable a honeypot plus minimum-submit-time challenge on the registration form. The challenge is always enforced on accounts flagged as `public_demo_tenant`.

This work arose from an effort to make public Hyku demo environments safe to operate. A demo tenant that is open to the world inherits every default credential and every open signup path in the codebase, so those defaults needed to become configurable and safe.

## Motivation

Four related issues, all verified against current `main`:

1. `RolesService.seed_superadmin!` and `RolesService.seed_qa_users!` created users with the hardcoded password `testing123`. Both methods refuse to run in production and staging, but any long-lived development, QA, or demo-like environment seeded with these tasks carries well-known credentials for accounts that hold the superadmin role on every tenant.
2. `Admin::UsersController#activate` set a user's password to `ENV.fetch('HYKU_USER_DEFAULT_PASSWORD', 'password')`. Any deployment that never set that variable was assigning the literal `password` to every activated user.
3. `ApplicationController#authenticate_if_needed` guarded hidden (non-public) tenants and staging environments with HTTP Basic credentials hardcoded to `samvera` / `hyku`. These values are published in the codebase, so the gate provided no real protection and could not be changed without patching the class attributes in application code.
4. `allow_signup` defaults to true per tenant, and the registration form had no bot protection of any kind. For a public demo tenant this means unattended bot registrations.

### Seed passwords (`app/services/roles_service.rb`)

`seed_superadmin!` and `seed_qa_users!` now read the password from `HYKU_SEED_PASSWORD`. When that variable is not set, they generate a random `SecureRandom.base58(24)` password, log it, and print it once so the operator can capture it. Within a single `seed_qa_users!` run, all newly created QA users share one generated password and it is announced once. Existing users are never touched; the password only applies to users the tasks create. The existing production and staging guards are unchanged.

Development bootstrap note: if your local workflow depended on `testing123`, set `HYKU_SEED_PASSWORD=testing123` in your local environment, or read the generated password from the task output.

### Activation password (`app/controllers/admin/users_controller.rb`)

`#activate` now uses `HYKU_USER_DEFAULT_PASSWORD` when it is present and otherwise assigns a random `Devise.friendly_token`. Deployments that set the variable see no behavior change. Deployments that never set it previously handed out the literal `password`; they now assign an unknown placeholder, and the activated user can set a real password through the standard password reset flow. The checked-in `.env` still sets this variable for local docker development, so local behavior is unchanged.

### HTTP Basic gate (`app/controllers/application_controller.rb`)

The `http_basic_auth_username` and `http_basic_auth_password` class attributes now default to nil and act as overrides. Effective credentials resolve in this order: class attribute, then `HYKU_BASIC_AUTH_USER` / `HYKU_BASIC_AUTH_PASSWORD` environment variables. When neither source provides both values, the gate fails closed and refuses access, with one exception: the development environment keeps the historical `samvera` / `hyku` fallback for local convenience. Comparison now uses `ActiveSupport::SecurityUtils.secure_compare`.

**Operator action required.** If you run staging environments or hidden tenants and relied on the documented `samvera` / `hyku` defaults, set `HYKU_BASIC_AUTH_USER` and `HYKU_BASIC_AUTH_PASSWORD` before deploying this change, or set the class attributes in an initializer. Without one of those, guarded pages will return 401 for all credentials. This is intentional; the published defaults were not protecting anything, and failing closed is the safer default for a gate whose purpose is to hide tenants.

### Signup spam challenge (registration form)

A new per-tenant boolean setting, `signup_spam_protection` (default false), adds two protections to `Hyku::RegistrationsController#create`:

* A honeypot text field, rendered visually hidden with `aria-hidden` and a negative tabindex, that must remain blank.
* A signed timestamp, generated with `Rails.application.message_verifier`, that enforces a minimum of 4 seconds between rendering the form and submitting it. Tampered or missing timestamps are rejected.

Rejected submissions redirect back to the signup form with a generic, translatable message. The challenge runs entirely in-app: no third-party service, no cookies, no fingerprinting, and no new gem. Both registration view variants render the same shared partial, and the fields do not render at all when the challenge is off.

The setting defaults to off so existing tenants see no behavior change, and tenant administrators can enable it from the account settings form like any other boolean setting (an ENV fallback, `HYKU_SIGNUP_SPAM_PROTECTION`, follows the standard settings precedence). Accounts flagged `public_demo_tenant` enforce the challenge unconditionally, since open self-registration on a public demo is the case this exists for.

The existing `allow_signup` setting, its default, and its enforcement are untouched.

### Documentation

`docs/configuration.md` gains rows for `HYKU_SEED_PASSWORD`, `HYKU_BASIC_AUTH_USER`, `HYKU_BASIC_AUTH_PASSWORD`, `HYKU_SIGNUP_SPAM_PROTECTION`, and the previously undocumented `HYKU_USER_DEFAULT_PASSWORD`.

## Placement rationale

This belongs in samvera/hyku core rather than in a downstream knapsack. Every one of the four defaults ships in core: the seed tasks, the activation fallback, the basic auth literals, and the unprotected registration action are all core code, so every Hyku adopter inherits them. A knapsack override could patch these for one installation, but the weak defaults would remain the out-of-the-box experience for everyone else, and credential defaults are exactly the kind of thing that should be safe upstream rather than fixed per deployment. The signup challenge follows the existing core extension points (`AccountSettings.setting`, the `Hyku::RegistrationsController` Devise override, and the shared Devise view partials), and it generalizes: any institution running a tenant with open signup benefits, whether or not it is a demo. The `public_demo_tenant` flag it integrates with is itself a core Account attribute.

## Specs

Spec-first commit history: the first commit contains only the specs, which fail against the previous implementation; the second commit makes them pass.

* `spec/services/roles_service_spec.rb`: for both seed methods, asserts that a configured `HYKU_SEED_PASSWORD` becomes the created user's password, that the historical `testing123` is no longer valid when the variable is unset, and that the generated password announcement is printed.
* `spec/controllers/admin/users_controller_spec.rb`: asserts activation uses the configured `HYKU_USER_DEFAULT_PASSWORD`, and that with the variable unset the literal `password` is not assigned while the invitation is still accepted.
* `spec/controllers/application_controller_spec.rb`: exercises `authenticate_if_needed` through an anonymous controller with the test-environment bypass stubbed out. Covers ENV-configured credentials granting access, the historical defaults being denied when ENV credentials are set, fail-closed behavior when nothing is configured, the development-only fallback, and attribute overrides taking precedence over ENV.
* `spec/requests/user_signup_challenge_spec.rb`: request specs against `POST /users` covering rejection when the honeypot is filled, rejection of too-fast submissions, rejection of tampered timestamps, acceptance of legitimate submissions, unchanged behavior when the setting is off, unconditional enforcement on public demo tenants, and rendering of the challenge fields only when the challenge is active.
* `spec/models/concerns/account_settings_spec.rb`: updated the public settings list to include `signup_spam_protection`.

## Notes for reviewers

* The basic auth gate remains skipped in the test environment and for JSON, API, and PDF paths, exactly as before. Those exemptions are pre-existing behavior and are out of scope here.
* The generated seed password is intentionally printed to stdout and the log. These tasks refuse to run in production and staging, and the operator needs the password to be able to log in; announcing it once at creation is the standard tradeoff for random seed credentials.
* The minimum-submit-time constant is 4 seconds. Password manager users click submit after page load and interaction, which comfortably clears 4 seconds; a legitimate user who somehow submits faster gets a retry message and succeeds on the second attempt because the form re-renders with a fresh timestamp.
* The timestamp is signed with the application secret via `message_verifier`, so it cannot be forged or predated by a client. A harvested timestamp older than the threshold would pass the timing check, so the honeypot and the timing check are deliberately layered.
* The checked-in `.env` and the deploy templates under `ops/` still contain weak values for local development convenience. Changing deployment templates is left out of this PR to keep it reviewable; see follow-ups.

## Follow-ups

* Rotate `INITIAL_ADMIN_PASSWORD` handling in `db/seeds.rb`, which runs without an environment guard and reads a checked-in default from `.env` and the deploy templates. That path deserves its own PR because it affects production deployments directly.
* Consider rack-attack throttles for the authentication endpoints (`/users/sign_in`, `/users/sign_up`, `/users/password`); the current initializer only throttles catalog requests.
* Translations for the two new locale strings in the non-English locale files.
* Consider removing `HYKU_USER_DEFAULT_PASSWORD=password` from the checked-in `.env` and the `ops/` templates once operators have had a release cycle to configure their own values.


orangewolf (changes requested):
this one should be broken up ever comma in the description is a separate feature ;-)

ShanaLMoore:
Closing as superseded. This PR's changes were split into three focused PRs per review feedback:

- #3189 - seed-password-env
- #3191 - basic-auth-env-config
- #3192 - signup-spam-challenge

Confirmed the file-level overlap directly: 14 of this PR's 16 changed files are identical to the union of those three (`app/controllers/application_controller.rb`, `app/controllers/hyku/registrations_controller.rb`, `app/models/concerns/account_settings.rb`, `app/services/roles_service.rb`, the devise signup views, `docs/configuration.md`, and their specs). Keeping this open alongside the split PRs risks double-review and a merge conflict if both land.

PR #3137, opened 2026-07-09
Fix/generator tag hyku identification

trmccormick:
Fixes #7509 ; refs samvera/hyrax#7509

Override generator meta tag to identify as Samvera Hyku

This PR implements a minimal method override to correctly identify Hyku instances as "Samvera Hyku" instead of "Samvera Hyrax". This leverages the `hyrax_generator_meta_tag` helper support added in Hyrax PR #7509.

## Background
When tools or services check the generator meta tag in the page header, Hyku instances were incorrectly reporting themselves as "Samvera Hyrax" (the upstream framework) instead of "Samvera Hyku" (the Hyku application). This made it difficult to distinguish between Hyku tenants and standalone Hyrax instances.

## Implementation
The fix uses a single method override:
- **Helper method** (`app/helpers/hyku_helper.rb`): Overrides `hyrax_generator_meta_tag` to return `"Samvera Hyku #{::Hyku::VERSION}"`

Hyrax's existing `_head_tag_content.html.erb` view already calls this helper method, so no view override is needed. This approach is minimal, leverages Hyrax's tested infrastructure, and keeps maintenance burden low.

## Verification
The fix has been verified:
- ✅ Helper method `hyrax_generator_meta_tag` defined in Hyku
- ✅ Returns correct format: `"Samvera Hyku #{::Hyku::VERSION}"`
- ✅ Hyrax view calls the overridden method automatically

After deployment, the page source will show: `<meta name="generator" content="Samvera Hyku 7.1.0">` instead of the previous `Samvera Hyrax` identification.

## Changes in this pull request
* Override `hyrax_generator_meta_tag` helper method in `app/helpers/hyku_helper.rb` to return Hyku-specific generator tag with version

@samvera/hyku-code-reviewers


app/helpers/hyku_helper.rb
```diff
@@ -40,4 +40,9 @@ def parent_path(parent_doc)
     path = "#{model.model_name.singular_route_key}_path"
     main_app.send(path, parent_doc.id)
   end
+
+  # Override Hyrax's hardcoded "Samvera Hyrax" generator tag with correct identification
+  def hyku_generator_meta_tag
```
orangewolf: if you override the `hyrax_generator_meta_tag` method directly can you skip the `_head_tag_content` override? Just trying to save the number of custom lines where we can =-)

orangewolf (changes requested):

app/helpers/hyku_helper.rb
trmccormick: thanks for the tip let me adjust this then.

trmccormick: adjusted

orangewolf (approved):

PR #3139, opened 2026-07-10
Deposit wizard

laritakr:
## Summary
Adds an opt-in guided, multi-step deposit wizard that walks users through choosing a work type, uploading files, entering work and per-file metadata, setting visibility, and reviewing before deposit. This is off by default behind a `deposit_wizard` feature flag; stock deposit is unchanged.

Ref https://github.com/research-technologies/enact_knapsack/issues/81

## Details

This supports all existing deposit functionality, and adds some much-needed new features: file level metadata and permissions on deposit (file visibility can vary from work visibility with initial deposit). It also allows adding to a parent work as part of the deposit flow when this feature is enabled.

The features which are included are configurable via config and/or feature flippers.

### Architecture
- New dashboard flow at `/dashboard/deposit_wizard`, reachable from a flag-gated entry point on the "My Works" page.
- Basic steps: type selection → file upload → work details → per-file metadata (only when files are attached) → review → deposit. Feature flippers allow adding options to the review step.
- Reuses Hyrax's public services (`ResourceForm`, `WorkUploadsHandler`, the `work_resource.create` transaction, admin-set presenter) — no Hyrax overrides; the wizard owns navigation and layout, not fields.
- Per-file metadata (title, creator, own visibility incl. embargo/lease) is collected inline before commit and applied to each FileSet.
- Metadata entered on any step is restored when navigating back; Next is disabled until a step's required fields are valid.
- Config-driven (`Hyku::DepositWizard::Config`) so a downstream app can supply container/child types and styling; default config gives a plain install a usable flat-work-type wizard.

## Screenshots & Video

<details>
<summary>Screen Recording of Deposit</summary>

https://github.com/user-attachments/assets/28603316-e340-4fce-b4bc-1266dbc97781

</details>

## Notes

Potential concepts for later enhancements include:
- file drag & drop sequencing
- file "pool" - upload all files at once, then select one or more to create a work for, cycling through creating works until all files are used.
- specify a file to be used as the work's thumbnail/representative image
- save the in-browser progress and resume prior to a work being saved (convert DepositWizard::State to ActiveRecord backed)
- show authority labels rather than authority URIs on review page

_For transparency, I do not know JS and CSS well at all, so I rely on Claude Code for those portions. I can read them and they seem to make sense to me, but I prefer someone with more background look at them with a bit more cautious eye. The rails portions are much more hands-on than those pieces._


app/views/hyrax/deposit_wizard/_visibility.html.erb
```diff
@@ -0,0 +1,93 @@
+<%# Renders the same fields as Hyrax's _form_visibility_component (visibility
```
orangewolf, line 1: there is too much logic in this view. maybe a helper or 3? and yes, I know that means they need tests... its part of the point ;-)

app/views/hyrax/deposit_wizard/file_meta.html.erb
```diff
@@ -0,0 +1,72 @@
+<%= render 'page_header' %>
```
orangewolf, line 1: this view is right on the edge of needing to be broken up in to partials. might be worth doing now, but I'll leave that up to you.

app/views/hyrax/deposit_wizard/review.html.erb
```diff
@@ -0,0 +1,88 @@
+<%= render 'page_header' %>
```
orangewolf, line 1: same partial break up comment as for file_metadata

config/locales/hyrax.en.yml
```diff
@@ -801,6 +801,25 @@ en:
         viz_after: Visibility will Change to
```
orangewolf, line 801: do you have a ticket to update the other languages? I'm a fan of doing it at the end, but we don't want to forget

app/controllers/concerns/hyrax/deposit_wizard/context.rb
```diff
@@ -0,0 +1,182 @@
+# frozen_string_literal: true
```
orangewolf: this many concerns and helper methods make me think we are missing an object. I would recommend moving all of this logic (and the next 2 files too in to a PORO (plain old ruby object) maybe with a concern or 2 maybe not in to the models directory. easier to test, better object oriented shape. hitting the object instead of lots of helper calls means less pollution in the global space in the views (there can only be one `wizard_config` across all helpers for example. `@deposit_wizard.config` doesn't have that problem.

orangewolf: I'm ok if you want to commit it this way and then refactor, but I think you are really going to wish you had refactored sooner rather than later.

orangewolf (changes requested):
pretty solid foundation, but the controller concerns and some of the individual view files are a little over grown and could use some refactoring.

config/locales/hyrax.en.yml
laritakr: it's on my todo list. :) Not a separate ticket, but I've got it planned.

orangewolf (approved):
this looks significantly better. Thank you for all the edits @laritakr

PR #3140, opened 2026-07-12
Deposit wizard extras phase

laritakr:
## Summary

Builds on prior PR https://github.com/samvera/hyku/pull/3139 to complete the guided deposit UI base functionality. The UI should now fully encompass the existing deposit UI features.

Adds optional, per-tenant "connect" capabilities to the guided deposit wizard's review step — nesting under a parent work, adding to collections, sharing with users/groups, and vanity URLs — each gated by its own feature and off by default.

## Details
- Four review-step capabilities, each an independent Flipflop feature: parent-connect, collection-connect, sharing, and redirects (redirects also respects the tenant's existing redirects gate + the work's schema).
- Each appears as a collapsible card (title + subtitle) that opens automatically when it already holds data.
- The vanity-URL card renders Hyrax's stock aliases form, so it inherits the full behavior (none-option, per-row remove, disable-until-path) rather than a partial reimplementation.
- Extras autosave to the wizard's session as the depositor edits them, so they survive a page refresh — the review step re-renders the chosen parent (by title), collections, share grants, and vanity URLs.
- A launch-with-context seam seeds `parent_id` / `add_works_to_collection` from the URL, so other entry points can later hand off into the wizard with a target pre-filled.
- All capabilities persist through the one existing deposit transaction (no new persistence path); the remove-control color was darkened to meet WCAG AA contrast.

## Screenshot

<details>
<summary>Deposit Review Page with additional features</summary>

screencapture-cat-hyku-localhost-direct-dashboard-deposit-wizard-review-2026-07-11-23_50_16

</summary>

### DIFF
--- a/app/controllers/hyrax/deposit_wizard_controller.rb
+++ b/app/controllers/hyrax/deposit_wizard_controller.rb
@@ -29,6 +29,7 @@ module Hyrax
 
     def start
       reset_state
+      seed_launch_context
       render :start
     end
 
@@ -69,6 +70,8 @@ module Hyrax
         return render(:review)
       end
 
+      capture_review_extras
+      build_work_form
       work = create_work
       return render(:review) unless work
 
@@ -97,6 +100,18 @@ module Hyrax
 
     def reset_state
       session[:deposit_wizard] = {}
+    end
+
+    # Seed wizard state from the same context params other entry points pass
+    # allowing a potential connection point with other deposit flows.
+    def seed_launch_context
+      wizard_state.parent_id = params[:parent_id] if wizard_config.enable_parent_connect && params[:parent_id].present?
+
+      return unless wizard_config.enable_collection_connect && params[:add_works_to_collection].present?
+
+      wizard_state.attributes = wizard_state.attributes.merge(
+        'member_of_collections_attributes' => { '0' => { 'id' => params[:add_works_to_collection] } }
+      )
     end
   end
 end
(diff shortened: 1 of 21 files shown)


app/controllers/hyrax/deposit_wizard_controller.rb
```diff
@@ -89,6 +111,14 @@ def ensure_enabled
       redirect_to hyrax.my_works_path, alert: t('hyku.deposit_wizard.disabled')
     end
 
+    # Hyrax's collection/search helpers (e.g. available_collections) read the
+    # @current_ability instance variable, which stock works controllers set via
+    # WorksControllerBehavior. This lean controller must set it itself, or those
+    # helpers bail with an empty list.
+    def assign_current_ability
+      @current_ability = current_ability
```
orangewolf, line 119: ummm, I'm pretty sure you can just call current ability whenever you want and its always the right one. I'm concerned what caching it here does

app/services/hyku/deposit_wizard/config.rb
```diff
@@ -45,6 +45,27 @@ class Config
...
+      # not here — they use their own Flipflop gate (see #redirects_available?).
+      attr_writer :enable_parent_connect, :enable_collection_connect, :enable_sharing
+
+      # Work types eligible as parents; +nil+ falls back to the tenant's available
+      # work types.
+      attr_accessor :parent_types
+
+      # Each reader returns the in-memory override when one was set, otherwise the
+      # tenant's Flipflop feature value.
+      %i[parent_connect collection_connect sharing].each do |capability|
```
orangewolf, line 60: "I want to be a regular ruby object soooo badly" - this code

orangewolf (approved):
this comment just echos the previous ones. no reason it can't be merged together now

PR #3143, opened 2026-07-14
🧹 Drop webm as a video derivative

kirkkwang:
Nothing is using webm and it takes a lot of resources to generate.  This commit will remove webm as a derivative format for video files.


laritakr (approved):

kirkkwang:
Adding a major-ver (breaking change) label since the question was raised.  If an application was using webm they are now longer being created by default.

PR #3145, opened 2026-07-16
Keep scroll position when toggling features

laritakr:
## Summary

The admin Features page now stays where you were after flipping a feature, instead of jumping back to the top on every toggle.


ShanaLMoore (approved):
nice!

PR #3146, opened 2026-07-16
Deposit wizard flow map

laritakr:
## Summary

Extracts the deposit wizard's step flow into configurable data, so a downstream app can add, remove, or reorder steps without editing wizard logic.
https://github.com/research-technologies/enact_knapsack/issues/81

## Details
- Step order, skips, prerequisites, back/forward navigation, and the progress rail move onto one `Flow` object; the presenter delegates to it and views stop hardcoding their Back targets.
- Steps declare named prerequisites instead of a hardcoded work-type flag — file upload no longer presumes a chosen work type, the seam a later guided flow needs to upload before predicting the type.
- Behavior-identical for every existing path; the request-spec suite is the parity guard.
- Config cull: unwired options dropped (`single_admin_set`, `enable_batch`, `file_meta`, `file_pool`); `item_types` wired to restrict the offered work types.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -37,6 +37,7 @@
 //= require hyku/admin/appearance/default_images
 //= require hyku/admin/appearance/fonts
 //= require hyku/admin/appearance/themes
+//= require hyku/admin/features_scroll
 //= require hyku/groups/per_page
 //= require hyku/groups/add_member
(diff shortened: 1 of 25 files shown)


orangewolf (approved):

orangewolf:
Code was discussed in detail via meeting

PR #3149, opened 2026-07-18
Add deposit wizard extension seams for downstream

laritakr:
## Summary
Adds the extension points a downstream app needs to insert its own steps into the guided deposit wizard more cleanly. This should be no-op other than fixes for a few minor bugs.

- Refs notch8/enact_knapsack#81

## Details
- Adds `State#extra`: a namespaced session bag so a downstream app persists its own step state without subclassing `State`.
- Updates presenter `advance_from` to fall back to advancing to the next visible step for any step it doesn't recognize, to minimize overrides needed when inserting steps.
- Removes a now-redundant hardcoded step case that the fallback covers.
- Removes the wizard's own per-file metadata partial in favor of overriding Hyrax's `_form_metadata`. This override can be removed once the change is ported to Hyrax.
- Fixes a bug: `build_work_form` restores saved values with `deserialize` rather than `validate`, so required-field errors no longer appear before the depositor submits.
- Ensures that the step to add a parent shows the button as disabled until a parent it added.
- Allows for icons to show on the various cards. Uses the i18n file, a method used by Hyrax for assigning icons to work types.


app/views/hyrax/deposit_wizard/_file_panel.html.erb
```diff
@@ -17,7 +17,9 @@
       config; a bare SimpleForm::FormBuilder renders labels/inputs differently from
       the work details form. %>
   <%= simple_fields_for "file_metadata[#{id}]", form do |f| %>
-    <%= render 'file_metadata', f: f, panel_id: id %>
+    <%# The shared work-form metadata partial (title, compounds, additional fields),
+        with a per-panel collapse id so multiple file panels don't collide. %>
+    <%= render 'hyrax/base/form_metadata', f: f, extended_terms_id: "file-extended-terms-#{id}" %>
```
ShanaLMoore, line 22: On flexible-metadata tenants this now renders the `schema_version` and `contexts` terms in each per-file panel. The deleted `_file_metadata.html.erb` stripped them (`primary_terms - %i[schema_version contexts]`); the shared partial renders `primary_terms` unfiltered, and a flexible `Hyrax::FileSetForm` includes both (`resource_form.rb`: `terms = [:schema_version, :contexts] + terms if model.flexible?`). `contexts` is a hidden field, but `schema_version` renders `shared/_schema_version`, a visible floated "Version After Save" card, on every file panel. Non-flexible tenants are unaffected. Suggest filtering those two terms out for the FileSet call site to restore the prior behavior.

ShanaLMoore (approved):
Approving; this is non-blocking. Verified the partial swap, the advance_from fallback, the deserialize fix, and the parent-next JS toggle against the pinned Hyrax revision and the wizard controller/flow. One follow-up worth doing (inline).

app/views/hyrax/deposit_wizard/_file_panel.html.erb
laritakr: There is css which hides the schema version for the work details, and this also covers the file set details. Because this is only a deposit (not edit) wizard, showing schema version was unnecessary (and rather confusing) since nothing is changing.

The issue will be if something different is needed when we get to editing, which is part of the reason for keeping this override in Hyku until we determine exactly what IS needed.

PR #3150, opened 2026-07-21
Deposit wizard integration modes

laritakr:
## Summary
Configures the guided deposit to be used across every deposit entry point in the site, based on per-tenant feature flags.
- Refs research-technologies/enact_knapsack#112

## Details
- Two per-tenant features select the deposit experience: `enable_guided_deposit` (off by default) and `enable_standard_deposit` (on by default).
- When guided deposit is enabled, every deposit entry point — the works page, homepage "share your work", collection "Add new work", and a work's "Attach a child work" — routes into the guided wizard instead of the standard single-page form.
- When only standard deposit is enabled, those entry points use the standard form (current behavior).
- Each enabled feature adds its own button to the dashboard works page; when both are enabled, the guided start screen also offers a "switch to the standard deposit form" link.
- Parent-connect, collection-connect, and depositor-sharing become configuration settings (default on) rather than feature flags.
- A directed handoff — "deposit into this collection" or "attach a child to this work" — carries that association through to deposit regardless of the picker settings.
- Moves the home page "Share your work" flipper into the deposit features section, since it is also a deposit related function.

## Testing
- **Standard only (default):** confirm every deposit entry point uses the standard form and the works page shows the standard "Add new work" button.
- **Guided only:** enable `enable_guided_deposit`, disable `enable_standard_deposit`. Confirm the works page, homepage "share your work", collection "Add new work", and a work's "Attach a child work" all route into the guided wizard.
- **Both enabled:** confirm both buttons appear on the works page, entry points route to guided, and the guided start screen shows a "switch to the standard deposit form" link.
- **Collection handoff:** from a collection's "Add new work" (guided enabled), complete the wizard and confirm the work lands in that collection.
- **Parent handoff:** from a work's "Attach a child work" (guided enabled), complete the wizard and confirm the work nests under the parent.

## Screenshot

Screenshot 2026-07-22 at 3 24 05 PM

## Follow-up work

This "dual-flipper" structure allows both deposit experiences to be selected and active, but it also allows neither to be active (in a situation where all deposits are intended to come via Bulkrax). This PR removes the options only from the dashboard-works page. It does not remove these other deposit entry points for the situation where both flippers are off:
- homepage (this button has its own feature flipper)
- add a child work from a work: work-show view
- deposit a new work within a collection: collection dashboard-show view

Turn off all UI deposit: https://github.com/research-technologies/enact_knapsack/issues/130

Override Cleanup: https://github.com/samvera/hyku/issues/3156

Accessibility Concerns: https://github.com/samvera/hyku/issues/3157

### DIFF
--- a/app/controllers/hyrax/deposit_wizard_controller.rb
+++ b/app/controllers/hyrax/deposit_wizard_controller.rb
@@ -29,6 +29,7 @@ module Hyrax
     def start
       reset_state
       seed_launch_context
+      assign_admin_sets_for_standard_chooser
       render :start
     end
 
@@ -57,7 +58,7 @@ module Hyrax
     end
 
     def parent_options
-      return head(:forbidden) unless wizard_config.capabilities.parent_connect?
+      return head(:forbidden) unless wizard_config.parent_connect?
 
       # FindWorksSearchBuilder excludes a "current" work by params[:id]; the wizard
       # has no current work, so a blank id excludes nothing.
@@ -139,7 +140,7 @@ module Hyrax
     # Mirror Hyrax's batch-upload guard: redirect to the dashboard rather than
     # exposing the wizard routes when the feature is off.
     def ensure_enabled
-      return if Flipflop.deposit_wizard?
+      return if wizard_config.enabled?
 
       redirect_to hyrax.my_works_path, alert: t('hyku.deposit_wizard.disabled')
     end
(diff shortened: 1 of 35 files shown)


ShanaLMoore (commented):
Nice work; the mode matrix is coherent, the docs are thorough, and the test coverage is solid. A few things to resolve before I approve, one of which I'd like a clear answer on.

### 1. Authorization on the `parent_id` handoff (blocking question)
`seed_launch_context` now seeds `parent_id` unconditionally (dropping the `parent_connect?` gate). At commit, `commit_params` forwards it into Hyrax's `Transactions::Steps::AddToParent`, which does no authorization: it loads the parent by id and does `parent.member_ids += [obj.id]`, then saves. So a signed-in depositor can `GET /deposit_wizard?parent_id=<any_work_id>`, finish the wizard, and have their work appended to an arbitrary work's members without edit rights on that parent. Previously, tenants with `deposit_wizard_parent_connect` off were not exposed via the wizard; this PR opens it in every wizard-enabled tenant.

Question: does the classic nested-new `#create` path add an `authorize!`/`can?(:edit, parent)` check that the wizard controller lacks? If yes, we should mirror it here. If the classic path is equally ungated, this is a pre-existing Hyrax gap and I'm fine treating it as out of scope, but let's confirm rather than assume. Same sanity check for `add_works_to_collection` (collection deposit rights are more likely enforced, but worth a look).

### 2. Full-file override of `app/views/hyrax/homepage/index.html.erb` (non-blocking)
I confirmed it's a faithful mirror of Hyrax 5.2.0, so nothing is dropped today. But the non-button parts will drift silently on the next Hyrax bump with no test to catch it. The TODO to upstream a share-button partial is the right call; can we file the follow-up issue now so it isn't lost?

### 3. `docker-compose.yml` `SOLR_HOST=solr` (non-blocking)
This dev-env SolrCloud fix is unrelated to deposit modes. Fine to keep, just noting it's riding along in a feature PR.

laritakr:
> Nice work; the mode matrix is coherent, the docs are thorough, and the test coverage is solid. A few things to resolve before I approve, one of which I'd like a clear answer on.
> 
> ### 1. Authorization on the `parent_id` handoff (blocking question)
> 
> ### 2. Full-file override of `app/views/hyrax/homepage/index.html.erb` (non-blocking)
> 
> ### 3. `docker-compose.yml` `SOLR_HOST=solr` (non-blocking)
> 

Thanks for the detailed review. 

- The solr_host change is going to be merged before this work in a separate PR if it works - I just pulled it in here to stabilize my testing environments.
- Follow-up tickets are a good plan. 
- I assumed that the parent work and collection did the verification appropriately because it was pre-existing behavior that I just connected in my code. But it looks like you possibly caught an issue in Hyrax.

I moved back to draft because I'm re-adjusting the feature flags.

laritakr:
Confirmed: The parent work & collection issues are existing bugs in Hyrax. Nice catch, Shana!

The wizard does nothing more to expose them.... the entry points are gated in the UI, and the search queries are also gated. It’s only the transaction actions which are not gated, a Hyrax bug in Valkyrie transactions. This is a low-risk vulnerability but was was mentioned at the weekly tech call.

ShanaLMoore (approved):
Approving. Restructure is clean and the defaults keep today's behavior, so no changes needed from me.

One thing I didn't see covered yet: with the flag rename, the old `deposit_wizard*` flags will reset on upgrade, and enabling guided now turns the parent/collection/sharing pickers on by default (they're config now, not per-tenant flags). Given the minor-ver label, might be worth a line in the release notes. The Hyrax auth bug and the homepage follow-up are already handled from the earlier thread.

Issue #1422, opened 2017-08-29
[HyBox/pilot] Blacklight::Exceptions::InvalidRequest: RSolr::Error::Http - 400 Bad Request
Error:     {metadata={error-class=org.apache.solr.common.SolrException,root-error-class=org.apache.zookeeper.KeeperException$ConnectionLossException},msg=Could not load collection from ZK: f7b21576-7cc1-4859-bfad-251059b1e13e,code=400}

URI: http://solr.pilot.vpc.hykudirect.org/solr/f7b21576-7cc1-4859-bfad-251059b1e13e/select?wt=json&qt=search&facet.field=resource_type_sim&facet.field=creator_sim&facet.field=tag_si...


cbeer:
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/48327/faults/34513805)

PR #3159, opened 2026-07-28
Route demo/staging mail through SendGrid instead of maildev

maxkadel:
hyku-demo and hyku-staging currently send outbound mail to an internal maildev catch-all, so password-reset emails never leave the cluster. This is blocking self-serve password resets for users recovered after the r2-friends DB loss (dev-ops #1422). 

Also re-enables `demo` in the deploy workflow_dispatch options, so that we can push this fix to demo - re-deploying the same branch to the demo environment for now.

@samvera/hyku-code-reviewers

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -9,7 +9,7 @@ on:
         default: "staging"
         type: choice
         options:
-          # - demo - pausing for accessibility testing
+          - demo
           - iiif
           - staging
           - test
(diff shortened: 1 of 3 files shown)


ShanaLMoore (approved):

ShanaLMoore:
i think we have a hykutest.org too, that may need this. @maxkadel

maxkadel:
> i think we have a hykutest.org too, that may need this. @maxkadel

The ticket only mentioned these two environments - I assumed because they're much older and have a lot more users on them https://github.com/notch8/dev-ops/issues/1422. hykutest.org is pretty new.

Could we maybe get this in and do that as another piece of work if / when we need it? (adding a host requires changes in 3 different places - here, Sendgrid, and the notch8-ops repo in order to actually work)

PR #3164, opened 2026-07-30
Develop on multiple hykus

maxkadel:
When running multiple instances of Hyku on a development machine, the support services get tangled up in each other. So if you're running `my_hyku` and `your_hyku` locally, for some commands `my_hyku` might connect to `your_hyku`'s database, solr, fedora, or redis.

This is at least a partial cause of this bug - Refs https://github.com/samvera-labs/hyku_knapsack/issues/59

This is because by explicitly naming our docker network, any hyku apps that come up will use that same network, and there's not a way for traefik to know where to send traffic. 

If we let docker name the default network, and then name stackcar as the proxy, we can keep the services separate while allowing for the proxy work stackcar does.

⚠️ If adding this change to a knapsack application, you must also change the docker-compose.yaml in that application:

```diff
   zoo:

 networks:
+  # Unnamed: Compose auto-namespaces this per checkout.
   default:
+  # Shared across checkouts so Traefik (`sc proxy up`) can route them all.
+  proxy:
     name: stackcar
```

@samvera/hyku-code-reviewers

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-APP_NAME=hyku
+# APP_NAME is intentionally left unset here so docker-compose works across checkouts
 CH12N_TOOL=fits_servlet
 CHROME_HOSTNAME=chrome
 COMPOSE_DOCKER_CLI_BUILD=1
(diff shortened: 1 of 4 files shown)


orangewolf (approved):

orangewolf:
Discussed at 8/5/2026 tech call.

PR #3166, opened 2026-07-30
Site instance memoization

maxkadel:
Significantly reduces db calls

In the views Site.instance was being called many times per view, creating a new database call each time. We can memoize the Site.instance per-request in the RequestStore, avoiding cross-site, cross-tenant, and cross-user contamination.

The Admin::WorkTypesController had its own implementation of Site which had to be brought in alignment with Site.instance.

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -35,6 +35,11 @@ class Site < ApplicationRecord
         end
       end
     end
+
+    # Clears the memoized Site so the next call to .instance re-fetches it.
+    def reset!
+      RequestStore.delete(:site_instance)
+    end
   end
 
   # Get all administrator emails associated with this site
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -27,9 +27,12 @@ class Site < ApplicationRecord
              to: :instance
 
     def instance
-      return NilSite.instance if Account.global_tenant?
-      first_or_create do |site|
-        site.available_works = Hyrax.config.registered_curation_concern_types
+      RequestStore.fetch(:site_instance) do
+        return NilSite.instance if Account.global_tenant?
+
+        first_or_create do |site|
+          site.available_works = Hyrax.config.registered_curation_concern_types
+        end
       end
     end
   end
--- a/spec/controllers/sites_controller_spec.rb
+++ b/spec/controllers/sites_controller_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe SitesController, type: :controller, singletenant: true do
 
     context "site with existing directory image" do
       before do
-        expect(Hyku::AvatarUploader).to receive(:storage).and_return(CarrierWave::Storage::File).at_least(3).times
+        allow(Hyku::AvatarUploader).to receive(:storage).and_return(CarrierWave::Storage::File)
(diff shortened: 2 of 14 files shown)


orangewolf:
Discussed via call on 7/24/2026. Max went back to further prove that the site still changes properly between requests

spec/controllers/sites_controller_spec.rb
```diff
@@ -71,7 +71,7 @@
 
     context "site with existing directory image" do
       before do
-        expect(Hyku::AvatarUploader).to receive(:storage).and_return(CarrierWave::Storage::File).at_least(3).times
+        allow(Hyku::AvatarUploader).to receive(:storage).and_return(CarrierWave::Storage::File)
```
orangewolf, line 74: During 8/7/2026 call it was discussed that this change is because the memoized site is no longer called 3 times for one test (yay!)

app/models/site.rb
```diff
@@ -27,11 +27,19 @@ class << self
...
+
+        first_or_create do |site|
+          site.available_works = Hyrax.config.registered_curation_concern_types
+        end
       end
     end
+
+    # Clears the memoized Site so the next call to .instance re-fetches it.
+    def reset!
+      RequestStore.delete(:site_instance)
```
orangewolf, line 41: I think this should be `RequestStore.clear!` to parallel what is implemented here: https://github.com/madebylotus/request_store-sidekiq/blob/master/lib/request_store/sidekiq/server_middleware.rb#L7

app/models/site.rb
```diff
@@ -27,11 +27,19 @@ class << self
              to: :instance
 
     def instance
-      return NilSite.instance if Account.global_tenant?
-      first_or_create do |site|
-        site.available_works = Hyrax.config.registered_curation_concern_types
+      RequestStore.fetch(:site_instance) do
```
orangewolf, line 30: I'm concerned about background jobs. See https://github.com/steveklabnik/request_store#using-with-sidekiq which talks only about Sidekiq but is likely a problem with any background processor.

orangewolf (changes requested):
I think we are missing 2 things here. Currently request_store is a dependency but only because it is a dependency of draper and log rage. We should add it explicitly to the Gemfile if we go with this code. 

Thing 2 is that the docs for request_store specifically state it doesn't handle background jobs properly and only suggest a solution for sidekiq. That's probably because it's a pretty old gem. Today we could easily listen for `active_job.started` or all of `active_job.completed`,`active_job.retry_scheduled`, `active_job.retry_stopped`, `active_job.discarded` and trigger the reset on those.

app/models/site.rb
maxkadel: Added ActiveJob hook to clear RequestStore

app/models/site.rb
maxkadel: I think the scope of the `RequestStore.clear!` would be too broad for this location, I think we just want to delete the one key.

I added the more general `clear!` to an ActiveJob `around_job` hook, which should catch the more specific actions you note (started, completed, retry_scheduled, retry_stopped, and discarded).

orangewolf: sounds good.

orangewolf (approved):
thank you! these changes look great.

PR #3168, opened 2026-07-31
🐛 Fix Download PDF button

kirkkwang:
Prior, the button showed up if any file set in the work was a PDF, but it linked to the work's representative file set, which is often something else, so a work with an image as its representative served that image under a "Download PDF" label.  The permission check used a third file set again, the file_set_id passed into the partial.

Now the partial picks the first PDF file set the user can download and uses it for all three: whether to show the button, what to authorize, and what to link to.  Only downloadable PDFs are considered, since file_set_presenters includes file sets the user cannot see.  At the moment if your work has more than one PDF, only the first one gets downloaded, so perhaps a todo for the future.

Also moves the hardcoded "Download PDF" label into a locale key (hyrax.base.download_pdf, all seven locales) and adds a view spec.

### DIFF
--- a/app/helpers/pdf_js_helper.rb
+++ b/app/helpers/pdf_js_helper.rb
@@ -5,9 +5,11 @@ module PdfJsHelper
     "/pdf.js/viewer.html?file=#{path}##{query_param}"
   end
 
-  def pdf_file_set_presenter(presenter)
-    # currently only supports one pdf per work, falls back to the first pdf file set in ordered members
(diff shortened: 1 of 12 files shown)


ShanaLMoore:
Should [this](https://github.com/samvera/hyku/blob/main/app/helpers/pdf_js_helper.rb#L10) get updated too?

kirkkwang:
@ShanaLMoore Thanks for the catch, it's adjusted as well now

laritakr (approved):
Looks great!

PR #3181, opened 2026-08-06
Practice research theme

kirkkwang:
<details><summary>Home page (default)</summary>
image
</details>

<details><summary>Show page (default)</summary>
image
</details>

<details><summary>Home page (blue and yellow)</summary>
image
</details>

<details><summary>Show page (blue and yellow)</summary>
image
</details>

<details><summary>Home page (green and orange)</summary>
image
</details>

<details><summary>Show page (green and orange)</summary>
image
</details>

<details><summary>Home page (black and red)</summary>
image
</details>

<details><summary>Show page (black and red)</summary>
image
</details>

<details><summary>Home page (purple and gold)</summary>
image
</details>

<details><summary>Show page (purple and gold)</summary>
image
</details>

Adds a **Practice Research** theme pair to Hyku core.

**Home page** — search-led hero, collections/works counts band, featured collections in a tinted band with stacked covers, featured works as thumbnail rows, browse by work type and subject, recently uploaded, featured researcher, about band. Editors reorder both featured grids by drag or keyboard.

**Show page** — tabbed body that leads with the narrative (context statement → additional information → child works → files, each pane paginated in deposit order), with a sidebar for PDF download, contributors/relationships cards from the compound JSON fields, and a citation picker with copy + EndNote export.

Also fixes two latent bugs that surfaced along the way, both affecting all themes:
- the homepage collections query sorted without a direction, which Solr rejects — the rescue silently turned that 400 into an empty `@collections` everywhere
- the parent-works rows rendered outside the relationships `<dl>` with stray `</td>` fragments; they now match the structure of Hyrax's own `_relationships_parent_row`, so each show theme's `dl.work-show` styles reach them (worth a visual glance on cultural/scholarly)

Two non-obvious bits preserved from stripped comments: `_controls.html.erb` passes `locale: nil` to `current_page?` because Hyrax merges the locale into `default_url_options`, so without it the Home tab never highlights on a bare `/`; and the featured re-filter exists because `FeaturedWorkList`/`FeaturedCollectionList` load presenters with a nil ability, so a work featured while public keeps rendering after being made private.

Strings ship in en/de/es/fr/it/pt-BR/zh. Covered by presenter, helper, controller, request, and feature specs.

https://claude.ai/code/session_01GdeMyefFsyep4fgFrcUjCU


app/presenters/practice_research/homepage_presenter.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    def model_counts
+      works = ::Hyrax::ModelRegistry.work_rdf_representations
+
+      facets.aggregations['has_model_ssim']
+            .items
+            .select { |item| works.include?(item.value) }
+            .to_h { |item| [item.value, item.hits] }
+    end
```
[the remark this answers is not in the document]
kirkkwang: `Hyrax::ModelRegistry.work_rdf_representations` returns `to_rdf_representation` values, which are the same strings `has_model_ssim` indexes — on this codebase's models that is `["GenericWork", "Image", "Etd", "Oer"]`, not RDF URIs. `spec/presenters/practice_research/homepage_presenter_spec.rb` exercises this against a real Solr core and asserts the work types module populates from an indexed work.

app/controllers/hyrax/homepage_controller.rb
```diff
@@ -80,6 +80,17 @@ def all_collections
...
+    # add this method to vary blacklight config and user_params
+    def search_service(*_args)
+      Hyrax::SearchService.new(
+        config: ::CatalogController.new.blacklight_config,
+        user_params: params.except(:q, :page),
+        scope: self,
+        current_ability:,
+        search_builder_class:
+      )
+    end
```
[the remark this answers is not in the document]
kirkkwang: This restores the visibility the method inherits: `Blacklight::Searchable#search_service` is public API on every catalog controller (it's the first method in that module, above any `private`), and this controller's override had shadowed a public method with a private one. Hyku has no routes that map `:action` dynamically, so it is not reachable as an action, and the service is constructed around `current_ability`, so callers cannot widen access through it.

app/presenters/hyku/work_show_presenter.rb
```diff
@@ -123,8 +124,37 @@ def valid_child_concerns
...
 
+    def partitioned_authorized_ids
+      @partitioned_authorized_ids ||= begin
+        models = authorized_member_models
+        file_set_models = ::Hyrax::ModelRegistry.file_set_rdf_representations
+
+        authorized_item_ids.select { |id| models.key?(id.to_s) }
+                           .partition { |id| file_set_models.include?(models[id.to_s]) }
+      end
+    end
```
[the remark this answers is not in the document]
kirkkwang: `file_set_rdf_representations` returns `["FileSet"]`, which is exactly what `has_model_ssim` holds for file set documents. `file_set_class_names` returns `["::FileSet", "::Hyrax::FileSet", "Hyrax::FileSet"]` — those constant-style strings never appear in `has_model_ssim`, so switching to them would break the partition (`HykuIndexing#file_set_model_filter` uses them to build a different kind of query). `spec/requests/practice_research_show_theme_spec.rb` proves the file set / child work split against a real Solr core.

app/assets/stylesheets/themes/practice_research/_primitives.scss
```diff
@@ -0,0 +1,75 @@
+// ---- practice research mixins ----
+
+@mixin pr-full-bleed {
+  margin-inline: calc(50% - 50vw);
+  padding-inline: calc(50vw - 50%);
+}
+
+@mixin pr-container-inset {
+  padding-left: 15px !important;
```
laritakr, line 9: In general, we try to not use !important as it can break other things. I'm curious if all of these are actually required?

laritakr (commented):
I love the style. I think everything looks pretty good but I am out of time today to review. No blocking things from me though if someone else reviews it. 

Just a few notes:
- might be good to ensure that things adjust gracefully if the metadata profile doesn't have the various properties being used. Did we test with a bare-bones metadata profile? Maybe a followup ticket to check this?
- I noticed odd citation data depending on how the date was entered... date isn't a fixed format so perhaps a bit more cleanup to be sure you have something that makes sense.
- I'd like verification that the use of !Important in the css is actually needed, as that is generally discouraged.
- I wonder about the labeling on the home page (Practice research, kept together
Artworks, documentation and writing, deposited as whole projects, in the order their makers intended.) Should this use one of the fixed text blocks so it can be varied in the UI? I suppose the use of the yml file is adequate but just wasn't sure which was better.

app/assets/stylesheets/themes/practice_research/_primitives.scss
kirkkwang: Audited every one — 38 declarations across the four files, and each survives for one of three reasons:

1. **Bootstrap utility classes in shared markup** (the largest group): the stock `_controls` partial used on show pages carries `bg-light`/`px-2`/`py-3`/`border-bottom`, `_user_util_links` carries `px-3`, the shared footer carries `p-1`/`p-2` and alignment utilities, and the masthead carries spacing utilities. Bootstrap 4 utilities all ship `!important`, so overriding them requires it — that's also why the same pattern appears throughout the existing theme stylesheets.
2. **`_appearance_styles`**: it's an inline `<style>` emitted after `application.css` and pins font families (body + every h1–h6), footer colors, `.card` borders, and `.nav-tabs a:hover` backgrounds with its own `!important` — specificity can never beat a later-emitted `!important`, only another one.
3. **Blacklight's `.navbar + .container { @extend .py-2 }`** — a utility extend, so `!important` — which the two `#content-wrapper { padding-top: 0 }` rules counter.

The audit did surface one self-inflicted case: `.pr-prose > :last-child` was out-shouting an `mb-4` in the theme's own `_featured_attributes` override. Removed the utility class instead and dropped that `!important` (now folded into the show-theme commit).

kirkkwang:
@laritakr thanks for taking a look! Addressed each note:

- **Bare-bones metadata profile**: the theme is defensive by design: every property read goes through `try`, `pr_metadata_fields` skips anything the presenter doesn't respond to, the context tab only appears when a featured field has content, and every homepage module hides when empty, so in theory it shouldn't blow up.
- **Citation dates**: yeah the citations do look a little odd, which come straight from Hyrax's `CitationsBehaviors` formatters, so this theme I did some work in Hyrax recently to fix at least the names and stray commands and colons but a date rework would definitely be a larger leg of work given how funky dates can get
- **`!important` audit**: details in the inline thread. Short version: all but one counter something that is itself `!important` (Bootstrap utilities in shared markup, `_appearance_styles`' late-emitted pins, Blacklight's `.py-2` extend).  Fixed one issue however so good call on checking that.
- **Hero copy**: good call because that was the intention and I missed one so now the *Home page text* content block now controls the entire hero copy (title and lede together), with the yml strings only as the fallback when the block is empty.  Headings inside the block pick up the hero title treatment automatically, so formatting it is just writing an `<h1>` in the block.

image
image

ShanaLMoore:
Ran axe against this branch with seeded content (tenant with 10 works, featured works set, both themes selected) while doing an accessibility pass across all the Hyku themes. Two things trace to the new theme itself:

- The footer stat text (`.pr-stats` area, `#a2a2a2` on `#3c3c3c`) measures 4.32:1 at 13px, just under the 4.5:1 AA bar. It's close enough that nudging the gray a step lighter (or bumping the font size to 18px+/14px bold, where the 3:1 large-text threshold applies) clears it.
- The hero blocks (`.pr-hero-eyebrow`, `.pr-hero-title`, `.pr-hero-lede`, `.pr-hero-links`, `.pr-stats`) sit outside any landmark, which axe flags as moderate `region` violations. Wrapping the hero in a `<section aria-label>` or similar would cover all five.

You'll also see a `link-in-text-block` hit on the footer Samvera link and two breadcrumb contrast hits on the show page - those are stock defects that exist on main too, and #3197 fixes them globally, so no need to chase them here.

Everything else scanned clean, and honestly the homepage reads better than any of the shipped themes. Nice work.

app/assets/stylesheets/themes/practice_research.scss
```diff
@@ -0,0 +1,1355 @@
...
+
+  .p-1,
+  .p-2,
+  .p-lg-2 {
+    padding: 0 !important;
+  }
+
+  .navbar-text,
+  .navbar-text div {
+    color: color-mix(in srgb, var(--pr-band-text, #ffffff) 64%, transparent) !important;
```
ShanaLMoore: axe measures this footer text at 4.32:1 (13px text needs 4.5:1) once the appearance styles' band background applies. 78% keeps the muted look and clears AA on both the theme band and the appearance default `#3c3c3c`:

```suggestion
    color: color-mix(in srgb, var(--pr-band-text, #ffffff) 78%, transparent) !important;
```

app/views/themes/practice_research/layouts/homepage.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for(:navbar) do %>
+  <%= render '/controls' %>
+  <% if controller_name == 'homepage' %>
+    <%= render 'hyrax/homepage/hero' %>
+    <%= render 'hyrax/homepage/stats' %>
```
ShanaLMoore: The hero and stats render straight into `body` via `content_for(:navbar)`, outside every landmark, so axe flags `.pr-hero-eyebrow`/`-title`/`-lede`/`-links` and `.pr-stats` with moderate `region` violations. The page has no banner landmark yet (the masthead is a `nav`), so wrapping both in a `header` covers all five nodes without new i18n:

```suggestion
    <header>
      <%= render 'hyrax/homepage/hero' %>
      <%= render 'hyrax/homepage/stats' %>
    </header>
```

ShanaLMoore (commented):
Follow-up to my axe findings above - here are both fixes as suggestions so they can be committed directly.

app/assets/stylesheets/themes/practice_research.scss
kirkkwang: Applied as suggested — 78% clears AA on both the theme band and the appearance default. Folded into the home page theme commit.

app/views/themes/practice_research/layouts/homepage.html.erb
kirkkwang: Applied — hero and stats now render inside a `header`, which gives the page its banner landmark. Folded into the home page theme commit.

kirkkwang:
Thanks @ShanaLMoore!  should be fixed now

- footer stat text bumped to 78% of the band text color, which clears 4.5:1 on both the theme band and the appearance default `#3c3c3c`
- hero + stats now render inside a `<header>`, giving the page its banner landmark and covering all five `region` nodes

ShanaLMoore (approved):

PR #3182, opened 2026-08-06
Upgrade patch version of Rails

maxkadel:
See https://github.com/rails/rails/security/advisories/GHSA-xr9x-r78c-5hrm

There is a vulnerability for applications that use ActiveStorage with vips, which riiif does. This is patched in Rails 7.2.3.2.

@samvera/hyku-code-reviewers


orangewolf (approved):
As discussed on the call this morning - we do not actually have exposure to this, but its good to patch anyway.

PR #3183, opened 2026-08-06
Guided Deposit Adjustments

laritakr:
## Summary
Fixes three ways the guided deposit wizard could lose a depositor's work, adds a way to abandon a deposit, and shows vocabulary terms by name instead of as URLs.

## Details
- Advancing past the upload step mid-upload silently dropped in-flight files; Next is now disabled until uploads settle, including for a file cancelled before it is ever sent.
- The detail steps discarded anything typed when Back was clicked, and the work detail step discarded the whole form when validation failed. Both now save what was entered; validation only decides where the depositor goes next.
- Every step offers "Discard this deposit", which clears the collected state and deletes the files staged for it. Previously the only way out was to re-enter the wizard, which nothing advertised, and also which left orphaned files.
- Nothing validated that a chosen parent could contain the work being deposited, and `Steps::AddToParent` checks neither the type pairing nor access to the parent. The wizard now refuses a parent whose type cannot hold the work, and — since `parent_id` arrives in params — one the depositor cannot edit. The standard deposit form applies the same check on create, which **changes existing behavior there**: posting a `parent_id` for a work you cannot edit was previously allowed.
- A child inherits its parent's admin set when the start screen is skipped, but only when the depositor may deposit into it: edit access to a work does not imply deposit rights to the admin set it sits in. Otherwise the parent is kept and the depositor stays on the start screen to choose one, rather than silently getting the default.
- The review summary shows an authority's label rather than the stored id, driven by the metadata profile so no property is named in code.
- Two changes only apply to installs that configure them, and are inert otherwise: a deposit started from an "attach a child" link opens on the parent step (needs `parent_connect_placement = :start`), and the path that deposits a container leads the start screen full-width (needs `container_type`).

## Testing
- [ ] Turn on `enable_guided_deposit`. Upload a large file on the upload step: Next stays disabled until it completes. Cancel an upload mid-flight and confirm Next re-enables rather than sticking. Queue a file and cancel it before it starts sending — Next should re-enable there too.
- [ ] On the work detail step, type into several fields and click Back, then forward again — the text is still there. Repeat with a required field left blank: the validation error shows, and navigating away and back still retains what was typed. Then repeat both on the file detail step.
- [ ] Click "Discard this deposit" with a file uploaded: it returns to the dashboard, the wizard starts clean, and the staged file is gone.
- [ ] From a work's show page, use "Add child work" and complete the deposit — it should succeed and the child should be nested.
- [ ] Set a work type's `valid_child_concerns` to `[]`, then try to select one of those works as a parent — the parent step refuses it.
- [ ] As a non-admin, try to select a work you cannot edit as a parent — it should be refused. Then post to a work controller with that same `parent_id`; it should also be refused.
- [ ] As a non-admin with edit access to a work in an admin set you cannot deposit into, follow that work's "attach a child" link: you should land on the start screen with an admin-set chooser rather than being skipped past it, and the parent should still be carried through.
- [ ] Enter a rights statement on the work detail step and confirm the review page names it rather than showing the URL.
- [ ] Check the Back and Next buttons still behave on the upload, parent, item type, and review steps — the shared footer changed shape.

### DIFF
--- a/app/assets/javascripts/hyrax/deposit_wizard.js
+++ b/app/assets/javascripts/hyrax/deposit_wizard.js
@@ -23,6 +23,40 @@
       uploadTemplateId: 'deposit-wizard-template-upload',
       downloadTemplateId: 'deposit-wizard-template-download'
     });
+    guardNextWhileUploading(uploader);
+  }
+
+  // The uploaded_files[] hidden input is written only by the download template,
+  // which the plugin swaps in on completion — so advancing mid-upload posts
+  // without the in-flight ids and silently drops those files.
+  function guardNextWhileUploading(uploader) {
+    var next = $('[data-behavior="files-next"]');
+    if (!next.length) return;
+    var inFlight = 0;
+
+    function sync() {
+      next.prop('disabled', inFlight > 0);
+    }
+
+    // Decrement on 'finished', not 'completed': completed fires only on success,
+    // so a failed upload would leave the count above zero and Next stuck for the
+    // rest of the page's life. 'finished' fires on both the success and failure
+    // paths. (Hyrax's own save_work/uploaded_files.es6 counts 'completed' and has
+    // exactly that leak — don't mirror it.)
+    uploader.on('fileuploadadded', function () {
+      inFlight += 1;
+      sync();
+      // No explicit return: the plugin cancels the upload when an 'added'
+      // handler returns false.
+    });
+    uploader.on('fileuploadfinished', function () {
+      inFlight = Math.max(0, inFlight - 1);
+      sync();
+    });
+
+    uploader.closest('form').on('submit', function (event) {
+      if (inFlight > 0) event.preventDefault();
+    });
   }
 
   // Visibility pills: mark the selected pill, expand its embargo/lease panel (by
--- a/app/assets/javascripts/hyrax/deposit_wizard.js
+++ b/app/assets/javascripts/hyrax/deposit_wizard.js
@@ -32,30 +32,36 @@
   function guardNextWhileUploading(uploader) {
     var next = $('[data-behavior="files-next"]');
     if (!next.length) return;
-    var inFlight = 0;
+    // Tracked per file rather than as a counter so the settle events can overlap
+    // without double-counting: an aborted upload fires both 'fail' and 'finished'.
+    var pending = [];
 
-    function sync() {
-      next.prop('disabled', inFlight > 0);
+    function settle(data) {
+      var files = (data && data.files) || [];
+      for (var i = 0; i < files.length; i += 1) {
+        var at = pending.indexOf(files[i]);
+        if (at !== -1) pending.splice(at, 1);
+      }
+      next.prop('disabled', pending.length > 0);
     }
 
-    // Decrement on 'finished', not 'completed': completed fires only on success,
-    // so a failed upload would leave the count above zero and Next stuck for the
-    // rest of the page's life. 'finished' fires on both the success and failure
-    // paths. (Hyrax's own save_work/uploaded_files.es6 counts 'completed' and has
-    // exactly that leak — don't mirror it.)
-    uploader.on('fileuploadadded', function () {
-      inFlight += 1;
-      sync();
+    uploader.on('fileuploadadded', function (e, data) {
+      pending = pending.concat((data && data.files) || []);
+      next.prop('disabled', pending.length > 0);
       // No explicit return: the plugin cancels the upload when an 'added'
       // handler returns false.
     });
-    uploader.on('fileuploadfinished', function () {
-      inFlight = Math.max(0, inFlight - 1);
-      sync();
+
+    // 'finished' covers success and a mid-flight abort, but a file cancelled before
+    // it is ever sent has no data.abort, so the plugin fires 'fail' alone — without
+    // that second binding Next would stay disabled until a reload. ('completed'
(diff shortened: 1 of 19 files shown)


ShanaLMoore (commented):
Really nice work on this one. The commit-by-commit split makes it easy to follow, and the comments explain the reasoning rather than restating the code, which made reviewing a 750-line diff genuinely pleasant. The `fileuploadfinished` vs `completed` note in particular saved me from having to work out why on my own.

One blocker on the new `create` guard, plus a few questions inline that aren't blocking.

Also worth noting: both Copilot comments were left against b22e03fd, and 6f82e285 added the `can?(:edit, parent)` checks they asked for. They're stale.

---

_Review assisted by Claude Code. The class-name mismatch above was verified by running it against a local Hyku instance._

app/controllers/concerns/hyku/works_controller_behavior.rb
```diff
@@ -38,6 +41,26 @@ def json_manifest
...
+    # parent_id reaches Steps::AddToParent straight from params, and that step
+    # validates neither the type pairing nor the user's access to the parent.
+    def ensure_parent_accepts_child
+      parent_id = params[:parent_id]
+      return if parent_id.blank?
+
+      parent = Hyrax.query_service.find_by(id: parent_id)
+      child_types = Hyrax::ChildTypes.for(parent: parent.class).map(&:to_s)
+      return if current_ability.can?(:edit, parent) &&
+                child_types.include?(self.class.curation_concern_type.to_s)
```
ShanaLMoore: This rejects every legitimate attach-child create. `valid_child_concerns` defaults to `Hyrax.config.curation_concerns`, which in Hyku is the ActiveFedora classes, but `curation_concern_type` on all four work controllers is the Valkyrie resource. So the two sides never match:

```
child_types           = ["GenericWork", "Image", "Etd", "Oer"]
curation_concern_type = "GenericWorkResource"
guard passes?         = false
```

`POST /concern/parent/:parent_id/generic_works`, i.e. the "Add child work" action off a work's show page, now redirects to root with "That work cannot contain this deposit." Same for Etd/Image/Oer.

The wizard side is fine, since `nesting_allowed?` compares `state.work_type`, which comes from `QuickClassificationQuery#authorized_models` and is the same AF name.

Both new controller examples assert a rejection, so nothing caught this. A positive case (valid parent, create succeeds) would have, and is worth adding whichever way you normalize the comparison.

app/controllers/hyrax/deposit_wizard_controller.rb
```diff
@@ -216,13 +241,60 @@ def assign_admin_sets_for_standard_chooser
...
+    # Skipping `start` skips its inline admin-set chooser, so a child work takes its
+    # parent's admin set. Only from a parent the depositor can edit, and only into an
+    # admin set they may deposit into: parent_id arrives in params, and edit on a work
+    # does not imply deposit rights to the admin set it sits in.
+    #
+    # False when the depositor must choose an admin set themselves, so the caller can
+    # keep them on the start screen rather than silently defaulting one.
+    def inherit_admin_set_from_parent
+      parent = Hyrax.query_service.find_by(id: wizard_state.parent_id)
+      return true unless current_ability.can?(:edit, parent)
```
ShanaLMoore: Should this return `false` instead? The return value means "no chooser needed", so a parent the depositor can't edit skips the start screen's admin-set chooser and quietly takes the default, which seems like the one case the stay-on-start behavior was meant for. Commit still refuses the parent so nothing gets through, and it's only reachable under `parent_connect_placement = :start`. The spec below checks `admin_set_id` ends up blank but not that they stay on start.

app/assets/javascripts/hyrax/deposit_wizard.js
```diff
@@ -23,6 +23,40 @@
...
+    // rest of the page's life. 'finished' fires on both the success and failure
+    // paths. (Hyrax's own save_work/uploaded_files.es6 counts 'completed' and has
+    // exactly that leak — don't mirror it.)
+    uploader.on('fileuploadadded', function () {
+      inFlight += 1;
+      sync();
+      // No explicit return: the plugin cancels the upload when an 'added'
+      // handler returns false.
+    });
+    uploader.on('fileuploadfinished', function () {
```
ShanaLMoore: Does `finished` fire when a queued file is removed before it's ever sent, as opposed to cancelled mid-flight? If not, `inFlight` stays above zero and Next is stuck until a reload, which is a rough failure mode for a change about not losing work. Your test plan covers cancel-mid-flight but I didn't see the before-send case.

app/controllers/hyrax/deposit_wizard_controller.rb
```diff
@@ -75,6 +77,13 @@ def save_extras
       head :no_content
     end
 
+    def discard
+      deleted = discard_staged_uploads
+      reset_state
```
ShanaLMoore, line 82: Worth clearing `session[:deposit_wizard_last]` here too, or does leaving the previous deposit's stash around not matter since the done screen reads it once?

ShanaLMoore (approved):

app/controllers/concerns/hyku/works_controller_behavior.rb
```diff
@@ -38,6 +41,31 @@ def json_manifest
 
     private
 
+    # parent_id reaches Steps::AddToParent straight from params, and that step
```
orangewolf, line 44: It seems really inefficient to look everything up here. Aren’t we loading the same parent and child to actually save them too?

PR #3189, opened 2026-08-07
🐛 Source seed passwords from ENV instead of hardcoding

kirkkwang:
Split out of #3128 (1 of 4), per review. Original work by @SixtyGrit.

`RolesService.seed_superadmin!` and `seed_qa_users!` created users with the well-known password `testing123`, leaving superadmin accounts with published credentials in any long-lived dev, QA, or demo environment. Both methods now read the password from `HYKU_SEED_PASSWORD` via `ENV.fetch` and raise `KeyError` when it is unset — no fallback, the operator sets it explicitly. Existing users are never touched and the production/staging guards are unchanged.

The checked-in `.env` sets `HYKU_SEED_PASSWORD=testing123` so the local docker workflow keeps working out of the box.


app/services/roles_service.rb
```diff
@@ -227,6 +229,30 @@ def seed_qa_users!
...
+    #
+    # @return [String]
+    def seed_user_password
+      ENV.fetch('HYKU_SEED_PASSWORD', nil).presence || generated_seed_password
+    end
+
+    ##
+    # @api private
+    # @return [String] a newly generated random password
+    def generated_seed_password
```
kirkkwang: no fallback to this, get rid of it, set explicitly through ENV

orangewolf:
reviewed on 8/11/26 call

app/services/roles_service.rb
```diff
@@ -227,6 +227,16 @@ def seed_qa_users!
...
     # rubocop:enable Metrics/MethodLength
+
+    ##
+    # Password assigned to users created by {.seed_superadmin!} and
+    # {.seed_qa_users!}.
+    #
+    # @return [String]
+    # @raise [KeyError] when HYKU_SEED_PASSWORD is not set
+    def seed_user_password
+      ENV.fetch('HYKU_SEED_PASSWORD')
```
orangewolf:
```suggestion
      @seed_user_password ||= ENV.fetch('HYKU_SEED_PASSWORD')
```

orangewolf (changes requested):
tiny thing but worthy. ENV won't change once the app is loaded. the rest looks great

PR #3190, opened 2026-08-07
🐛 Stop defaulting activated users' password to 'password'

kirkkwang:
Split out of #3128 (2 of 4), per review. Original work by @SixtyGrit.

`Admin::UsersController#activate` fell back to the literal string `password` when `HYKU_USER_DEFAULT_PASSWORD` was unset, so any deployment that never set the variable assigned that password to every user an admin activated. The action now uses the variable when present and otherwise assigns a random `Devise.friendly_token` placeholder; the activated user sets a real password through the standard reset flow. Also documents the previously undocumented `HYKU_USER_DEFAULT_PASSWORD`.

Deployments that set the variable (including the checked-in `.env` for local docker) see no behavior change.


orangewolf (approved):

orangewolf:
discussed on call 8/7/2026

PR #3191, opened 2026-08-07
🐛 Make the HTTP basic auth gate configurable and fail closed

kirkkwang:
Split out of #3128 (3 of 4), per review. Original work by @SixtyGrit.

`ApplicationController#authenticate_if_needed` guarded hidden (non-public) tenants and staging environments with credentials hardcoded to `samvera` / `hyku`, which could not be changed without patching application code.

The `http_basic_auth_username` / `http_basic_auth_password` class attributes now default from the `HYKU_BASIC_AUTH_USER` / `HYKU_BASIC_AUTH_PASSWORD` environment variables, falling back to the historical `samvera` / `hyku` values, so deployments can rotate the credentials without a code change and unconfigured deployments behave exactly as before. Comparison uses `ActiveSupport::SecurityUtils.secure_compare`.


orangewolf (changes requested):
Please:
1 - maintain the existing defaults, no need to cause a breaking change
2 - either do the the class attributes or the method. both are overkill.

orangewolf (approved):
I love that this is now a 4 line change. thanks for cleaning all that up

PR #3192, opened 2026-08-07
🎁 Add an optional in-app spam challenge to user signup

kirkkwang:
Split out of #3128 (4 of 4), per review. Original work by @SixtyGrit; reworked per review to reuse the contact form's negative captcha.

`allow_signup` defaults to true per tenant and the registration form had no bot protection, so a public demo tenant collects unattended bot registrations. A new per-tenant boolean setting, `signup_spam_protection` (default false, ENV fallback `HYKU_SIGNUP_SPAM_PROTECTION`), puts the registration form behind the same `negative_captcha` protection the contact form already uses: the real inputs render under per-request hashed names while hidden decoy inputs carry the real names, so submissions that fill a decoy — or POST plain `user[...]` params directly — are rejected. The captcha setup shared by both controllers lives in a new `NegativeCaptchaBehavior` concern.

Rejected submissions redirect back to the signup form with a generic, translatable message. The challenge runs entirely in-app: no third-party service, no cookies, no fingerprinting, no new gems (`negative_captcha` was already in the Gemfile). Both registration view variants render the same shared fields partial, and the form is unchanged when the challenge is off. Accounts flagged `public_demo_tenant` enforce the challenge unconditionally. The existing `allow_signup` setting is untouched.


orangewolf (changes requested):
Love this idea. But we already have something similar on the contact page... why are we reinventing spam protection for this form instead of using the same method and consolidating code?

orangewolf (approved):

PR #3193, opened 2026-08-07
Add bullet gem

maxkadel:
Adds bullet in development and based on an env variable to allow us to catch ActiveRecord N+1s before they hit production.

Passive logging only (Bullet.rails_logger = true, no Bullet.raise) so a
detected N+1 never turns into a broken request on staging, which real
testers/demos hit. Footer only in development.

Note: only catches genuine ActiveRecord association N+1s - Valkyrie's
Postgres-backed orm_resources lookups go through explicit find calls, not
AR association loading, so Bullet won't see those.


orangewolf (approved):
We discussed via call on 8/7/2026.

PR #3194, opened 2026-08-07
Add db-query-matchers testing gem for dev/test environments 

maxkadel:
## Summary
Adds the `db-query-matchers` gem + a `spec/support/db_query_matchers.rb` config file (excludes one-time schema-introspection queries from counts via `schemaless = true`).

## Why its own PR
Several N+1-fix branches want this same gem for their own regression specs (an in-progress `fix_user_hyrax_groups_n_plus_one`, `fix_hyrax_group_site_role_n_plus_one`, and #3166). Landing it here first, standalone, avoids three separate branches each independently creating the same new file and touching `Gemfile`/`Gemfile.lock` - a new file added on diverged branches is a classic git add/add conflict, not something that reliably auto-resolves in a 3-way merge just because the content is identical. The other branches will rebase onto `main` once this merges and drop their now-duplicate copies.

No app code changes, test-only.


spec/support/db_query_matchers.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'db_query_matchers'
```
orangewolf: this is not where require statements go ;-)

orangewolf (changes requested):

spec/support/db_query_matchers.rb
orangewolf: it likely works just the same w/o the require statement

orangewolf:
this got merged already. a seperate PR would have been better because it would track to release notes, but that's life

PR #3195, opened 2026-08-07
Stop rendering literal "false" in tenant page chrome

ShanaLMoore:
### What

An out-of-the-box tenant renders a literal `false` into user-visible chrome, and the default-theme logo ships with no alt attribute:

* The cultural theme homepage prints `false` where its Recent Works empty-state message should be.
* Every stock homepage hero renders `title="false"` on the banner container.
* `app/views/_logo.html.erb` passes `alt_text:` (not an `image_tag` option), so the logo `` has no alt at all; the themed logo partials render `alt="false"` when no alt text is configured.

### Why it happens

Three related leaks with one root pattern (falsy placeholders reaching ERB):

1. `ApplicationHelper#missing_translation` shadowed Rails' private `TranslationHelper` hook of the same name. Rails calls that hook for every missing view translation, so **any** missing key rendered as literal `false` instead of Rails' standard missing-translation span. Removed; `locale_for`/`hint_for` and the appearance color-input hint now use `I18n.t(..., default: nil)` and keep their existing fallback semantics (covered by specs).
2. `block_for` returned literal `false` for unset content blocks. It now returns `nil`; every other caller already guards with `||`.
3. The cultural theme's `_recently_uploaded` partial is rendered by its explicit `themes/...` path (from `_recent_works_section`), so lazy i18n lookup resolves under `themes.*` where no keys exist. It now uses the absolute `hyrax.homepage.recently_uploaded.no_public` key. The other theme variants resolve through the injected theme view roots and are unaffected.

The logo partials now pass `alt:` and fall back to the institution's `application_name` when no alt text is configured, so the logo is never unlabeled.

### Testing

Failing-spec-first history: view specs for the cultural empty state and `_logo`, helper specs for `block_for` nil semantics, the `locale_for` simple_form fallback, and a regression spec asserting missing view translations never render `false`. Affected suites run locally in Docker (helpers, content block, homepage views, appearance views, forms, cultural theme feature spec) plus RuboCop on changed files: green.

### Placement

samvera/hyku core: the shadowed helper, the content-block helper, and the affected partials all live here (generally useful, not hykuUP-specific).

### Screenshots (1440px and 375px)

The shots use an empty tenant with the **cultural_repository** theme selected: the visible `false` only manifests there, because that theme is the one rendering the affected partial by its explicit path. Look for the "Recent Works" section: `false` before, "No public work has been contributed." after.

<details><summary>Before - desktop</summary>

Before - desktop

</details>

<details><summary>Before - mobile</summary>

Before - mobile

</details>

<details><summary>After - desktop</summary>

After - desktop

</details>

<details><summary>After - mobile</summary>

After - mobile

</details>


app/views/hyrax/admin/appearances/_color_input.html.erb
```diff
@@ -1,16 +1,16 @@
 <div class='row'>
   <div class='col-lg-10'>
-    <% hint = t("hyrax.admin.appearances.show.forms.#{color_name}.hint") %>
+    <% hint = I18n.t("hyrax.admin.appearances.show.forms.#{color_name}.hint", default: nil) %>
```
orangewolf: remove the I18n here please

app/helpers/application_helper.rb
```diff
@@ -79,28 +79,21 @@ def alttext_for(collection)
...
-    hint = locale_for(type: 'hints', term:, record_class:)
-
-    return hint unless missing_translation(hint)
+    locale_for(type: 'hints', term:, record_class:)
   end
 
+  # NOTE: do not define a helper named missing_translation here. Rails'
+  # ActionView::Helpers::TranslationHelper calls a private hook by that name for
+  # every missing view translation; an ApplicationHelper method shadows it and
+  # its return value gets rendered into the page (a literal "false", historically).
```
orangewolf: this code comment is about a bug that you are removing. its a comment about the past transient state of the code and is more likely to be confusing to future readers than helpful IMO

orangewolf (changes requested):

ShanaLMoore:
Agree with your feedback @orangewolf

ShanaLMoore:
Addressed in b0d481b5: the hint lookup moved to `ApplicationHelper#color_hint_for`, so the view no longer calls I18n, and the historical comments are gone (helper, spec, and the `block_for` note). Merged main.

ShanaLMoore:
@orangewolf both points addressed in b0d481b5.

- The I18n lookup moved out of the view into `color_hint_for`, so `_color_input.html.erb` just asks for a hint.
- The `missing_translation` comment is gone. The reason it existed is in the commit body now, where it stays attached to the change.

Ready for another look.

PR #3196, opened 2026-08-07
Give cultural-theme recent-works cards alt text, link names, and list markup

ShanaLMoore:
### What

On a cultural-theme tenant homepage with content, every recent-works card fails axe: the thumbnail `` has no alt attribute (**critical**, one per card), the image-only link wrapping it has no discernible name (**serious**, one per card), and the cards render as `<div>`s directly inside the `<ol>` (**serious** `list` violation).

### Fix

`themes/cultural_repository/hyrax/homepage/_recent_document.html.erb` now renders each card as an `<li>` and passes `thumbnail_alt_text_for(recent_document)` through to `render_thumbnail_tag`, which also gives the image-only link its accessible name. The institutional, neutral, and community variants already pass alt text and do not render into an `<ol>`, so they are untouched.

### Testing

Failing-spec-first history: a view spec asserts li markup, thumbnail alt, and the link's accessible name. Cultural theme view + feature specs and RuboCop run locally in Docker: green.

axe on a seeded cultural-theme homepage: 6 critical `image-alt`, 6 serious `link-name`, and 1 serious `list` violations before; all cleared after. (Hyku 7.1 ships a VPAT, so these are regressions worth closing.)

No screenshots: this change is markup-only (alt attributes, accessible names, `li` vs `div`), so it renders pixel-identically before and after. The axe counts above are the evidence; the rendered-HTML difference is visible in the view spec.

### Placement

samvera/hyku core: the theme and its partials live here.


app/views/themes/cultural_repository/hyrax/homepage/_recent_document.html.erb
```diff
@@ -1,11 +1,12 @@
 <%# OVERRIDE Hyrax v5.0.0rc2 template for client theming and shared search %>
-<div class="recently-uploaded col-6 col-md-2">
+<%# li, not div: the collection renders inside the ol in _recently_uploaded %>
+<li class="recently-uploaded col-6 col-md-2">
   <%= link_to [main_app, recent_document] do %>
-    <%= render_thumbnail_tag(recent_document, {class: 'img-fluid center-block'}, {suppress_link: true}) %>
+    <%= render_thumbnail_tag(recent_document, {class: 'img-fluid center-block', alt: thumbnail_alt_text_for(recent_document)}, {suppress_link: true}) %>
```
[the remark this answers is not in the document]
ShanaLMoore: Good catch - that clearing rule would indeed have stopped matching. Went with your second option: dropped the element type so the selector targets `.recently-uploaded:nth-of-type(2n+3)` (30f897f6). The nth-of-type positions are unchanged since the cards are uniform siblings either way; the featured-works selectors keep their `div` type because those cards are still divs.

app/assets/stylesheets/themes/cultural_repository.scss
```diff
@@ -282,7 +282,8 @@
   }
 
   @media screen and (max-width: 992px) {
-    div.recently-uploaded:nth-of-type(2n+3),
+    // no element type: the recent-works cards are li (they render inside an ol)
```
orangewolf: this code comment is about a historical code change. these are an anti pattern

app/views/themes/cultural_repository/hyrax/homepage/_recent_document.html.erb
```diff
@@ -1,11 +1,12 @@
 <%# OVERRIDE Hyrax v5.0.0rc2 template for client theming and shared search %>
-<div class="recently-uploaded col-6 col-md-2">
+<%# li, not div: the collection renders inside the ol in _recently_uploaded %>
```
orangewolf: this code comment is about a historical code change. these are an anti pattern

orangewolf (changes requested):

ShanaLMoore:
@orangewolf agreed, that's annoying/unecessary

ShanaLMoore:
Addressed in a1187b19: both historical comments removed. Merged main.

ShanaLMoore:
@orangewolf both historical comments removed in a1187b19, in `cultural_repository.scss` and `_recent_document.html.erb`.

Ready for another look.

PR #3197, opened 2026-08-07
Make the default link palette meet WCAG AA

ShanaLMoore:
### What

Two out-of-the-box accessibility failures in the default palette:

* The shipped `link_color` default (`#2e74b2`) misses WCAG AA on the backgrounds it renders over: **4.16:1** on Bootstrap's `.breadcrumb` (`#e9ecef`) and **4.41:1** on striped rows (`#f2f2f2`). axe reports serious `color-contrast` violations on work-page breadcrumbs and filename links, and on seeded homepage featured/recently-uploaded title links. (Hyku 7.1 ships a VPAT; these are the standing serious violations on stock pages.)
* Footer links are styled `text-decoration: none` while sitting inside text they don't contrast with by 3:1 (the "A service of Samvera" link measures **1.18:1** against its surrounding text), a serious `link-in-text-block` violation on every page with a footer.

### Fix

* `link_color` default: `#2e74b2` -> `#2a6aa3`, the nearest darker shade that clears 4.5:1 on white (5.69), `.breadcrumb` (4.80), and striped rows (5.09). Tenants that set their own link color through Admin > Appearance are unaffected; per-field restore-default picks up the new value.
* Footer links get `text-decoration: underline` (a non-color distinguisher), removing the explicit `none`.

### Testing

Failing-spec-first history: the appearance form spec now computes WCAG contrast ratios for the shipped default against the three out-of-the-box backgrounds and requires >= 4.5:1. Appearance form + view specs and RuboCop run locally in Docker: green.

axe after the change: work page and seeded homepage report **zero violations** (previously 3 serious color-contrast nodes on the work page, 2-3 on the homepage, plus link-in-text-block).

### Placement

samvera/hyku core: the default palette and footer styles live here.

No screenshots: the visual delta is a ~4% luminance step on link text (`#2e74b2` -> `#2a6aa3`) plus an underline on footer links. Side-by-side page screenshots do not show it legibly. The contrast ratios and axe counts above are the evidence, and the spec asserts the ratios directly so the threshold cannot silently regress.


app/assets/stylesheets/hyku.scss
```diff
@@ -79,8 +79,11 @@ footer.navbar {
         font-weight: 500;
     }
 
+    // Links in the footer sit inside blocks of text whose color they don't
+    // contrast with by 3:1 (axe link-in-text-block), so they need a non-color
+    // distinguisher.
     a {
-        text-decoration: none;
+        text-decoration: underline;
     }
```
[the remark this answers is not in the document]
ShanaLMoore: Right - cultural_repository strips underlines on hover/focus with !important, which would have removed the distinguisher exactly on keyboard focus. Now underlining base, :hover, and :focus with !important (ce772caa); `footer.navbar a:hover` outranks the theme's `a:hover` on specificity, so it holds across themes while tenant custom CSS can still override with a more specific selector.

app/assets/stylesheets/hyku.scss
```diff
@@ -79,8 +79,15 @@ footer.navbar {
...
-        text-decoration: none;
+    // Links in the footer sit inside blocks of text whose color they don't
+    // contrast with by 3:1 (axe link-in-text-block), so they need a non-color
+    // distinguisher. Explicit :hover/:focus with !important because theme CSS
+    // (e.g. cultural_repository's a:hover) strips underlines with !important;
+    // footer.navbar a outranks those selectors on specificity.
+    a,
+    a:hover,
+    a:focus {
+        text-decoration: underline !important;
```
orangewolf: please use a specific enough target instead of adding `!important`. almost nothing is ever that `!important` that you can't ever override it for any reason ;-)

orangewolf (changes requested):

app/assets/stylesheets/hyku.scss
ShanaLMoore: Agree, !important is not ideal

ShanaLMoore:
Addressed in 0c01a00c: dropped the `!important`. The only rule the footer underline had to outrank was cultural_repository's `a:hover, a:focus { text-decoration: none !important }`, so that one loses the flag too - it still beats Bootstrap's `a:hover` on specificity, and `footer.navbar a:hover` now beats it. Comment trimmed as well. Merged main.

ShanaLMoore:
@orangewolf `!important` is gone in 0c01a00c.

The only rule that needed beating was the cultural theme's `a:hover` / `a:focus`, and `footer.navbar a:hover` already outranks it on specificity. So both rules drop the flag and the footer rule wins on its own.

Ready for another look.

orangewolf (approved):
thank you! my war against !important is !important to me ;-)

PR #3211, opened 2026-08-10
Show a visible IIIF manifest link under the viewer

ShanaLMoore:
### What

Work show pages that render in the IIIF viewer expose no visible manifest link anywhere, even though the manifest powers the viewer one iframe away. For the IIIF-literate audience evaluating repository platforms, a visible manifest link is an interoperability proof: drag it onto any IIIF viewer and the object follows.

### Fix

`hyrax/base/_representative_media` renders a small "IIIF manifest" link directly below the viewer, pointing at the work's `manifest_url`. It renders only when the work is IIIF-viewer-backed (`presenter.iiif_viewer?`), so document pages using the PDF.js path and works with no viewer are untouched. The label ships in all seven maintained locales.

### Also fixed: viewer overflow
It also fixes the viewer-wrapper geometry this exposed: the viewer iframe is `position: absolute` with `height: 100%`, and without a positioned wrapper it anchored to a distant ancestor and overflowed its 500px box by ~50px, painting over whatever content followed the viewer (which is why the link was invisible in the first capture). `.viewer-wrapper` is now `position: relative`, so the iframe fills exactly its box.

### Testing

Failing-spec-first: a view spec asserts the link (with the manifest href) when the presenter is IIIF-backed and its absence otherwise. Verified live on a seeded tenant: link renders under the viewer and the manifest URL serves 200 JSON. RuboCop and locale YAML parse clean.

### Placement

samvera/hyku core: the partial is a Hyku override and the behavior is generally useful.

### Screenshots

Before - viewer with no manifest link (1440px):

Before: viewer without manifest link

After - "IIIF manifest" link under the viewer (1440px):

After: IIIF manifest link under the viewer


app/views/hyrax/base/_representative_media.html.erb
```diff
@@ -5,6 +5,12 @@
     <div class="center-block">
       <%= iiif_viewer_display presenter %>
     </div>
+    <%# A visible manifest link: interoperability proof for IIIF-literate
+        audiences (drag it onto any IIIF viewer). Outside the viewer wrapper:
+        its fixed-height container paints over content placed inside it. %>
```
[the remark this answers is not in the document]
ShanaLMoore: Reworded in a28734c3 - the constraint is the absolutely positioned iframe filling the wrapper, so anything else inside the wrapper gets covered; the comment now says that directly.

app/assets/stylesheets/viewer.scss
```diff
@@ -1,5 +1,9 @@
 .viewer-wrapper {
   height: 500px;
+  // The viewer iframe is position: absolute with height: 100%; without a
+  // positioned wrapper it anchors to a distant ancestor and overflows this
+  // box, painting over whatever follows the viewer.
```
orangewolf: this is 3 lines about how css works. that feels like 3 lines too many

app/views/hyrax/base/_representative_media.html.erb
```diff
@@ -5,6 +5,13 @@
     <div class="center-block">
       <%= iiif_viewer_display presenter %>
     </div>
+    <%# A visible manifest link: interoperability proof for IIIF-literate
+        audiences (drag it onto any IIIF viewer). Kept outside the viewer
+        wrapper: the viewer iframe is absolutely positioned to fill the
+        wrapper, so it would cover anything else placed within it. %>
```
orangewolf:
```suggestion
    
```

orangewolf (changes requested):
this makes the iiif link the 4th most prominent thing on the page (above the other metadata and right after the title, the viewer et all). I'm not sure that makes sense to me. but I don't know where else you'd put it so lets clean up the code comments and get it in to get some feedback on it.

ShanaLMoore:
I agree with your feedback @orangewolf

ShanaLMoore:
Addressed in b44199f2: took your suggestion on the manifest-link comment, and viewer.scss is down to a one-line trailing note. Merged main.

ShanaLMoore:
@orangewolf comments trimmed in b44199f2, and the branch is caught up with main as of 5bd274f5.

The merge conflicted in six locale files: main added `hyku.works.errors.work_type_not_offered` at the same insertion point where this branch adds `hyku.work_show.iiif_manifest`. Both keys are kept. `en.yml` merged cleanly on its own and its layout is what the other six now follow. All seven parse and carry both keys.

On placement: taking you at your word to land it and gather feedback. If the manifest link reads too prominent once it is in front of people, the obvious alternative is down with the rest of the metadata, and moving it later is a one-line change.

PR #3220, opened 2026-08-15
Size DB connection pool per-role via DB_POOL, not a flat 50

maxkadel:
`database.yml` hardcoded `pool: 50` for every process. With the Hyrax Helm chart, `DATABASE_URL`'s own hardcoded `?pool=5` also silently overrides it whenever present, so `database.yml`'s setting was dead code regardless. Hit this directly: `RAILS_MAX_THREADS=12` but only 5 real connections per pod, causing `ActiveRecord::ConnectionTimeoutError` under load.

`bin/web`/`bin/worker` now set `DB_POOL` from their own concurrency knob (Puma threads, or the active job queue's thread count), and strip any `?pool=` from `DATABASE_URL` rather than mirroring it — one calculation, one place. `database.yml` reads `DB_POOL` (fallback 10).

Companion chart change: samvera/hyrax#7586 (adds an opt-in `postgresql.poolInUrl: false` so `DATABASE_URL` can omit `?pool=` at the source — this works today independent of that landing).

### DIFF
--- a/bin/db_pool_env.rb
+++ b/bin/db_pool_env.rb
@@ -1,0 +1,20 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Shared by bin/web and bin/worker: sizes DB_POOL from a role-appropriate thread
+# count and strips DATABASE_URL's ?pool= so database.yml's DB_POOL stays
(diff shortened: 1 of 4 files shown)


maxkadel:
Should be able to use updated hyrax helm chart release directly, do not need this workaround.

PR #3222, opened 2026-08-17
🎁 Add columns for staff-managed local vocabularies

ShanaLMoore:
Refs notch8/palni_palci_knapsack#703 (private tracker)

## Why

Adding a local term list today means a developer edits `config/authorities/*.yml` and redeploys, turning a minutes-long metadata task into a days-long developer ticket. This is the schema staff-managed vocabularies need. Database only: no UI and no authority registration yet.

## Worth arguing with

Two changes touch existing behavior. The rest is additive columns and indexes, explained in the commit message.

- `uri` uniqueness moves from global to `[local_authority_id, uri]`. Today two vocabularies in one tenant can't both carry an "Other" term. Strictly looser than what it replaces, so it can't fail on existing rows.
- `uri` is required on new terms and immutable afterward, since works persist it as the term id and reassigning it orphans every record citing the old value.

| action | result |
| --- | --- |
| create a term with no uri, or a blank one | rejected |
| create a duplicate uri within one vocabulary | rejected |
| create the same uri in a different vocabulary | allowed |
| change a term's uri | rejected |
| clear a term's uri | rejected |
| edit label, toggle active, reorder | allowed |
| fill in a uri on a term that has none | rejected |
| several terms with no uri in one vocabulary | allowed |

MeSH is unaffected, since `rake mesh:import_tenant` loads through `insert_all`. File-based vocabularies are untouched and keep working with zero rows in these tables.

## Testing

`bundle exec rspec spec/models/qa spec/authorities/qa spec/db/migrate`, 38 examples, 0 failures.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/models/qa/local_authority_entry.rb
+++ b/app/models/qa/local_authority_entry.rb
@@ -1,7 +1,33 @@
 # frozen_string_literal: true
 
 module Qa
+  # `uri` holds the term identifier and is the value Bulkrax imports. It is not
+  # required to be an actual URI.
   class LocalAuthorityEntry < ApplicationRecord
     belongs_to :local_authority
+
+    store_accessor :data, :alt_labels, :definition
+
+    before_validation { self.uri = nil if uri.blank? }
+
+    validates :uri, presence: true, on: :create
+    validates :uri, uniqueness: { scope: :local_authority_id }, allow_nil: true
+    validate :uri_is_not_reassigned, on: :update
+
+    scope :active, -> { where(active: true) }
+    scope :inactive, -> { where(active: false) }
+    scope :ordered, -> { order(:position, :label) }
+
+    private
+
+    def uri_is_not_reassigned
+      return unless uri_changed?
+
+      errors.add(:uri, 'cannot be changed, because works store it as the term id')
+    end
+
+    scope :active, -> { where(active: true) }
+    scope :inactive, -> { where(active: false) }
+    scope :ordered, -> { order(:position, :label) }
   end
 end
--- a/db/migrate/20260817170001_add_active_data_and_position_to_qa_local_authority_entries.rb
+++ b/db/migrate/20260817170001_add_active_data_and_position_to_qa_local_authority_entries.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddActiveDataAndPositionToQaLocalAuthorityEntries < ActiveRecord::Migration[7.2]
+  def change
+    add_column :qa_local_authority_entries, :active, :boolean, default: true, null: false
+    add_column :qa_local_authority_entries, :data, :jsonb, default: {}, null: false
+    add_column :qa_local_authority_entries, :position, :integer
+
+    add_index :qa_local_authority_entries,
+              %i[local_authority_id active],
+              name: 'index_qa_local_authority_entries_on_authority_and_active'
+
+    add_index :qa_local_authority_entries,
+              %i[local_authority_id position label],
+              name: 'index_qa_local_authority_entries_on_authority_and_sequence'
+  end
+end
--- a/spec/models/qa/local_authority_entry_spec.rb
+++ b/spec/models/qa/local_authority_entry_spec.rb
@@ -3,7 +3,118 @@
 require 'spec_helper'
 
 RSpec.describe Qa::LocalAuthorityEntry, type: :model do
+  let(:authority) { Qa::LocalAuthority.create!(name: 'lab_names') }
+
   it "belongs to a local authority" do
     expect(described_class.reflections['local_authority'].macro).to eq :belongs_to
+  end
+
+  describe 'uri' do
+    it 'is required on a new term' do
+      entry = described_class.new(local_authority: authority, label: 'Anthropology')
+
+      expect(entry).not_to be_valid
+      expect(entry.errors[:uri]).to be_present
+    end
+
+    it 'rejects a blank one on a new term' do
+      expect(described_class.new(local_authority: authority, label: 'Anthropology', uri: '')).not_to be_valid
+    end
+
+    # Works store the uri as the term id, so reassigning it orphans every record
+    # citing the old value. Deactivating the term is the supported alternative.
+    it 'cannot be changed' do
+      entry = described_class.create!(local_authority: authority, label: 'Anthropology', uri: 'anth')
+
+      expect(entry.update(uri: 'anthropology')).to eq false
+      expect(entry.errors[:uri].join).to include('cannot be changed')
+      expect(entry.reload.uri).to eq 'anth'
+    end
+
+    it 'cannot be cleared' do
+      entry = described_class.create!(local_authority: authority, label: 'Anthropology', uri: 'anth')
+
+      expect(entry.update(uri: nil)).to eq false
+      expect(entry.update(uri: '')).to eq false
+    end
+
+    it 'leaves the rest of a term editable' do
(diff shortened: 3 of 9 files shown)


spec/models/qa/local_authority_entry_spec.rb
```diff
@@ -3,7 +3,49 @@
 require 'spec_helper'
 
 RSpec.describe Qa::LocalAuthorityEntry, type: :model do
+  let(:authority) { Qa::LocalAuthority.create!(name: 'lab_names') }
```
laritakr, line 6: Should this do cleanup after the spec completes as well?

db/migrate/20260817170001_add_active_data_and_position_to_qa_local_authority_entries.rb
```diff
@@ -0,0 +1,31 @@
...
+# `position` lets staff pin a display order; entries without one fall back to
+# label order.
+class AddActiveDataAndPositionToQaLocalAuthorityEntries < ActiveRecord::Migration[7.2]
+  def change
+    # Postgres 11+ stores a constant default in the catalog rather than
+    # rewriting the table, so this stays fast on tenants carrying a full MeSH
+    # import (~31k rows).
+    add_column :qa_local_authority_entries, :active, :boolean, default: true, null: false
+    add_column :qa_local_authority_entries, :data, :jsonb, default: {}, null: false
+    add_column :qa_local_authority_entries, :position, :integer
```
laritakr, line 7: If we are adding position for sequence, shouldn't it index on this first, and THEN on id?

app/models/qa/local_authority_entry.rb
```diff
@@ -1,7 +1,23 @@
...
+    # Attributes a term may carry beyond label and identifier. They live in the
+    # `data` hash rather than their own columns so the set can grow without a
+    # migration each time.
+    store_accessor :data, :alt_labels, :definition
+
+    scope :active, -> { where(active: true) }
+    scope :inactive, -> { where(active: false) }
+
+    # Staff-pinned order first, then alphabetical for the terms left unpinned.
+    scope :ordered, -> { order(Arel.sql('position ASC NULLS LAST')).order(:label) }
```
laritakr: Seems like an index doing this in the DB would help it be more efficient.

spec/models/qa/local_authority_entry_spec.rb
ShanaLMoore: `rails_helper` has a global `config.after { DatabaseCleaner.clean }` on the transaction strategy, so these roll back automatically. Nothing to add here.

app/models/qa/local_authority_entry.rb
ShanaLMoore: Added, thanks. `EXPLAIN (ANALYZE, BUFFERS)` on a 30k-term vocabulary, first page of 25:

- without the index: `Seq Scan` over all 30,000 rows plus a top-N heapsort, 1053 buffers
- with it: plain `Index Scan`, 27 buffers

Also dropped the `Arel.sql('position ASC NULLS LAST')` from the scope. Postgres already sorts NULLs last for ASC, so plain `order(:position, :label)` is equivalent and matches the index.

db/migrate/20260817170001_add_active_data_and_position_to_qa_local_authority_entries.rb
ShanaLMoore: Added the index, but I kept `local_authority_id` leading: `[local_authority_id, position, label]`. Every query filters to one vocabulary before it orders, so the equality column goes first and `position, label` then supplies the sort.

I measured both orders rather than guessing. On a 32k-row table they tie (28 vs 29 buffers), because Postgres will happily use a `position`-first index with `local_authority_id` as a non-leading index condition. So the column order isn't what matters here, having the composite index at all is: 27 buffers versus 1053 with no index.

Caveat on that: I tried one dominant vocabulary and a 2k-term vocabulary inside a 32k table, and didn't find a distribution where the two orders diverge. That's "no measured difference," not "proven equivalent." Happy to flip it if you'd still rather lead with position.

app/models/qa/local_authority_entry.rb
```diff
@@ -1,7 +1,15 @@
...
+  # `uri` holds the term identifier and is the value Bulkrax imports. It is not
+  # required to be an actual URI.
   class LocalAuthorityEntry < ApplicationRecord
     belongs_to :local_authority
+
+    store_accessor :data, :alt_labels, :definition
+
+    scope :active, -> { where(active: true) }
+    scope :inactive, -> { where(active: false) }
+    scope :ordered, -> { order(:position, :label) }
```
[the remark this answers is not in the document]
ShanaLMoore: Checked this against Postgres rather than assuming, and it's the other way around: ASC sorts NULLs **last** by default.

```
SELECT v FROM (VALUES (2),(NULL),(1)) t(v) ORDER BY v ASC;
 1
 2
 NULL
```

So `order(:position, :label)` puts pinned terms first and unpinned (NULL) ones after, which is what the spec asserts. That spec passes, which is the other half of the check: if NULLs sorted first, `orders pinned terms first, then unpinned terms by label` would fail.

This is also why the previous revision's `Arel.sql('position ASC NULLS LAST')` was redundant and got dropped.

ShanaLMoore:
Two follow-ups from the Copilot review, plus a backwards-compatibility pass. Pushed in 8679a3dbc.

**The suppressed `presence` comment was a real bug, now fixed.** `name_changed?` is false on a new record left at nil, so guarding presence on it let a nameless vocabulary save:

```ruby
Qa::LocalAuthority.new(name: nil).valid?  # => true
Qa::LocalAuthority.new(name: nil).save    # => true  (NULL name persisted)
```

`presence` is now unguarded; only the format, uniqueness and collision rules stay guarded on `name_changed?` so pre-existing rows remain editable. Added a spec for the nil case, which the empty-string spec was missing.

**The NULLS-ordering comment is inverted**, answered inline: Postgres ASC sorts NULLs last, so pinned terms do come first.

**Backwards compatibility.** Nothing here requires an app to move from file-based to DB-backed vocabularies, and I verified that rather than assuming it:

- File-based authorities still resolve with zero rows in the qa tables: `Qa::Authorities::Local.subauthority_for('licenses')` returns `FileBasedAuthority`, 17 terms, search and find both working, `Qa::LocalAuthority.count == 0` throughout. The YAML path never touches these tables.
- `TableBasedAuthority#all`, `#search` and `#find` return exactly what they did before, including inactive terms. Nothing filters on `active` yet; wiring that up is #699's job, so no existing suggestion list changes behavior in this PR.
- `Qa::Authorities::Mesh#search` still returns its expected shape. MeSH is the only existing consumer of these tables in Hyku, and it creates `name: 'mesh'`, which passes the new format rule.
- Both migrations are purely additive: nullable columns, or NOT NULL with a constant default, plus two new indexes. No column or index is dropped or narrowed, and `change` reverses cleanly (rolled back and re-ran across all tenants).

One change prompted by this question: the name format now allows hyphens as well as underscores. A downstream knapsack could reasonably already have `local/lab-names`, and rejecting that on rename would have been a gratuitous break.

db/migrate/20260817170001_add_active_data_and_position_to_qa_local_authority_entries.rb
laritakr: I had the wrong terminology... I think this is what I actually meant! lol

laritakr (approved):

PR #3224, opened 2026-08-18
Seed a label and description for every local vocabulary

ShanaLMoore:
Loading vocabularies into the database left `label` and `description` null, so the admin UI in #3225 has nothing to show a staff member but a lowercase key like `oer_types`. Reseeding never fixes it because `populate_qa` only creates rows it doesn't already find.

Both values are now derived at seed time: the name titleized, and the term count with up to three example terms.

- A vocabulary can override either by carrying `label:` or `description:` alongside `terms:` in its yml. Qa reads only the `terms` key, so the extra keys are inert to it and a knapsack gets both by editing the yml it already ships. `oer_types` is the only one here that needs it, because titleize renders it "Oer Types".
- Derived text is stored rather than computed, so a description goes stale as terms change. It reads as a starting point for staff to replace, which is why it states what the vocabulary holds rather than what it means.
- The migration skips any row an admin has already edited, as does reseeding.

Testing: `bundle exec rspec spec/services/local_vocabulary_service_spec.rb spec/models/qa spec/authorities/qa` - 42 examples, 0 failures. Migration run across all local tenants, and the derived path checked against real knapsack ymls from enact, adventist, wvu and palni_palci, none of which carry either key.

### DIFF
--- a/app/services/local_vocabulary_service.rb
+++ b/app/services/local_vocabulary_service.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class LocalVocabularyService
+  SAMPLE_TERM_COUNT = 3
+
   def self.seed!(paths = Qa::Authorities::Local.subauthorities_path)
     seed_files(paths).map { |file| seed_vocabulary!(file) }
   end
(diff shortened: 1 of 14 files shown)


ShanaLMoore:
Superseded by #3226 - GitHub won't reopen this one after the branch was force-pushed.

PR #3225, opened 2026-08-18
Browse controlled vocabularies from the dashboard

laritakr:
## Summary
Adds a dashboard page listing every controlled vocabulary available to a tenant, and a read-only page showing the terms of each local one.

Refs 
- notch8/palni_palci_knapsack#698
- notch8/palni_palci_knapsack#699

## Details
- The index lists locally-managed, file-based, and external-service vocabularies together, with the source key to paste into a metadata profile.
- External services come from the questioning_authority gem rather than a hardcoded list.
- The show page lists a local vocabulary's terms with the term ID that gets stored on a record and supplied in a Bulkrax import.
- Vocabularies holding an imported copy of an external one, such as MeSH, are listed but not opened — a MeSH release runs to roughly 30,000 terms. Each authority declares whether it owns its terms, so this is not a hardcoded exception.
- Depositors can view; only admins will be able to edit. A new ability module separates the two, gated on the same check Bulkrax uses for importing.
- Neither ticket is closed by this. Term editing, activate/deactivate toggles, copy-to-clipboard for import values, pagination, and sorting remain open on 699; usage locations within the metadata profile remain open on 698.

## Testing
- Visit Dashboard → Tasks → Controlled Vocabularies as an admin. Confirm the list includes vocabularies from all three sources and that only locally-managed ones are links.
- Open a local vocabulary and confirm its terms show a label, a term ID, and a status.
- Confirm MeSH appears with a "Not imported" badge and is not a link.
- Sign in as a user who can deposit but is not an admin. Confirm the menu item appears under a "Tasks" heading and the page loads.
- Sign in as a user who cannot deposit. Confirm the menu item is absent and the URL is refused.
- Set a GeoNames username in account settings, then confirm GeoNames reads "Ready" rather than "Needs setup".

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -780,3 +780,32 @@ dd {
     font-size: 0.875rem;
     color: #6c757d;
 }
+
+// Status badges for the controlled vocabulary dashboard.
+//
+// Set explicitly rather than reusing badge-warning and friends: the themes
+// flatten most Bootstrap badge colours to grays
+// Needs no action, so it recedes.
+.badge-quiet {
+    background-color: #e9ecef;
+    color: #495057;
+}
+
+// Needs attention.
+.badge-attention {
+    background-color: #b45309;
+    color: #fff;
+}
+
+// The question mark beside a column heading that carries a tooltip
+.column-hint {
+    border: 0;
+    cursor: help;
+    font-size: 0.875em;
+    vertical-align: baseline;
+
+    &:hover,
+    &:focus {
+        text-decoration: none;
+    }
+}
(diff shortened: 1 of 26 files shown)


ShanaLMoore (approved):
Nice work, especially deriving the remote list from qa itself and the `locally_owned?` approach instead of special-casing MeSH. A few non-blocking notes (skipping what Copilot already raised):

- `remote_configured?`: the rescue returns nil, and `Entry#configured?` reads nil as configured, so a check that raises shows "Ready" instead of "Needs setup". Returning false there (or a distinct unknown state) seems safer.
- A tenant could create a database vocabulary named `geonames` or `fast` and collide with a remote source key. A name validation would close that off.
- Profiles already citing `local/foo` now silently degrade to free text, since the fallback only checks the bare name. Probably fine if nothing shipped with that form, but worth a release note if anything did.
- `deposit_wizard/presenter.rb` has its own registered-then-tolerant fallback that behaves slightly differently (no `exists?` guard). It could call `controlled_vocabulary_service_for` now.
- `Qa::AuthorityRegistry.remote_services` walks constants and parses yaml on every index render; the result is boot-stable, so memoizing would cut most of the page cost. Similarly `find!` computes `database` twice via `all`.
- When `file_terms` rescues, the page says "no terms yet", which reads wrong when the terms just couldn't be listed.

None of these block merging. Approving.

PR #3226, opened 2026-08-18
Seed a label and description for every local vocabulary

ShanaLMoore:
Loading vocabularies into the database left `label` and `description` null, so the admin pages in #3225 have nothing to show staff. Replaces #3224.

- Descriptions are written in the ymls rather than generated from the terms, which would repeat the term count and term list the same page renders and go stale on the first edit. Labels come from `display_label`, so only `oer_types` needs a `label:` key - titleize gives "Oer Types".
- Label and description merge across authority paths, though terms still come from the winning file alone. HykuKnapsack's directory comes first, so a knapsack overriding a Hyku vocabulary would otherwise carry Hyku's description off with it.
- `rake populate_qa` fills both columns as it creates a row, and on an existing row with either one blank. Blanks only, so an admin's edit survives a reseed.
- A vocabulary is named after its file, and `Qa::LocalAuthority` rejects a name outside `NAME_FORMAT`, so the task checks every filename before it touches a tenant. One `ETD_Departments.yml` in a knapsack would otherwise raise part way through `Account.find_each` and half-seed the deploy.

## What a knapsack gets

| Vocabulary yml | `label:` | `description:` | Seeded label | Seeded description |
| --- | --- | --- | --- | --- |
| Hyku `resource_types` | - | yes | `Resource Types` | Hyku's |
| Hyku `oer_types` | yes | yes | `OER Types` | Hyku's |
| Knapsack overrides `resource_types`, own terms | - | - | `Resource Types` | Hyku's, merged |
| Knapsack overrides `licenses` | - | yes | `Licenses` | the knapsack's |
| Knapsack-only `reading_rooms` | - | - | `Reading Rooms` | `nil` |
| Knapsack-only `etd_departments` | - | - | `Etd Departments` ⚠️ | `nil` |
| Knapsack-only `donor_names` | yes | yes | `Donors` | the knapsack's |
| Row an admin edited | either | either | untouched | untouched |

A label is always derived, so `label:` is needed only where titleize mangles an acronym. A description never is, and the views omit the line rather than show an empty field.

Testing: `bundle exec rspec spec/services/local_vocabulary_service_spec.rb spec/models/qa spec/authorities/qa` - 52 examples, 0 failures. Every matrix row run against a real tenant.


ShanaLMoore:
@laritakr @kirkkwang - testing notes. There's no migration; `rake populate_qa` does all of it.

**If you haven't run it on this branch:** just run it. Seeding fills `label` and `description` as it creates a row, and fills either one that's blank on a row that already exists.

```
bundle exec rake populate_qa
```

**If you already ran it** - from `main`, or from an earlier push of #3224 - check what you have first:

```ruby
Apartment::Tenant.switch('<tenant>') { Qa::LocalAuthority.pluck(:name, :label, :description) }
```

Null columns: re-running `populate_qa` fills them. Columns already holding text: only blanks are filled, so reseeding won't replace it, and you'll need to clear them. Test tenant only - this drops anything typed in by hand:

```ruby
Account.find_each do |a|
  Apartment::Tenant.switch(a.tenant) { Qa::LocalAuthority.update_all(label: nil, description: nil) }
end
```

Either way, expect all ten labeled on #3225's index, `oer_types` reading "OER Types" rather than "Oer Types", and a description line under each.

app/services/local_vocabulary_service.rb
```diff
@@ -26,4 +32,29 @@ def self.seed_vocabulary!(file)
 
     [name, authority.local_authority_entries.count]
   end
+
+  # Only blanks are filled, so a value staff edit in the admin UI survives a reseed.
+  def self.backfill_metadata(authority, metadata = {})
```
laritakr, line 52: Just thinking out loud... this makes sense for seeding, but we'll want to consider how we can reuse portions of this for UI uploads. From the UI, they may want to make some changes.

spec/services/local_vocabulary_service_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    it 'is nil for a vocabulary that carries none, rather than inventing one' do
+      expect(described_class.default_description({})).to be_nil
+    end
+  end
+
+  describe 'the vocabularies hyku ships' do
+    it 'describes every one of them' do
+      ymls = Dir[Rails.root.join('config', 'authorities', '*.yml')].map { |file| YAML.load_file(file) }
```
laritakr: Two things on this assertion.

**The filenames get discarded.** `.map { YAML.load_file(file) }` drops `file`, so when this fails the output is a diff of parsed yml hashes with no indication which file is missing a description.

**`include('description')` only tests key presence.** A file with a bare `description:` and no value passes this spec, but `default_description`'s `.presence` then turns it into `nil` and the column ends up null. So the spec can be green while a vocabulary is undescribed.

Checking `.present?` and keeping the basenames fixes both — a failure then reads `expected [] to be empty, got ["new_vocabulary.yml"]`:

```suggestion
      undescribed = Dir[Rails.root.join('config', 'authorities', '*.yml')]
                    .reject { |file| (YAML.load_file(file) || {})['description'].present? }
                    .map { |file| File.basename(file) }

      expect(undescribed).to be_empty
```

app/services/local_vocabulary_service.rb
```diff
@@ -2,20 +2,26 @@
...
-  def self.seed_vocabulary!(file)
-    name = File.basename(file, '.yml')
-    authority = Qa::LocalAuthority.find_or_create_by!(name:)
+  # @param contents [Array<Hash>] every yml this name resolves to, overriding path first
+  def self.seed_vocabulary!(name, contents)
+    terms = Array.wrap(contents.first['terms'])
+    authority = Qa::LocalAuthority.find_or_initialize_by(name:)
+
+    backfill_metadata(authority, merged_metadata(contents))
+    authority.save! if authority.changed?
```
laritakr: This can raise `ActiveRecord::RecordInvalid` on a filename, and it's hard to diagnose when it does.

`name` comes from `File.basename(file, '.yml')` on line 13, and `Qa::LocalAuthority` validates it against `NAME_FORMAT = /\A[a-z0-9][a-z0-9_-]*\z/` whenever `name_changed?` — which is always true on a newly initialized record. So a knapsack shipping `ETD_Departments.yml` or `donor names.yml` raises here, and the message names the attribute rather than the file, giving an operator no path to the offending yml.

Since `rake populate_qa` is the deploy mechanism, that exception propagates out of `Apartment::Tenant.switch` and aborts `Account.find_each` mid-loop: tenants before the failure are seeded, tenants after aren't.

The validation exposure isn't new — `find_or_create_by!` could raise the same way. What's new is that this line is now reached for *every* authority on *every* reseed, including existing rows, because `backfill_metadata` marks the record dirty.

Re-raising with the filename would make it diagnosable:

```suggestion
    begin
      authority.save! if authority.changed?
    rescue ActiveRecord::RecordInvalid => e
      raise e.class, "#{name}.yml: #{e.message}"
    end
```

Worth considering whether `populate_qa` should collect bad files and report them after the loop instead of dying on the first one — otherwise one malformed knapsack yml leaves the deploy half-seeded.

spec/services/local_vocabulary_service_spec.rb
ShanaLMoore: Applied in 556498e9, thank you - both were real. A bare `description:` on a new yml was passing and seeding null.

app/services/local_vocabulary_service.rb
ShanaLMoore: Both parts done in 556498e9.

On the re-raise, I had to go through the instance rather than the class - `RecordInvalid#initialize` takes a record, so `raise e.class, "..."` passes the string in where the record belongs and dies calling `.errors` on a String. `raise e, "#{file}: #{e.message}"` clones the exception and replaces the message, so the class and the record survive.

On the half-seeded deploy: rather than collect failures per tenant, `populate_qa` now checks every filename against `NAME_FORMAT` before it touches the first tenant and exits if any fail. A bad filename is the same bad filename in every tenant, so collecting after the loop would report it once per tenant and still have written to some of them. Verified locally with an `ETD_Departments.yml` on `AUTHORITIES_PATH`:

```
Rename these ymls before seeding. A vocabulary is named after its file, and the name has to be
lowercase letters, numbers, underscores and hyphens, starting with a letter or number:
  /tmp/vocabtest/ETD_Departments.yml
```

Exits 1, and the good yml sitting beside it seeded in no tenant. `seed_vocabulary!` still raises rather than skipping, since `CreateAccount` calls `seed!` and a new tenant quietly missing a vocabulary seems worse than a failed create.

app/services/local_vocabulary_service.rb
ShanaLMoore: Agreed, and I think the split lands about here. `default_label`, `default_description` and the `NAME_FORMAT` check in `invalid_files` are all upload-shaped already - an upload has to reject a bad name too, and it needs a label when the file carries none.

`backfill_metadata` is the piece an upload should not reuse. Filling blanks only is what keeps a reseed off an admin's edit, but a staff member uploading a replacement yml is expecting their label and description to take, not to be ignored because a value is already there. So an upload wants the same defaulting with a plain assignment, and the merge across authority paths doesn't apply at all - there's one file, not an override chain.

Happy to pull those out into something the upload path can call when we get to it; leaving it as is until we know what the upload form actually collects.

laritakr (approved):

PR #3228, opened 2026-08-19
Show which properties and work types use a vocabulary

### LINKED ISSUE #708: Show which properties and work types use a vocabulary
# User Story
**As a** repository manager,
**I want** the vocabulary page to tell me which metadata properties use the vocabulary and which work types those properties appear on,
**so that** I know what I am about to affect before I deactivate a term or change a label.

# Problem Statement
The vocabulary show page has a "Source" row that says where the vocabulary came from, which the index already shows and which nobody needs twice. What is missing is the answer staff actually want on that page: what does this vocabulary control? Today the only way to find out is to read the tenant's metadata profile by hand, and because properties are not shared across every work type, reading it wrong is easy.

# Acceptance Criteria
- [ ] Remove the "Source" row from the vocabulary show page. The index keeps its Source column, which is what separates editable tenant vocabularies from config-file and external ones.
- [ ] In its place, list every property in the tenant's current metadata profile whose `controlled_values.sources` cites this vocabulary's source key.
- [ ] For each property, show the work types it is available on, read from `available_on.class` in the profile. Properties are not available on the same set of work types, so the list is per property rather than one list for the vocabulary.
- [ ] Use the property's display label where the profile gives one, with the raw property name available so it can be matched to the profile.
- [ ] Empty state when no property cites the vocabulary. That is a real and useful answer: the vocabulary is not in use and a term can be changed freely.
- [ ] Works for a vocabulary cited by more than one property, and for a property citing more than one vocabulary.
- [ ] Reads the tenant's current profile once per request. `Hyrax::FlexibleSchema.order("created_at asc").last` is already being called per form render in `app/helpers/hyrax/form_helper_behavior.rb`; do not add another uncached call per row.
- [ ] Tenant-scoped: usage reflects the tenant's own profile, not another tenant's.
- [ ] Test coverage for a vocabulary used by several properties, one used by none, and a property available on a subset of work types.

# Screenshots or Video
<details>
<summary>View Mockups/Screenshots</summary>
</details>

# Testing Instructions
1. As a tenant admin, open Dashboard > Controlled Vocabularies and pick a vocabulary the profile uses, for example `licenses`.
2. Confirm the Source row is gone and a usage section lists the citing properties, each with the work types it applies to.
3. Cross-check one property against the tenant's metadata profile: the work types shown should match that property's `available_on.class`.
4. Open a vocabulary no property cites and confirm the empty state explains it is unused rather than showing a blank section.
5. Confirm the index page still shows its Source column.

## Sample Files
- The tenant's current metadata profile, for cross-checking the property and work type lists.

# Notes
*   Part of Epic #695.
*   Changes the show page added in samvera/hyku#3225.
*   Profile shape: `properties.<name>.controlled_values.sources` is the array of source keys, and `properties.<name>.available_on.class` is the list of classes. See `config/metadata_profiles/m3_profile.yaml` for the shipped example, for instance `license`.
*   Usage counters over deposited works are a separate stretch item and are not in scope here. This is profile configuration, not a count of records.

ShanaLMoore:
Staff deciding whether a term is safe to deactivate or relabel had no way to see what a vocabulary controls without reading the tenant's metadata profile by hand. The vocabulary show page now lists every property citing the vocabulary and the work types each property is available on, with an explicit empty state when nothing cites it.

Worth a look:
- Properties display by raw profile name only (per LaRita, so they can be matched to the profile or a Bulkrax import), not display label.
- Works with `HYRAX_FLEXIBLE=false` too: usage comes from the static form mapping plus the partials that read an authority directly (`oer_types`, `geonames`), and file-backed vocabularies note their terms need a developer. Database-backed and remote ones don't get that note, since the qa fallback serves tenant rows and remote terms live in the service.
- The shipped profile's `CollectionResource` label "pcdmcollection" is corrected to "Collection", with a locale override for tenants whose saved profile predates the fix.
- The dashboard's locale block is translated into de/es/fr/it/pt-BR/zh (it was English-only).

Testing: `rspec spec/services/controlled_vocabulary_usage_spec.rb spec/requests/controlled_vocabularies_spec.rb` (31 examples, 0 failures), rubocop clean, eager load OK.

Closes notch8/palni_palci_knapsack#708.

## Screenshots

**No properties using it (empty state):**

vocab-show-unused-empty-state

**One property using it:**

vocab-show-licenses-used-by

**Multiple properties, per-property work types:**

vocab-show-multi-properties

**Flexible metadata off, file-backed vocabulary (usage still shown, terms need a developer):**

vocab-show-no-profile-mode

**Index with its Source column:**

vocab-index-source-column

### DIFF
--- a/app/services/controlled_vocabulary_usage.rb
+++ b/app/services/controlled_vocabulary_usage.rb
@@ -36,7 +36,7 @@ class ControlledVocabularyUsage
       properties.filter_map do |name, config|
         next unless config.is_a?(Hash) && cites?(config, key)
 
-        Property.new(name: name, work_types: work_types(profile, config))
+        Property.new(name: name, work_types: work_types(config))
       end
     end
 
@@ -76,15 +76,8 @@ class ControlledVocabularyUsage
       model_classes.filter_map do |klass|
         next unless klass.fields.include?(name.to_sym)
 
-        WorkType.new(name: klass.name, label: model_label(klass))
+        WorkType.new(name: klass.name, label: class_label(klass.name))
       end
-    end
-
-    # The Valkyrie suffix is an implementation detail; staff know these as work
-    # types, matching what a profile-backed tenant sees.
-    def model_label(klass)
-      I18n.t("hyku.admin.controlled_vocabulary.work_types.#{klass.name.underscore}",
-             default: klass.name.demodulize.delete_suffix('Resource').titleize)
     end
 
     # Valkyrie classes only: the registry also lists each type's ActiveFedora
@@ -103,22 +96,18 @@ class ControlledVocabularyUsage
       end
     end
 
-    def work_types(profile, config)
+    def work_types(config)
       Array(config.dig('available_on', 'class')).map do |class_name|
-        WorkType.new(name: class_name, label: work_type_label(profile, class_name))
+        WorkType.new(name: class_name, label: class_label(class_name))
       end
     end
 
-    # Labels the shipped profile got wrong; profiles saved before the fix still
-    # carry them, so they fall through to the locale instead of the profile.
-    MISLABELED_CLASSES = ['pcdmcollection'].freeze
-
-    def work_type_label(profile, class_name)
-      label = profile.dig('classes', class_name, 'display_label').presence
-      return label if label && MISLABELED_CLASSES.exclude?(label)
-
+    # From the class name, not the profile's display_label: profiles have shipped
+    # with labels like "pcdmcollection", and the class name is stable across
+    # tenants. The Valkyrie suffix is an implementation detail staff never see.
+    def class_label(class_name)
       I18n.t("hyku.admin.controlled_vocabulary.work_types.#{class_name.underscore}",
-             default: label || class_name)
+             default: class_name.demodulize.delete_suffix('Resource').titleize)
     end
   end
 end
--- a/app/services/controlled_vocabulary_usage.rb
+++ b/app/services/controlled_vocabulary_usage.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+# What a controlled vocabulary controls: the properties that cite it and the work
+# types each property is available on, read from the tenant's current metadata
+# profile, or from the static form mapping when flexible metadata is off.
+# Configuration only, not a count over deposited works.
+class ControlledVocabularyUsage
+  Property = Struct.new(:name, :work_types, keyword_init: true)
+  WorkType = Struct.new(:name, :label, keyword_init: true)
+
+  class << self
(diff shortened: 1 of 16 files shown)


app/services/controlled_vocabulary_usage.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+    end
+
+    def work_types(profile, config)
+      Array(config.dig('available_on', 'class')).map do |class_name|
+        WorkType.new(name: class_name, label: work_type_label(profile, class_name))
+      end
+    end
+
+    # Labels the shipped profile got wrong; profiles saved before the fix still
```
laritakr: Is this exception necessary - Can we just use the class's name for everything and just ignore what the profile defines for the labels?

Calling class_name.safe_constantize&.name on the models stored in the profile will return the actual name regardless of what label is stored there.

laritakr (dismissed):
Just one non-blocking comment/question

app/services/controlled_vocabulary_usage.rb
ShanaLMoore: Done in ce5b9dd1: work type labels now come from the class name (locale override for acronyms like OER/ETD), profile display_labels are ignored, and the mislabeled-classes exception is gone.

app/services/controlled_vocabulary_usage.rb
```diff
@@ -0,0 +1,124 @@
...
+    # carry them, so they fall through to the locale instead of the profile.
+    MISLABELED_CLASSES = ['pcdmcollection'].freeze
+
+    def work_type_label(profile, class_name)
+      label = profile.dig('classes', class_name, 'display_label').presence
+      return label if label && MISLABELED_CLASSES.exclude?(label)
```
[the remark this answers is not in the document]
ShanaLMoore: Removed entirely in ce5b9dd1; labels no longer consult profile display_labels, so the constant is gone.

config/locales/en.yml
```diff
@@ -338,6 +339,15 @@ en:
         terms: Terms
         unconfigured: Needs setup
         unconfigured_help: This service needs credentials before it can be used. Until then a field citing it accepts free text instead of offering suggestions.
+        unused: No metadata property uses this vocabulary.
+        unused_profile_hint: To start using it, add this vocabulary's source key to a property's controlled values in the tenant's metadata profile.
+        used_by: Used by properties
+        used_by_hint: The metadata properties controlled by this vocabulary, as named in Bulkrax imports and metadata profiles.
+        used_by_work_types: 'Available on:'
+        work_types:
```
laritakr, line 347: The collection key is now unused.

Non-blocking comment... we have other i18n cleanup so it can also be handled lately.

laritakr (approved):

PR #701, opened 2017-01-26
Updating CONTRIBUTING.md as per Hydra v11.0.0

jeremyf:

PR #3230, opened 2026-08-20
Download a vocabulary's terms as CSV or YAML

### LINKED ISSUE #709: CSV/yaml download for round-tripping a vocabulary's terms
# User Story
**As a** metadata librarian,
**I want** to download a vocabulary's terms as a CSV or yaml, edit it, and upload it back,
**so that** I can make a batch of changes in a spreadsheet instead of clicking through terms one at a time.

# Problem Statement
CSV import lets terms in, but there is no way to get the current list out. Anyone wanting to bulk-edit an existing vocabulary has to retype it, and anyone auditing one has to read it off the screen. Without an export that matches the import format, "bulk edit" is really "bulk replace, from memory".

# Acceptance Criteria
- [x] Download control on the vocabulary show page producing a CSV or yaml of that vocabulary's terms, next to the add term button.
- [x] Vocabulary index page has a download on each row (dropdown with yml and csv options).
- [x] Available when HYRAX_FLEXIBLE is true and false.
- [x] CSV columns are exactly the columns #701's importer will accept (id, label, active), so a downloaded file works as an import template.
- [x] The yaml download uses the qa authority file format (terms: id/term/active) with the source key, label, and description as header keys, directly usable as a config-file vocabulary.
- [x] Inactive terms are exported with `active=false` rather than omitted.
- [x] Terms export in the order the terms table shows them, with no truncation cap.
- [x] Filename is the source key, for example `licenses.csv`.
- [x] Available for any vocabulary whose terms can be listed, including config-file vocabularies.
- [x] Only users who can view the vocabulary can download it; tenant-scoped like the rest of the page.
- [x] Large vocabularies stream rather than loading every term into memory at once.

# Screenshots or Video
<details>
<summary>View Mockups/Screenshots</summary>
</details>

# Testing Instructions
1. As a tenant admin, open a vocabulary with a mix of active and inactive terms and download the CSV.
2. Confirm the columns are id, label, active and that inactive terms are present with `active=false`.
3. Download the yaml and confirm it parses as a qa authority file (terms: id/term/active).
4. Download from a config-file vocabulary and confirm the file is usable as an import template.
5. Repeat a download with HYRAX_FLEXIBLE=false and confirm it works.

## Sample Files
- CSV downloaded from a shipped vocabulary, for use as the round-trip fixture.
- The same file with one label edited and one row set inactive.

# Notes
*   Part of Epic #695.
*   Round-trip behavior (re-upload no-op, label edits matching on id, deleted rows never deleting terms) moved to #701, which owns the importer. This ticket is export only and does not depend on #701.
*   Stretch scope alongside #701 under SOW 4.4: pick up after primary scope.

ShanaLMoore:
Staff bulk-editing or auditing a vocabulary had no way to get the current term list out; the coming CSV import (#701 in the PALS tracker) needs a matching export or "bulk edit" means retyping from memory. The vocabulary show page and each index row now download terms as CSV (columns id, label, active, the import template) or as yaml in the qa authority file format, usable directly as a config-file vocabulary.

Worth a look:
- Works with `HYRAX_FLEXIBLE` true or false; downloads gate on the entry having terms to list (tenant row or config file), so remote authorities 404.
- Inactive terms export with `active=false`, rows follow the terms-table order, filenames are the source key (`licenses.csv`), so a yml download drops straight into `config/authorities/`.
- The yaml opens with the source key, label, and description above `terms:`; qa ignores the extra keys, and the bulk importer can use them to create or update a whole vocabulary from one file. Position is deliberately not exported: row order carries it in both formats.
- The response streams, but term rows are read eagerly inside the request: a query left inside the streamed body runs after Apartment has switched schemas back and exports another tenant's terms. The request spec caught this live.
- Round-trip acceptance criteria moved from the export ticket to the import ticket, which owns that behavior.

Testing: `rspec spec/services/controlled_vocabulary_export_spec.rb spec/requests/controlled_vocabularies_spec.rb` (41 examples, 0 failures), rubocop clean, downloads verified in the browser in both formats.

Closes notch8/palni_palci_knapsack#709.

## Screenshots

**Show page, download dropdown next to Add term:**

vocab-show-download-dropdown

**Index, per-row download dropdown:**

vocab-index-download-dropdown

# EXAMPLE DOWNLOADS

[accessibility_features.yml](https://github.com/user-attachments/files/31282173/accessibility_features.yml)

[accessibility_features.csv](https://github.com/user-attachments/files/31282176/accessibility_features.csv)

### DIFF
--- a/app/services/controlled_vocabulary_export.rb
+++ b/app/services/controlled_vocabulary_export.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# Streams a vocabulary's terms as CSV (the #701 import template) or as a yaml
+# file in the qa authority format.
+class ControlledVocabularyExport
+  COLUMNS = %w[id label active].freeze
+
+  def initialize(entry)
+    @entry = entry
+  end
+
+  def filename(format)
+    "#{@entry.source_key}.#{format}"
+  end
+
+  def csv
+    stream(CSV.generate_line(COLUMNS)) do |term|
+      CSV.generate_line(term.values_at(*COLUMNS))
+    end
+  end
+
+  def yml
+    stream("#{yml_header}terms:\n") do |term|
+      entry = { 'id' => term['id'], 'term' => term['label'], 'active' => term['active'] }
+      { 'terms' => [entry] }.to_yaml.delete_prefix("---\nterms:\n")
+    end
+  end
+
+  private
+
+  # qa's file reader only consumes terms:, so the label and description ride
+  # along for the #701 importer without breaking the config-file format.
+  def yml_header
+    header = { 'source_key' => @entry.source_key, 'label' => @entry.label }
+    header['description'] = @entry.description if @entry.description.present?
+    header.to_yaml.delete_prefix("---\n")
+  end
+
+  # The rows are read up front: a streamed body is drained after Apartment has
+  # switched schemas back, so a query run inside it reads the wrong tenant.
+  def stream(header)
+    rows = terms
+    Enumerator.new do |lines|
+      lines << header
+      rows.each { |term| lines << yield(term) }
+    end
+  end
+
+  def terms
+    return database_terms if @entry.vocabulary
+
+    Qa::Authorities::Local.subauthority_for(@entry.source_key).all.map do |term|
+      term = term.with_indifferent_access
+      { 'id' => term[:id], 'label' => term[:label], 'active' => term[:active] != false }
+    end
+  rescue StandardError => e
+    Rails.logger.warn("Unable to export terms for #{@entry.source_key}: #{e.message}")
+    raise ActiveRecord::RecordNotFound, "No terms to export for #{@entry.source_key}"
+  end
+
+  def database_terms
+    @entry.vocabulary.local_authority_entries.ordered.pluck(:uri, :label, :active).map do |id, label, active|
+      { 'id' => id, 'label' => label, 'active' => active != false }
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/services/controlled_vocabulary_export.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# Streams a vocabulary's terms as CSV (the #701 import template) or as a yaml
```
kirkkwang: not sure if i like the reference to the ticket here, might end up meaning nothing to someone in the future

kirkkwang (dismissed):

app/services/controlled_vocabulary_export.rb
ShanaLMoore: Agreed, both ticket references reworded to "the bulk import(er)" in 7978af34.

app/services/controlled_vocabulary_export.rb
```diff
@@ -0,0 +1,68 @@
...
+class ControlledVocabularyExport
+  COLUMNS = %w[id label active].freeze
+
+  def initialize(entry)
+    @entry = entry
+  end
+
+  def filename(format)
+    "#{@entry.source_key}.#{format}"
+  end
```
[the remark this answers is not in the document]
ShanaLMoore: Description updated: the naming changed to bare source key per Kirk (so the yml drops straight into config/authorities/), and the body now says so.

kirkkwang (approved):

PR #3232, opened 2026-08-21
Heritage feedback

kirkkwang:
## 🧹 Adjust wording for Heritage theme

4f594209645eda7caeb03dfb1837ac89b2ae8f58

Adjust wording based on community feedback.

## 🐛 Send homepage facet "more" to the catalog

f0a022cdce381f71a04966c9a3d1e4eab8d56aed

The "more" link opened a modal onto the homepage instead of the facet
list. HomepageController inherits CatalogController for the facets, but
upstream #search_facet_path names an action and no controller, so
url_for resolved it against a controller with no facet route.

### DIFF
--- a/app/assets/stylesheets/themes/heritage/_primitives.scss
+++ b/app/assets/stylesheets/themes/heritage/_primitives.scss
@@ -121,7 +121,7 @@
 }
 
 // Blacklight's facet cards, restyled into the theme's continuous ruled column.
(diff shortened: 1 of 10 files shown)


ShanaLMoore (approved):

ShanaLMoore (approved):
Fix is right, matches the search_action_url override above it, and the spec fails without it. Locales all check out.

PR #3233, opened 2026-08-21
Vocabulary dashboard follow-ups: review step, terms, and source key resolution

laritakr:
## Summary
Addresses followup issues from the initial creation of the views.
- Refs notch8/palni_palci_knapsack#710

## Details
- Creating a vocabulary shows the derived source key for confirmation first, since a vocabulary cannot be deleted and its key is permanent once set.
- Names colliding with a remote service or an imported copy are refused; a duplicate now names the source key it collides on.
- New terms are given a sequence number so they land at the end of the list. Existing terms without one are left alone, to be sequenced by hand.
- `populate_qa` now seeds a vocabulary only once: a second run leaves one the tenant already has untouched and imports only the ones it has not seen. It used to re-create a term deleted in the dashboard and backfill any the yaml file had gained since.
- A profile citing `loc/genre_forms` now opens its page and reports its usage, instead of 404ing on the older spelling.
- A yaml vocabulary with no registered service class renders as a dropdown rather than free text.
- Column hint icons explain themselves on hover instead of looking clickable and doing nothing.
- "Back to edit" on the review step is a link rather than a GET form submit, which was putting the CSRF token in the query string.

## Testing
- Create a vocabulary from `/dashboard/controlled_vocabularies`. Confirm the review step shows the derived source key (try an accented name — `Café Terms` should show `cafe_terms`), and that "Back to edit" keeps what you typed.
- Confirm creation, then add a term. It should appear at the end of the term list.
- Try creating a vocabulary named `geonames` or `mesh` — both should be refused, naming the key.
- Open a vocabulary that holds an imported copy (`mesh`, once imported). The "Add term" button should be absent.
- Hover and then tab to a column heading's `?` mark on the index and show pages. The explanation should appear both ways.
- Confirm the terms download (CSV and YAML) still works from both the index and a vocabulary's page.
- Delete a term from a seeded vocabulary, then run `bundle exec rake populate_qa`. The term should stay deleted, and the task should report that vocabulary as already imported.

### DIFF
--- a/app/assets/stylesheets/hyku.scss
+++ b/app/assets/stylesheets/hyku.scss
@@ -797,8 +797,28 @@ dd {
     color: #fff;
 }
 
-// The question mark beside a column heading that carries a tooltip
+// Restores the scroll Bootstrap gives .table-responsive: Hyrax sets `overflow-x:
+// visible` on it app-wide in _file-listing.scss, so without this the table is cut
+// off at the container edge. Scoped here so Hyrax's own tables keep its behavior.
+.controlled-vocabularies-scroll {
+    overflow-x: auto;
+}
+
+// One line per cell, so a narrow container scrolls instead of breaking words
+// mid-letter ("Accessibilit / y"). The description spans the row and wraps.
+.controlled-vocabularies {
+    th,
+    td:not(.controlled-vocabulary-description) {
+        white-space: nowrap;
+    }
+}
+
+// The explanation behind a column heading's question mark. CSS rather than
+// Bootstrap's tooltip plugin, which nothing initializes on these pages.
 .column-hint {
+    position: relative;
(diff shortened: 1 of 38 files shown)


ShanaLMoore:
Was dropping `work_types.collection_resource` from the locales intentional? `model_classes` still includes collection classes, and the `class_label` fallback derives "Collection" from the class name, which is only right in English - de/es/fr etc. lose "Sammlung"/"Colección". Looks like the same reason `etd_resource` and `oer_resource` were kept.

ShanaLMoore (approved):
Nice work - the review step, the collision validation, and the seed-once populate_qa all hold up, and the request specs prove the behavior rather than just the status codes. The locale question I left is non-blocking.

PR #3238, opened 2026-08-24
Performance profiling tooling

maxkadel:
Add tooling for Rails performance profiling, to help developers find inefficient or slow Rails code, and fix it.

This is to support folks who are experiencing slow performance in production, especially with installations with a lot of tenants and data.

The miniprofiler will only show in either development mode (local developer development), or in  environments where `HYKU_MINI_PROFILER_ENABLED` has been set to `true`. Should not be enabled in full production instances, as the profiler itself can slow performance.


kirkkwang (approved):
this is great, i was installing rack mini profiler locally and didn't commit but this will definitely save that extra step for me

PR #3239, opened 2026-08-24
Seed rake task for performance testing

maxkadel:
This PR creates a pair of rake tasks for seeding an environment with a large number of objects, to aid in performance testing (see also PR #3238 which adds tooling for performance)

Also fixes bug where the visibility was not being set successfully in the Sample::ValkyrieService.

@samvera/hyku-code-reviewers

### DIFF
--- a/app/services/sample/valkyrie_service.rb
+++ b/app/services/sample/valkyrie_service.rb
@@ -59,6 +59,17 @@ module Sample
 
     private
 
+    # 80% public / 15% authenticated / 5% private, so access-control checks have real variety to chew on.
+    VISIBILITY_POOL = (
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC] * 80 +
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED] * 15 +
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE] * 5
+    ).freeze
+
+    def random_visibility
+      VISIBILITY_POOL.sample
+    end
+
     def confirm_cleanup # rubocop:disable Metrics/AbcSize
       # Skip confirmation if CONFIRM environment variable is set to 'true'
       return true if ENV['CONFIRM']&.downcase == 'true'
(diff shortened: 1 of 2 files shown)


app/services/sample/valkyrie_service.rb
```diff
@@ -26,7 +26,7 @@ def create_sample_data # rubocop:disable Metrics/AbcSize
 
         index_all_works(collections + images + generic_works)
 
-        print_completion_summary(collections, images, generic_works, total_works)
+        print_completion_summary(collections, images, generic_works, [], total_works)
```
kirkkwang: not a blocker since this looks more like a fix, but if this is the pattern we're keeping then kwargs would help, i had to look at the def to see that this is for oers, and what about etds?  Kind of confusing at first glance

app/services/sample/valkyrie_service.rb
```diff
@@ -59,6 +59,17 @@ def clean_sample_data
 
     private
 
+    # 80% public / 15% authenticated / 5% private, so access-control checks have real variety to chew on.
+    VISIBILITY_POOL = (
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC] * 80 +
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED] * 15 +
+      [Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE] * 5
```
kirkkwang: not a blocker again, i like the defaults but would be nice to have a way to control the numbers through env and fallback to the 80/15/5 split, the problem might be, well does it always add up to 100? maybe we can account for that or raise if it doesn't add up to 100.

app/services/sample/valkyrie_service.rb
```diff
@@ -210,12 +221,14 @@ def build_collection(index, collection_type) # rubocop:disable Metrics/AbcSize
...
         subject: sample_data[:subjects][index % sample_data[:subjects].length],
-        visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC,
         collection_type_gid: collection_type.to_global_id.to_s,
         depositor: user.user_key
       }
 
       collection = Hyrax.persister.save(resource: CollectionResource.new(collection_attrs))
+      # visibility= builds an ACL that needs an explicit acl.save - setting it via the constructor is silently ignored.
+      collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
+      collection.permission_manager.acl.save
```
kirkkwang, line 233: nice catch!

lib/tasks/perf_seed.rake
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+namespace :perf do
+  desc 'Create N dummy tenants with sample Valkyrie works, users, and admin sets, for tenant-count performance testing'
+  task :seed_tenants, [:count, :quantity, :users_per_tenant, :admin_sets_per_tenant] => :environment do |_t, args|
```
kirkkwang: non-blocking, took me a little while to understand what `count` is, maybe `tenant_count` instead?  I think it was because we have `quantity` right next to it

kirkkwang (dismissed):
left some non-blocking comments for your consideration

app/services/sample/valkyrie_service.rb
maxkadel: ~It could just be ratios and calculate percent from that?~ 

It looks like `.sample` doesn't need it to add up to 100, just needs to be the ratio you want.

app/services/sample/valkyrie_service.rb
maxkadel: Yes, this was just to get it to not raise, but making it kwargs would probably make sense - probably a different PR though?

lib/tasks/perf_seed.rake
maxkadel: totally

kirkkwang (approved):

PR #3240, opened 2026-08-24
Reduce ability role queries

kirkkwang:
## 🐛 Preload roles' resources in User#hyrax_groups

a63d49dd97992d296c89e0a6ab5fbc55d4c412ec

map(&:resource) loaded each Hyrax::Group with its own query, which
Bullet reported as an N+1 on every request including each thumbnail
download.

## 🧹 Ask rolify for cached roles in permission path

b874b082da1c87319bff3912e6fa9a3573e73b1d

has_role? queries on every call, so checking all eight default roles
cost eight queries.  has_cached_role? filters the already-loaded
association instead.  Do not drop discard_partially_loaded_roles: roles
are per-tenant while users are not, so a User loaded under another
tenant reports its roles.

## 🧹 Cache per-request reads that repeat every render

4b6259e18cb6eeb74a42b6de06304d32ecbcf5a1

ContentBlock rows were re-queried per call, 100 of them on a 100-doc
catalog page, and Qa authorities once per subauthority.  RequestStore
holds both for the request.  Site.reset! drops all three tenant-scoped
keys when the tenant switches; after_commit drops the ContentBlock one
when a block is written.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -170,15 +170,21 @@ class ApplicationController < ActionController::Base
 
   # Find themes set on Site model, or return default
   def home_page_theme
-    current_account.sites&.first&.home_theme || 'default_home'
+    current_account_site&.home_theme || 'default_home'
   end
 
   def show_page_theme
-    current_account.sites&.first&.show_theme || 'default_show'
+    current_account_site&.show_theme || 'default_show'
   end
 
   def search_results_theme
-    current_account.sites&.first&.search_theme || 'list_view'
+    current_account_site&.search_theme || 'list_view'
+  end
+
+  def current_account_site
+    return @current_account_site if defined?(@current_account_site)
+
+    @current_account_site = current_account.sites&.first
   end
 
   # Add context information to the lograge entries
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -42,10 +42,10 @@ class Ability
 
     @user_groups = default_user_groups
     # TODO: necessary to include #hyrax_group_names?
-    @user_groups |= current_user.hyrax_group_names if current_user.respond_to? :hyrax_group_names
-    @user_groups |= ['registered'] if !current_user.new_record? && current_user.roles.count.positive?
+    @user_groups |= current_user_hyrax_groups(Site.instance).map(&:name) if current_user.respond_to?(:hyrax_groups)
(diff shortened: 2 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -170,15 +170,21 @@ def current_account
...
 
   def search_results_theme
-    current_account.sites&.first&.search_theme || 'list_view'
+    current_account_site&.search_theme || 'list_view'
+  end
+
+  def current_account_site
+    return @current_account_site if defined?(@current_account_site)
+
+    @current_account_site = current_account.sites&.first
```
maxkadel, line 187: How is this setup different from memoization using `||=`? ~or maybe |= would be better since it doesn't evaluate the second expression unless the first one is false/nil.~ (had it backwards)

```suggestion
    @current_account_site ||= current_account.sites&.first
```

app/models/ability.rb
```diff
@@ -89,11 +89,11 @@ def superadmin_permissions
 
   # TODO: move method to GroupAwareRoleChecker, or use the GroupAwareRoleChecker
   def superadmin?
-    current_user.has_role? :superadmin
+    current_user.has_cached_role? :superadmin
```
maxkadel, line 92: Cool find! I did not know about this method!

maxkadel (approved):
Looks good! One non-blocking question.

@orangewolf was very concerned about surprising behavior with using the RequestStore, since it can lead to very strange and hard-to-debug behavior, so it might be nice to have him take a look at it, but I'm not blocking on that. From what I can tell it's thread safe.

edit: Oh! It might also be neat to test the n+1 and show the improvement & prevent regression, but again, not blocking.

app/controllers/application_controller.rb
kirkkwang: @maxkadel i was a little confused by that too but here's the gist.  Using `defined?(@current_account_site)` works better on the admin proprietor page where `current_account.sites&.first` is `nil` but `@current_account_site` will get set nonetheless.   Using `||=` would fall through _because_ `@current_account_site` is `nil` and would call `current_account.sites&.first` again.  The `defined?(@current_account_site)` checks the existence of `@current_account_site`.

maxkadel: Thanks for explaining that, sounds good!

PR #3242, opened 2026-08-24
🎨 Open the vocabulary edit form in a modal

ShanaLMoore:
Nick's feedback on #3241: clicking Edit takes you out of the vocabulary page to edit two fields. The record is already loaded, so the form can open on the page instead.

Stacked on #3241, so review that one first. Base is `vocabulary-metadata-edit`.

- Bootstrap 4 modal, the pattern the codebase already uses for citations and the screening room player. No new javascript file, no Hotwire, nothing added to the stack.
- The button keeps its `href` to the edit page. That is what a reader without javascript gets, and what `update` re-renders when the form comes back invalid, so `edit.html.erb` stays.
- Saving still round-trips: the page reloads to the same vocabulary with its flash. If the reload is what Nick is actually reacting to, a `remote: true` submit is the next step, but it has to handle the label-blanked fallback and the description row appearing or disappearing, so I would rather see whether this is enough first.

Testing: `bundle exec rspec spec/requests/controlled_vocabularies_spec.rb`. 69 examples, 0 failures. Rubocop clean. Opened, saved, and confirmed the flash by hand.

## Screenshots

The form opens over the vocabulary page:
Edit form open in a modal

Saved, back on the same page:
Vocabulary page after saving from the modal

### DIFF
--- a/app/views/hyrax/dashboard/controlled_vocabularies/_edit_modal.html.erb
+++ b/app/views/hyrax/dashboard/controlled_vocabularies/_edit_modal.html.erb
@@ -1,0 +1,36 @@
+<div class="modal fade" id="edit-vocabulary" tabindex="-1" role="dialog"
+     aria-labelledby="edit-vocabulary-title" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <%= form_for vocabulary,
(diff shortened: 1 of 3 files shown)


ShanaLMoore:
Superseded by #3243, which edits the fields in place rather than in a modal.

PR #3244, opened 2026-08-25
Hide menu link for vocabulary sidebar

laritakr:
## Summary

Adds a config variable HYKU_VOCABULARY_SIDEBAR_ENABLED to control the menu option in production environments. Variable defaults to false until we are ready to have it appear in production. 

No code or routes are disabled... this only hides the menu option.


ShanaLMoore (dismissed):

ShanaLMoore:
actually, if you're adding this env var we will need to set it in staging and test-deploy files so we can see it in those envs @laritakr

ShanaLMoore (approved):

PR #3245, opened 2026-08-25
Remove duplicate licenses

laritakr:
## Summary

The licenses yml has two duplicate entries which have been removed. These cannot migrate into our new database, so the inactive entries have been removed since the current behavior shows them as active.

### DIFF
--- a/config/authorities/licenses.yml
+++ b/config/authorities/licenses.yml
@@ -42,12 +42,6 @@ terms:
   - id: http://creativecommons.org/licenses/by-nc-sa/3.0/us/
     term: Attribution-NonCommercial-ShareAlike 3.0 United States
     active: false
-  - id: http://creativecommons.org/publicdomain/mark/1.0/
-    term: Public Domain Mark 1.0
-    active: false
-  - id: http://creativecommons.org/publicdomain/zero/1.0/
-    term: CC0 1.0 Universal
-    active: false
   - id: http://www.europeana.eu/portal/rights/rr-r.html
     term: All rights reserved
     active: false


ShanaLMoore (approved):
Approving. The change is correct and behavior-preserving.

Verified against `main`: `http://creativecommons.org/publicdomain/mark/1.0/` and `http://creativecommons.org/publicdomain/zero/1.0/` each appear twice with identical `term`, once `active: true` and once `active: false`. The removed entries are the `active: false` copies, which is the right pair to drop:

- `Qa::Authorities::Local::FileBasedAuthority#find` returns the first matching term, so label lookups for existing works already resolved to the `active: true` entry. No stored value changes meaning.
- Hyrax's select service filters `all` on `active`, so the edit-form dropdown was already showing one option each.
- The only visible difference is anything that renders `all` unfiltered (the controlled vocabularies dashboard), which no longer lists these two URIs twice.
- `LocalVocabularyService.seed_terms!` numbers `position` from the array index, so terms after the removal shift down two. Harmless, and only affects tenants seeded after this merge.

Two stale references to clean up, either here or as a follow-up:

- `app/services/local_vocabulary_service.rb` - the `seed_terms!` comment says "a shipped file can list one id twice (licenses.yml does, for two Creative Commons URIs)". `find_or_create_by!` is still worth keeping as a guard for knapsack-supplied files, but the parenthetical is no longer true.
- `docs/controlled-vocabularies.md` around line 573 - same claim, "`licenses.yml` does, for two Creative Commons URIs".

No test coverage needed for a data-only yml edit; `spec/authorities/qa/authorities/local_vocabulary_spec.rb` reads terms from the yml rather than asserting a fixed count, so nothing should break. No security or performance implications.

PR #3248, opened 2026-08-27
🎁 Index controlled vocabulary labels

kirkkwang:
Locally created vocabulary terms index their id, so the search, facets, and work page showed the raw id instead of the label. Resolve the id through the term's authority at index time. Values carrying a URI scheme are left as-is, since OAI serializes those same Solr fields.


laritakr:
I think there's a better way, that fits into the to_solr method defined. I'm working on a plan for that, but it will likely make most of this unnecessary.

laritakr:
The need for this is replaced by https://github.com/samvera/hyku/pull/3253

PR #3249, opened 2026-08-27
Delete static public/robots.txt shadowing the dynamic robots view

### LINKED ISSUE #2919: Add Hyrax robots.txt work to Hyku
Hyrax added a dynamic robots.txt in https://github.com/samvera/hyrax/pull/7316. We should get this work into Hyku, and ensure it works with Hyku's path structure.

maxkadel:
We need to delete the auto-created public/robots.txt comment-only file so that we can display the dynamic one from Hyrax.

Confirmed live on a production tenant (`d-scholarship.pitt.edu/robots.txt`) — it serves only the stub, not the dynamic content.

Closes #2919

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,0 @@
-# See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file


laritakr (approved):
looks good to me

Issue #3035, opened 2026-04-30
Add Optional Clover IIIF Viewer Support to Hyku


trmccormick:
Enable the Clover IIIF Viewer as an optional feature in Hyku, allowing it to be toggled on or off per tenant (or globally), and used in place of Universal Viewer or PDF.js. This should require minimal changes and allow for easy customization of displayed metadata.

Summary of Required Changes:

Settings:

Add a clover_viewer_enabled setting to the tenant or global settings (e.g., in app/models/concerns/account_settings.rb).
Presenter Logic:

Update the work show presenter (e.g., app/presenters/hyku/work_show_presenter.rb) to check if Clover is enabled and select the appropriate viewer.
View Partial:

Create a partial for the Clover viewer (e.g., app/views/hyrax/base/_clover_iiif_viewer.html.erb).
Static Assets:

Add Clover static files (clover.html, clover.js, clover.css) to [iiif_viewer](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Customization:

Allow customization of metadata display in clover.js (e.g., by editing the metadata array in the JS config).
Acceptance Criteria:

Clover viewer can be enabled/disabled via settings.
When enabled, Clover is used for IIIF objects instead of UV or PDF.js.
Static assets are served correctly.
Metadata display is customizable.
No impact on existing viewers when Clover is disabled.
References:

Example files: app/models/concerns/account_settings.rb, app/presenters/hyku/work_show_presenter.rb, app/views/hyrax/base/_clover_iiif_viewer.html.erb, public/iiif_viewer/clover.html, public/iiif_viewer/clover.js, public/iiif_viewer/clover.css
Clover Viewer: https://clover-iiif.github.io/

Current Testing: Current testing show that it works easily as a replacement for the UV viewer. 

<img width="1205" height="840" alt="Image" src="https://github.com/user-attachments/assets/554d0d6f-ab10-4b0a-80d6-6c50358f9a38" />

<img width="1238" height="893" alt="Image" src="https://github.com/user-attachments/assets/c0760280-28b7-4a80-923b-161c79eb0bae" />

ShanaLMoore:
cc @kirkkwang could you leave comments here, that you shared in today's office hr.

kirkkwang:
https://github.com/samvera/hyrax/pull/7400
Here's the PR in Hyrax, run the install generator.

PR #3256, opened 2026-08-29
Add per-work viewer selection behind a feature flag

laritakr:
## Summary
Lets each work choose which player displays its media, instead of every work using the same viewer for the whole tenant.
- Refs notch8/palni_palci_knapsack#724, samvera/hyku#3035

## Details
- Adds a `media_viewer` property to Generic Work, Image, ETD, and OER, offering Universal Viewer, Clover IIIF, Ramp, or PDF.js. Leaving it blank keeps today's behavior.
- Off by default behind the `per_work_media_viewer` feature flag, which hides both the form field and the viewer override. In flexible mode the tenant's metadata profile must also declare the property, so adoption is per tenant.
- Clover and Ramp install from npm into gitignored `public/` directories, following the existing Universal Viewer pattern. Clover is pinned to 3.6.0 — later versions changed the web component API or require a newer Node than the base image.
- No show-page templates change. The viewer is chosen by the presenter, so the existing themes pick it up unmodified.
- Ported from the Enact knapsack, where it has been running in production.


app/models/oer_resource.rb
```diff
@@ -11,6 +11,7 @@ class OerResource < Hyrax::Work
     include Hyrax::Schema(:bulkrax_metadata)
     include Hyrax::Schema(:with_pdf_viewer)
     include Hyrax::Schema(:with_video_embed)
+    include Hyrax::Schema(:with_media_viewer)
```
kirkkwang, line 14: a follow up ticket to hyku_knapsack templates would be nice since we're including this now, i'm thinking whenever a Hyku release gets cut

app/presenters/hyku/media_viewer_behavior.rb
```diff
@@ -0,0 +1,43 @@
...
+      return super unless IIIF_VIEWERS.include?(chosen_viewer)
+
+      # An explicit choice bypasses the tenant-level checks in super (image server,
+      # media type, viewable members): the depositor has already made the call.
+      representative_id.present? && representative_presenter.present?
+    end
+
+    def show_pdf_viewer?
+      # Wins over the tenant-wide default_pdf_viewer flag and the older per-work
+      # show_pdf_viewer boolean, both of which super still honors for unset works.
```
kirkkwang, line 27: i'm wondering if we still want this? maybe it's time to think about deprecating this and the default pdf viewer feature now that we can select the viewer, i think maybe we can fallback to pdf js when appropriate but a user can still choose a player and the show page would respect that.  The feature then turns into something like "split_pdf_into_images?" and maybe should live in IIIF Print.

kirkkwang:
I think this looks good, added some thoughts, all non-blocking.  The only thing we need to do is a follow up ticket with each player in each of the show themes.  For example, Clover a little wonky even in the default theme.

kirkkwang (approved):

PR #3260, opened 2026-08-31
Refuse work types the tenant does not offer

### LINKED ISSUE #3259: Work types absent from a tenant's metadata profile are reachable by URL: custom types 500, stock types save a record
## Summary

`curation_concerns_basic_routes` draws `new` and `create` for every entry in `Hyrax.config.registered_curation_concern_types`, and neither the routes nor the controllers consult `Site#available_works` or the tenant's flexible metadata profile. The only place that filtering happens is the work type picker, in `Hyrax::QuickClassificationQueryDecorator#filtered_available_works`, which is private to that decorator.

So a request that arrives by URL rather than through the picker reaches a work type the tenant does not offer, and the outcome depends on how the class gets its attributes:

- A work type whose attributes come from the profile (`acts_as_flexible_resource`) raises `NoMethodError: undefined method 'depositor='` while the form is built, so the depositor gets a **500** rather than a validation message.
- A stock type keeps its attributes regardless of the profile, so `new` renders a working form and `create` **saves the record**. No error, no warning.

The second case is the more serious one: it fails open and leaves a persistent work of a type the tenant does not offer, invisible to the picker and to Admin → Work Types.

## Steps to reproduce

On a multitenant Hyku with `HYRAX_FLEXIBLE=true`:

1. Pick a tenant and note a registered work type its metadata profile does not declare. Uploading a profile at Admin → Metadata Profiles that omits, say, `OerResource` is enough; `Oer` stays in `Hyrax.config.registered_curation_concern_types` and in `Site#available_works`.
2. Sign in as a user who can deposit.
3. Dashboard → Works → Add New Work. The type is correctly **absent** from the list, since the picker filters.
4. Request `/concern/oers/new` directly. The form renders.
5. Fill in a title, pick an admin set, accept the deposit agreement, save.

The work is created. Same steps against a work type that draws its attributes from the profile give a 500 at step 4 instead.

## Expected behaviour

A work type the tenant does not offer is refused at the controller, for every type, with a message. Neither a 500 nor a saved record.

## Why the picker filter is not enough

`filtered_available_works` already computes the correct set (`available_works & profile_work_types`), but it is private to `QuickClassificationQueryDecorator`, so the controllers cannot reuse it, and nothing else on the create path applies it. `Site#available_works` is not consulted by the controllers at all — removing a type from that column changes the picker and nothing else.

## Suggested fix

1. Extract the set `filtered_available_works` computes into something both the picker and the controllers can call — a method on `Site`, or a small service.
2. Guard `new` and `create` in the works controllers against it, rejecting with a flash rather than raising.
3. Reconsider the search-only early return in `filtered_available_works`:

   ```ruby
   return available_works if Site.account&.search_only?
   ```

   On a search-only tenant the profile filter is skipped entirely, so an undeclared type **is** offered in the picker and the 500 is reachable by clicking rather than only by URL. Whatever the guard does, it and the picker should agree on those tenants.

## Version

Reproduced on Hyku `v7.1.0-308-gb9dd441d` with flexible metadata enabled.

Reproduced through a downstream knapsack, but no knapsack code is involved in the stock-type case: the types, the routes, `Site#available_works`, and the picker decorator are all Hyku's. I have not run the reproduction against a clean Hyku checkout, so if any of the above depends on something in our overrides I would want to know.

We are carrying a downstream controller guard as an interim fix and would be glad to port it upstream if the shape above sounds right.

ShanaLMoore:
Fixes #3259.

## TL;DR

- **Problem:** a signed-in depositor can go straight to `/concern/<type>/new` for a work type the tenant does not offer. Stock types render a form and **save a real record**; profile-backed types raise `NoMethodError: undefined method 'depositor='` and return a 500.
- **Cause:** routes exist for every registered curation concern, and the only filtering lives in `QuickClassificationQueryDecorator`, private to the picker. The controllers never check.
- **Fix:** move that filter to `Site#offerable_work_types` and guard `new` and `create` with it, redirecting with a flash instead of raising or saving.

- **Search-only tenants keep their existing bypass.** They have no profile of their own, so their offerable set stays whatever the site enables. That keeps the picker and the guard in agreement, but leaves #3259's third point unaddressed — happy to follow up if you would rather those tenants were filtered too.

## Reproduction

Clean Hyku, `HYRAX_FLEXIBLE=true`, no downstream app. Tenant offers `GenericWork, Image, Etd`; `Oer` is registered but not offered, and the picker correctly omits it.

`/concern/oers/new` renders anyway, and saving creates a real work (`Oer` id `5e18d8d5…`, since deleted):

| before | before | after |
|---|---|---|
| form renders | work created | refused |

Images live on the `work-type-guard-screenshots` branch to keep them out of this one; delete it once reviewed.

## Both flexible modes

With `HYRAX_FLEXIBLE=false`, `offerable_work_types` returns `available_works` untouched and never reaches `Hyrax::FlexibleSchema`, so the guard still enforces the offer list — a type absent from `available_works` is refused either way. `curation_concern_type` is the `Resource` class in both modes, so the `ModelRegistry` lookup is mode-independent.

The suite runs flexible by default, so I ran the request spec with the var set explicitly both ways rather than assume the non-flexible path.

## Testing

```
bundle exec rspec spec/requests/work_type_not_offered_spec.rb                      # 3 examples, 0 failures
HYRAX_FLEXIBLE=false bundle exec rspec spec/requests/work_type_not_offered_spec.rb # 3 examples, 0 failures
bundle exec rspec spec/models/site_spec.rb \
  spec/services/hyrax/quick_classification_query_decorator_spec.rb                 # 37 examples, 0 failures
```

The existing picker spec passing unchanged against the extracted method is the check that the move preserved behaviour. Rubocop clean. `hyku.works.errors.work_type_not_offered` added to `en` and the six translated locales.

### DIFF
--- a/app/controllers/concerns/hyku/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyku/works_controller_behavior.rb
@@ -7,6 +7,7 @@
 #           - use Hyku::WorkShowPresenter rather than Hyrax's presenter
 #           - refuse a parent_id the depositor cannot edit or whose type cannot
 #             contain the work being created
+#           - refuse a work type the tenant does not offer
 module Hyku
   # include this module after including Hyrax::WorksControllerBehavior to override
   # Hyrax::WorksControllerBehavior methods with the ones defined here
@@ -18,6 +19,7 @@ module Hyku
     included do
       # add around action to load theme show page views
       around_action :inject_show_theme_views, except: :delete
+      before_action :ensure_work_type_offered, only: %i[new create]
       before_action :ensure_parent_accepts_child, only: :create
       self.show_presenter = Hyku::WorkShowPresenter
 
@@ -42,6 +44,18 @@ module Hyku
     end
 
     private
+
+    # Routes are drawn for every registered curation concern, so /concern/<type>/new is
+    # reachable on a tenant that does not offer the type.
+    def ensure_work_type_offered
+      return if Site.instance.offerable_work_types.include?(offered_work_type_name)
+
+      redirect_to main_app.root_path, alert: I18n.t('hyku.works.errors.work_type_not_offered')
+    end
+
+    def offered_work_type_name
+      Hyrax::ModelRegistry.rdf_representations_from([self.class.curation_concern_type]).first
+    end
 
     # parent_id reaches Steps::AddToParent straight from params, and that step
     # validates neither the type pairing nor the user's access to the parent.
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -23,6 +23,23 @@ class Site < ApplicationRecord
   belongs_to :account, optional: true
   accepts_nested_attributes_for :account, update_only: true
 
+  # A type the profile omits gets no metadata behaviour, so offering it can only fail.
+  #
+  # @return [Array<String>] work type names this tenant may offer a depositor
+  def offerable_work_types
+    enabled = available_works || []
+    return enabled unless Hyrax.config.flexible?
+    # Search-only tenants have no profile of their own.
+    return enabled if account&.search_only?
+
+    profile = Hyrax::FlexibleSchema.current_version
+    return enabled unless profile
+
+    declared = (profile['classes']&.keys || []).map { |klass| klass.gsub(/Resource$/, '') }
+
+    enabled & declared & Hyrax.config.registered_curation_concern_types
+  end
+
   class << self
     delegate :account, :application_name, :institution_name, :favicon,
              :institution_name_full, :reload, :update, :contact_email,
--- a/spec/requests/work_type_not_offered_spec.rb
+++ b/spec/requests/work_type_not_offered_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Requesting a work type the tenant does not offer', type: :request, singletenant: true, clean: true do
+  let(:admin) { FactoryBot.create(:admin) }
+
+  before do
+    FactoryBot.create(:admin_group)
+    FactoryBot.create(:registered_group)
+    FactoryBot.create(:editors_group)
+    FactoryBot.create(:depositors_group)
+
+    login_as admin
+  end
+
+  context 'when the site does not enable the type' do
+    before { Site.instance.update!(available_works: %w[Image]) }
+
+    it 'refuses new with a message rather than building the form' do
+      get new_hyrax_generic_work_path
+
+      expect(response).to redirect_to(root_path)
+      expect(flash[:alert]).to eq(I18n.t('hyku.works.errors.work_type_not_offered'))
+    end
+
+    it 'refuses create rather than persisting a work' do
+      expect do
+        post hyrax_generic_works_path, params: { generic_work: { title: ['Should not be created'] } }
+      end.not_to change { Hyrax.query_service.count_all_of_model(model: GenericWorkResource) }
+
+      expect(response).to redirect_to(root_path)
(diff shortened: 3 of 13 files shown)


app/controllers/concerns/hyku/works_controller_behavior.rb
```diff
@@ -43,6 +45,20 @@ def json_manifest
...
+    def ensure_work_type_offered
+      return if Site.instance.offerable_work_types.include?(offered_work_type_name)
+
+      redirect_to main_app.root_path, alert: I18n.t('hyku.works.errors.work_type_not_offered')
+    end
+
+    # curation_concern_type is the Resource class (OerResource), available_works holds the
+    # bare name (Oer), and profiles use either spelling.
+    def offered_work_type_name
+      self.class.curation_concern_type.to_s.demodulize.sub(/Resource\z/, '')
```
kirkkwang: i feel like there's a better way to do this, there's like a model registry lookup or something like that

ShanaLMoore: you're 100% right, good catch thanks

ShanaLMoore: Good call, and the idiom was already two methods down in this file. Swapped to `Hyrax::ModelRegistry.rdf_representations_from([self.class.curation_concern_type]).first`, matching what `ensure_parent_accepts_child` does, and dropped the string munging along with the comment explaining it.

Verified it returns the names `Site#available_works` actually holds:

```
GenericWorkResource -> "GenericWork"
ImageResource       -> "Image"
EtdResource         -> "Etd"
OerResource         -> "Oer"
```

app/models/site.rb
```diff
@@ -23,6 +23,23 @@ class Site < ApplicationRecord
...
   accepts_nested_attributes_for :account, update_only: true
 
+  # A type the profile omits gets no metadata behaviour, so offering it can only fail.
+  #
+  # @return [Array<String>] work type names this tenant may offer a depositor
+  def offerable_work_types
+    enabled = available_works || []
+    return enabled unless Hyrax.config.flexible?
+    # Search-only tenants have no profile of their own.
+    return enabled if Site.account&.search_only?
```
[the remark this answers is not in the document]
ShanaLMoore: Fixed: reads `account&.search_only?` now. Worth flagging that this changed an existing spec — the search-only context in `quick_classification_query_decorator_spec.rb` stubbed `Site.account`, which the refactor no longer calls, so it now stubs the association on the site instance. Same for two stubs I had added in `site_spec.rb`, which were inert once the method stopped reading the class method; a nil account is already not search-only, so I dropped them rather than restubbing.

spec/requests/work_type_not_offered_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(response).to redirect_to(root_path)
+      expect(flash[:alert]).to eq(I18n.t('hyku.works.errors.work_type_not_offered'))
+    end
+
+    it 'refuses create rather than persisting a work' do
+      expect do
+        post hyrax_generic_works_path, params: { generic_work: { title: ['Should not be created'] } }
+      end.not_to change { Hyrax.query_service.find_all_of_model(model: GenericWorkResource).count }
+
+      expect(response).to redirect_to(root_path)
```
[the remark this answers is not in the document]
ShanaLMoore: Both applied: the create example now asserts `flash[:alert]` matches the guard message, so a rejection for unrelated reasons no longer satisfies it, and it counts with `count_all_of_model` instead of enumerating.

kirkkwang (approved):

PR #3261, opened 2026-08-31
Re-register Wings models on reload rather than once at boot

ShanaLMoore:
Follow-up to #3260, found while pulling that change into a downstream app.

## TL;DR

- **Problem:** in development, every work type is refused after the first code reload, so no deposit can be started until the server restarts.
- **Cause:** `config/initializers/wings.rb` registered Wings models and set up lazy migration in `after_initialize`. Both attach singleton methods to the resource classes, which live in `app/models` and are reloadable, so once Zeitwerk rebuilt those classes the methods were gone.
- **Fix:** move the class-touching work into `to_prepare` so it re-runs per reload. Adapter registration and Valkyrie config stay in `after_initialize`, since neither depends on reloadable constants.

## What actually breaks

`Hyrax::ValkyrieLazyMigration.migrating` defines `to_rdf_representation`, `migrating_from`, `migrating_to` and includes the module. After a reload the fresh class has none of it, so `Hyrax::ModelRegistry.rdf_representations_from` falls through to `klass.name`:

```
before reload: GenericWorkResource.to_rdf_representation → "GenericWork"
after reload:  GenericWorkResource.to_rdf_representation → "GenericWorkResource"
```

The guard from #3260 then compares `GenericWorkResource` against `available_works`, which holds `GenericWork`, and refuses everything. That guard is just what made it visible — anything relying on those singleton methods sees the same staleness, including `internal_resource` defaults and Wings conversion.

## Testing

Verified across repeated `Rails.application.reloader.reload!` calls, with the app's own resource classes:

```
before: "GenericWork" / "Oer"
after1: "GenericWork" / "Oer"
after2: "GenericWork"
registry: ["GenericWork", "Oer"]
```

On `main` the same probe degrades to `"GenericWorkResource"` after the first reload.

My first attempt at this PR normalized the name inside the guard. @kirkkwang pointed out that was treating the symptom, which was right — that version is gone and this fixes the registration lifetime instead.

### DIFF
--- a/app/controllers/concerns/hyku/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyku/works_controller_behavior.rb
@@ -53,12 +53,8 @@ module Hyku
       redirect_to main_app.root_path, alert: I18n.t('hyku.works.errors.work_type_not_offered')
     end
 
-    # The registry falls back to the class name once a reload detaches
-    # to_rdf_representation, which would otherwise refuse every type in development.
     def offered_work_type_name
-      Hyrax::ModelRegistry
-        .rdf_representations_from([self.class.curation_concern_type])
-        .first.to_s.sub(/Resource\z/, '')
+      Hyrax::ModelRegistry.rdf_representations_from([self.class.curation_concern_type]).first
     end
 
     # parent_id reaches Steps::AddToParent straight from params, and that step
--- a/app/controllers/concerns/hyku/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyku/works_controller_behavior.rb
@@ -53,8 +53,12 @@ module Hyku
       redirect_to main_app.root_path, alert: I18n.t('hyku.works.errors.work_type_not_offered')
     end
 
+    # The registry falls back to the class name once a reload detaches
+    # to_rdf_representation, which would otherwise refuse every type in development.
     def offered_work_type_name
-      Hyrax::ModelRegistry.rdf_representations_from([self.class.curation_concern_type]).first
+      Hyrax::ModelRegistry
+        .rdf_representations_from([self.class.curation_concern_type])
+        .first.to_s.sub(/Resource\z/, '')
     end
 
     # parent_id reaches Steps::AddToParent straight from params, and that step
--- a/config/initializers/wings.rb
+++ b/config/initializers/wings.rb
@@ -28,18 +28,6 @@ Rails.application.config.after_initialize do
     require 'wings'
     require 'freyja'
 
-    WINGS_CONCERNS ||= [AdminSet, Collection, Etd, GenericWork, Image, Oer].freeze
-
-    WINGS_CONCERNS.each do |klass|
-      Wings::ModelRegistry.register("#{klass}Resource".constantize, klass)
-      Wings::ModelRegistry.register(klass, klass)
-    end
-
-    Wings::ModelRegistry.register(FileSet, FileSet)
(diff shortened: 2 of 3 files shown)


app/controllers/concerns/hyku/works_controller_behavior.rb
```diff
@@ -53,8 +53,12 @@ def ensure_work_type_offered
       redirect_to main_app.root_path, alert: I18n.t('hyku.works.errors.work_type_not_offered')
     end
 
+    # The registry falls back to the class name once a reload detaches
+    # to_rdf_representation, which would otherwise refuse every type in development.
     def offered_work_type_name
-      Hyrax::ModelRegistry.rdf_representations_from([self.class.curation_concern_type]).first
+      Hyrax::ModelRegistry
+        .rdf_representations_from([self.class.curation_concern_type])
+        .first.to_s.sub(/Resource\z/, '')
```
kirkkwang: i feel like this is treating a symptom instead of the actual cause, can you investigate here?  I feel like this being in an after_initialize is the actual problem since I've seen old local dev only wonky issues before.  It always seems to point at stuff like that.

https://github.com/samvera/hyku/blob/reference-theme/config/initializers/wings.rb#L167-L204

ShanaLMoore: You were right — investigated and it is the `after_initialize`, exactly as you suspected. Rewrote the PR around that; the normalization in the guard is gone.

What happens: `Hyrax::ValkyrieLazyMigration.migrating` (called from `wings.rb`, lines 53-59 on main) defines `to_rdf_representation`, `migrating_from`, `migrating_to` and includes the module on the resource classes. Those live in `app/models`, so they are reloadable, and `after_initialize` only ever runs against the class objects present at boot. After a reload the fresh class has none of it:

```
before reload: GenericWorkResource.to_rdf_representation → "GenericWork"
after reload:  GenericWorkResource.to_rdf_representation → "GenericWorkResource"
```

The registry then falls back to `klass.name`, which is what the guard was tripping over. Same staleness would reach anything else depending on those singleton methods, not just #3260.

Moved the registration and `migrating` calls into `to_prepare`; adapter registration and the Valkyrie config stay in `after_initialize` since they do not touch reloadable constants. Verified `to_rdf_representation` holds across repeated `reload!` calls now.

kirkkwang (dismissed):

ShanaLMoore:
@kirkkwang one commit landed after your approval, so this needs another look: 77cf8b2 puts `WINGS_CONCERNS` back. Nothing in hyku used it, but it was top-level and a knapsack could be referencing it. It holds names now rather than class objects, since a constant holding reloadable classes would pin a stale generation.

Verified against hykuup_knapsack, with only this branch's `wings.rb` swapped into the submodule checkout:

- development, repeated `reload!`: `to_rdf_representation` stays `"GenericWork"` / `"Oer"`. On d464f060 it degrades to `"GenericWorkResource"` after the first reload.
- `RAILS_ENV=production` with eager loading: Freyja adapter, registry `["GenericWork", "Oer"]`, resolver, `model_name.route_key` and `internal_resource` all unchanged from baseline.
- `HYRAX_SKIP_WINGS=true`: pg adapter, correct names, Wings undefined.

Copilot's two comments were against the abandoned first approach. The `Resource` stripping it flagged is gone, and the request spec is byte-identical to what merged in #3260.

kirkkwang (approved):

PR #3262, opened 2026-09-01
Key the redirects predicate as property_uri

ShanaLMoore:
The `redirects` property added in #3251 declares its RDF predicate as `predicate:`. `Hyrax::FlexibleSchema#values_map` assigns `values['predicate'] = values['property_uri']` unconditionally, so the declared URI is discarded and the property ends up with a nil predicate. `property_uri` is the input key; `predicate` is derived from it.

- `redirects` was the only property in this profile using `predicate`, and the only one of the 80 with no `property_uri`. The other 72 that declare a predicate all use `property_uri`.
- Behaviour change is confined to the property gaining the predicate it was always meant to have. Nothing reads `predicate` from profile YAML, so no existing consumer moves.

Testing: profile parses, and no property is left declaring `predicate` without `property_uri`. No spec asserts a predicate for `redirects`.

### DIFF
--- a/config/metadata_profiles/m3_profile.yaml
+++ b/config/metadata_profiles/m3_profile.yaml
@@ -2151,7 +2151,7 @@ properties:
       display: false
     indexing:
       - editor_only
-    predicate: http://samvera.org/ns/hyku/redirects
+    property_uri: http://samvera.org/ns/hyku/redirects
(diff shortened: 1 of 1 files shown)


laritakr (approved):
good catch

PR #3265, opened 2026-09-02
🐛 Stop the Site spec sharing one RequestStore

kirkkwang:
A describe-level stub returned one hash to every thread, so the thread-isolation example raced against its own harness.  Scope the stub to the contexts that want a fake store.


maxkadel (approved):
Thanks Kirk. I hate when a flaky spec ends up being one I wrote.

kirkkwang:
> Thanks Kirk. I hate when a flaky spec ends up being one I wrote.

Lolol I think I saw my name in there too somewhere

PR #3276, opened 2026-09-02
Deploy on merge

maxkadel:
- Deploy main to the test environment, staging to the staging environment, and demo to the demo environment

This will only run once the build-test-lint workflow succeeds, to make sure the Docker image has been built before trying to deploy.

The main (default), staging, and production branches already have rulesets to protect against bad actors merging & deploying trojan code using this workflow.

@samvera/hyku-code-reviewers

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,6 +1,13 @@
 name: "Deploy"
-run-name: Deploy (${{ github.ref_name }} -> ${{ inputs.environment }}) by @${{ github.actor }}
+run-name: Deploy (${{ github.event.workflow_run.head_branch || github.ref_name }} -> ${{ inputs.environment || (github.event.workflow_run.head_branch == 'main' && 'test') || (github.event.workflow_run.head_branch == 'staging' && 'staging') || (github.event.workflow_run.head_branch == 'production' && 'demo') }}) by @${{ github.actor }}
 on:
+  workflow_run:
+    workflows: ["build-test-lint"]
+    types: [completed]
+    branches:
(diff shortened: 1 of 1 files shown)


ShanaLMoore (approved):
**Two notes, neither blocking:**

**The `build-test-lint.yaml` hunk is now redundant.** #3279 landed the same two lines on `main` earlier today. Still shows as 2 additions here, so worth rebasing to drop it and keep the diff to one file.

**No production gating.** After this, merging to `production` deploys hykudemo.org unattended. The promotion PR already needs 2 approvals and all checks, so that may well be gate enough, but it is a behaviour change worth announcing to the maintainers who are not in this thread. If you would rather it paused, a required reviewer on the `demo` environment does it without any YAML.

-----

**Approving**. Verified the mapping against the ops templates rather than just the case statement:

| branch | environment | hostname |
|---|---|---|
| `main` | `test` | `hykutest.org` |
| `staging` | `staging` | `hykustaging.org` |
| `production` | `demo` | `hykudemo.org` |

All three correct.

The `workflow_run` re-checkout is the part I would have got wrong. `workflow_run` runs against the default branch HEAD rather than the commit that was built, so without re-checking out `head_sha` and recomputing `TAG`, a deploy from `staging` would have deployed `main`'s code under `main`'s image tag. Gating on `conclusion == 'success' && event == 'push'` also means the image is guaranteed to exist before the pull, which a plain `push` trigger does not give you.



I closed #3280, which was a worse version of this. Yours gates on the build, mine raced it.

ShanaLMoore (approved):

PR #3277, opened 2026-09-02
chore: add .ruby-version specifying Ruby 3.3.0 for Stack Car compatib…

trmccormick:
**Title**: Add .ruby-version for Stack Car development compatibility

**Description**:

## Problem
Developers using Stack Car for local Hyku development encounter `rbenv: command not found` errors when their system's default Ruby version differs from the one where Stack Car is installed. Stack Car requires Ruby 3.3.0, but without a `.ruby-version` file, rbenv cannot automatically switch to the correct version.

## Solution
Add `.ruby-version` specifying Ruby 3.3.0 to normalize the development environment. This ensures all developers using Stack Car (the standard local dev tool for HTTPS/IIIF support) automatically get the correct Ruby version, preventing compatibility issues.

Also updated `.gitignore` to allow `.ruby-version`, as the existing comment already recommends this file be checked in to "normalize the environment."

## Changes proposed in this pull request:
* Add `.ruby-version` file specifying Ruby 3.3.0
* Update `.gitignore` to allow `.ruby-version` (add negation pattern `!.ruby-version` to the existing `.ruby-*` rule)

## Testing
Verified locally: With `.ruby-version` in place, `rbenv` automatically switches to 3.3.0 when entering the Hyku directory, making the `sc` command available without errors.


trmccormick:
Here is the description of my original problem that cause me to look into this this morning as summarized by claude. 

Ruby 3.4.5 was the problematic version — that was your global rbenv default that was causing the Stack Car issues.

The problem:

You had Ruby 3.4.5 set globally in rbenv
Stack Car gem (stack_car) was only installed in Ruby 3.3.0
Without a .ruby-version file, rbenv defaulted to 3.4.5
When you tried to run sc command, it failed with "command not found" because that gem doesn't exist in 3.4.5
The fix:

Added .ruby-version specifying 3.3.0
Now rbenv auto-switches to 3.3.0 when entering the directory
The sc command is available since it exists in that version
So in summary: 3.4.5 broke it → 3.3.0 (via .ruby-version) fixed it

maxkadel (changes requested):
I worry that this could have production implications, and make production upgrades more difficult. It also assumes that all developers are using rbenv, which I don't think is the case.

I think a better solution for your issue would be to just `gem install stack_car` in your new default rbenv ruby version. From what I can see stack_car itself does not require ruby 3.3.0.

trmccormick:
> I worry that this could have production implications, and make production upgrades more difficult. It also assumes that all developers are using rbenv, which I don't think is the case.
> 
> I think a better solution for your issue would be to just `gem install stack_car` in your new default rbenv ruby version. From what I can see stack_car itself does not require ruby 3.3.0.

yes i think that most likely is the better option. Thanks Max

PR #3283, opened 2026-09-04
Honor collection and admin set model config

laritakr:
## Summary
Collections and admin sets now follow the class named in `config.collection_model` / `config.admin_set_model`, so an installation that substitutes its own class gets the right search-result display, sample data, and demo resets.
- Refs notch8/utk_knapsack#13
- Depends on samvera/hyrax#7631, which fixes the same defect upstream.

## Details
- **Collections could render as works in search results.** The two catalog partials identified collections by descent from a hardcoded class, so a configured class outside that hierarchy lost its type badge and thumbnail treatment. Both now use `SolrDocument#collection?`.
- **The spec suite could never test a substituted class.** `spec/rails_helper.rb` pinned `collection_model` and `admin_set_model` for the whole suite, overriding the initializer, so specs exercised those two classes whatever the app configured.
- **Sample data and demo tenant resets** named their classes directly rather than resolving from config.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ gem 'good_job', '~> 4.10' # Updated for Rails 7.2 compatibility
 gem 'googleauth', '~> 1.9.0'
 gem 'google-protobuf'
 gem 'grpc'
-gem 'hyrax', github: 'samvera/hyrax', branch: 'main'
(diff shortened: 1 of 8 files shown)


kirkkwang (approved):
non-blocking, it seems like we can drop the commits about changing it to a branch and then changing it back to main

laritakr:
> non-blocking, it seems like we can drop the commits about changing it to a branch and then changing it back to main

I just did a squash merge

PR #3284, opened 2026-09-08
Give option to run against Postgres 18 in test and development

maxkadel:
This PR gives the option to run Postgres 18 in test and development, and fixes a spec that was failing against Postgres 18.

I'm planning on doing a separate PR that will run CI against postgres 18 and solr 9 on a weekly schedule, to catch any regressions, without ballooning CI, once this and the Solr 9 PR are in.

@samvera/hyku-code-reviewers

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -111,7 +111,7 @@ services:
       - VIRTUAL_PORT=5432
       - VIRTUAL_HOST=db.hyku.test
     volumes:
-      - db:/var/lib/postgresql/data
+      - db:/var/lib/postgresql
     networks:
       internal:
 
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -101,7 +101,7 @@ services:
       internal:
 
   db:
-    image: postgres:11.1
+    image: postgres:18
     env_file:
       - .env
     environment:
--- a/docker-compose.single.yml
+++ b/docker-compose.single.yml
@@ -94,7 +94,7 @@ services:
       - "traefik.http.services.fits-${APP_NAME}.loadbalancer.server.port=8080"
 
   db:
-    image: postgres:11.1
+    image: postgres:18
     env_file:
       - .env
     ports:
--- a/docker-compose.single.yml
+++ b/docker-compose.single.yml
@@ -106,7 +106,7 @@ services:
       - VIRTUAL_PORT=5432
       - VIRTUAL_HOST=db.hyku.test
     volumes:
-      - db:/var/lib/postgresql/data
+      - db:/var/lib/postgresql
 
   adminer:
     image: adminer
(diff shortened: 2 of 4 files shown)


orangewolf (dismissed):

spec/models/featured_collection_list_spec.rb
```diff
@@ -11,10 +11,13 @@
   let(:collection2) { create(:collection, user:, title: ["Collection Title 2"]) }
 
   describe 'featured_collections' do
+    let(:featured_collection1) { create(:featured_collection, collection_id: collection1.id) }
```
orangewolf, line 14: I don't think this change is related to the postgresql update and I also don't think it improves readability of the spec. Lets forgo it

docker-compose.production.yml
```diff
@@ -111,7 +111,7 @@ services:
       - VIRTUAL_PORT=5432
       - VIRTUAL_HOST=db.hyku.test
     volumes:
-      - db:/var/lib/postgresql/data
+      - db:/var/lib/postgresql
```
orangewolf, line 114: This mount change makes sense to me based on current Postgres docker image docs, but it makes this change a major version change with a required migration plan for users of the docker-compose.production.yml file.

docker-compose.single.yml
```diff
@@ -106,7 +106,7 @@ services:
       - VIRTUAL_PORT=5432
       - VIRTUAL_HOST=db.hyku.test
     volumes:
-      - db:/var/lib/postgresql/data
+      - db:/var/lib/postgresql
```
orangewolf, line 109: I believe anyone taking this or the below Postgres change will have to clear and recreate their volume and likely totally wipe their local Hyku in the process.

orangewolf (changes requested):
as this stands its a major version change. Is there any way around that? If not, are we willing to do a full major number release process for this change or should we hold it from main until we are? 

Also there is a spec change that doesn't seem to belong here.

spec/models/featured_collection_list_spec.rb
maxkadel: This refactor was needed to get the test on line 57 passing - that test was previously only passing by chance, we have to explicitly give an order for the featured collections in order for it to pass with Postgres 18.

docker-compose.production.yml
maxkadel: So should we just leave the production one at 11, and bump the others to 18? Or would we consider any change to any of the docker-compose file volumes a breaking change?

I don't know how to do deprecation warnings for things like docker files - do we just put it in the release notes when we're ready?

maxkadel (commented):
Should we consider a change to the volumes in any of the docker-compose files a breaking change? Or only to the `.production` version?

orangewolf (approved):
This is a great refactor. Thanks for breaking this up for us

PR #3288, opened 2026-09-09
Solr 9 compatibility

maxkadel:
Request for reviewers: I have done my best to change the default to Solr 9 without making this a breaking change for existing applications, but I'd love to be double checked on that. 

If that would be a breaking change, I'd rather keep the Solr 8 default with a Solr 9 override option, and change the default for the next major version.

Solr 8 is (way) End of Life. We should make Hyku Solr 9 compatible, and since some applications are still on Solr 8, still test against Solr 8 until we remove Solr 8 support in a major version change.

Parallel changes already in Hyrax's Dassie and Koppie test applications.

* Use Solr 9 compatible fields
* Test against both Solr 8 and Solr 9 - when I was working on this, I found that I introduced some breaking changes against Solr 8, so I think it's worth testing against both until we explicitly drop support for Solr 8.

@samvera/hyku-code-reviewers


Gemfile.lock
```diff
@@ -294,15 +294,15 @@ GEM
       rubocop-performance
       rubocop-rails
       rubocop-rspec
-    blacklight (7.41.0)
+    blacklight (7.42.0)
```
orangewolf, line 297: this is in another pr and shouldn't be here

docker-compose.yml
```diff
@@ -74,12 +74,14 @@ services:
       timeout: "8s"
 
   solr:
-    image: ghcr.io/samvera/hyku/solr:${TAG:-latest}
+    # -9 suffix: pulls the Solr 9 variant (built by build-solr9) independent of web/worker's shared ${TAG}.
+    image: ghcr.io/samvera/hyku/solr:${TAG:-latest}-9
```
orangewolf, line 78: solr 9 being the default makes this whole thing a breaking change. I think having solr 8 be the default still but allow testing and local running with solr 9 is the right thing for this PR. then we can have a seperate PR that switches the default for the major version number bump.

orangewolf (changes requested):
I think we should isolate Solr 9 compatibility from Solr 9 as the default. The first is patch level change... the second is major version bump.

We need instructions for the release on how to upgrade the schema and solrconfig w/o having to reindex and ideally how to upgrade from Solr 8 to Solr 9 for local development w/o loosing your index. We also need release notes to tell people where to go for full production Solr upgrade instructions.

I also wonder if the docker-compose.production.yml needs to be updated for Solr 9 compatibility.

maxkadel:
@orangewolf - I've switched this branch back to defaulting to Solr 8, with Solr 9 as an option.

orangewolf (dismissed):

orangewolf (approved):

PR #3290, opened 2026-09-10
Add rake task to clean up redundant Hyrax::FileMetadata Solr documents

maxkadel:
Companion to samvera/hyrax#7645, which adds `Hyrax.config.index_file_metadata_as_document?` to stop indexing `Hyrax::FileMetadata` as its own redundant Solr document (app-level, no Hyku changes needed to use it).

This PR is just `rake hyku:solr:cleanup_file_metadata` - deletes existing over-indexed FileMetadata documents across tenants after that flag is turned off, since disabling the flag only stops new documents from being created. Dry-run by default (`PERFORM=true` to actually delete, `CNAME=<cname>` to limit to one tenant).

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -54,6 +54,7 @@ module AccountSettings
     setting :s3_bucket, type: 'string', private: true
     setting :shared_login, type: 'boolean', disabled: true
     setting :signup_spam_protection, type: 'boolean', default: false
+    setting :skip_file_metadata_solr_indexing, type: 'boolean', default: false, disabled: true
     setting :smtp_settings, type: 'hash', private: true, default: {}
     setting :solr_collection_options, type: 'hash', default: solr_collection_options
     setting :solr_max_results, type: 'string', default: '10000'
--- a/app/services/hyrax/listeners/metadata_index_listener_decorator.rb
+++ b/app/services/hyrax/listeners/metadata_index_listener_decorator.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax to honor the skip_file_metadata_solr_indexing account setting - FileSetIndexer already has these fields.
+module Hyrax
+  module Listeners
+    module MetadataIndexListenerDecorator
+      def on_file_metadata_updated(event)
+        return if current_account&.skip_file_metadata_solr_indexing
+        super
+      end
+
+      def on_file_metadata_deleted(event)
+        return if current_account&.skip_file_metadata_solr_indexing
+        super
+      end
+
+      private
+
+      def current_account
+        Account.find_by(tenant: Apartment::Tenant.current)
+      end
+    end
+  end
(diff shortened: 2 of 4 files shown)


maxkadel:
Worth noting: a standard reindex already gets you to the same place as this PR, without needing this change at all — `ReindexFileSetsJob`/`ReindexWorksJob`/etc. only ever call `Hyrax.query_service.find_all_of_model` for `FileSet`/work/collection/admin-set classes and `Hyrax.index_adapter.save` on those results. None of them touch `Hyrax::FileMetadata` as its own resource, so a reindex never recreates the standalone FileMetadata documents in the first place — they're only ever created via the upload/characterization event path (`MetadataIndexListener#on_file_metadata_updated`).

Confirmed this empirically on a real tenant migration: after a full reindex (not a re-ingest — same files, same DB, no new uploads), the destination Solr collection landed at the "clean" document count with zero `Hyrax::FileMetadata` documents, while the untouched source tenant still had ~55,000 of them. So this PR isn't changing reindex behavior — it's making the *live* upload/characterization path match what reindexing already does.

maxkadel:
Per-tenant rather than a bare flag so one institution can be tried/verified before flipping every tenant sharing the install - but it's not instead of app-level: `HYKU_SKIP_FILE_METADATA_SOLR_INDEXING=true` already applies to every tenant via `setting`'s existing `Per Tenant > ENV > default` fallback, no per-tenant config needed. Happy to drop the per-tenant column for a bare ENV check if staged rollout isn't actually needed here.

orangewolf (changes requested):
I think this code is in the wrong place, unless we know of instances in Hyrax that make use of this indexing, I feel like it's probably indexed in error o more accurately part of a plan that was never realized. I love the feature flag approach here. I would suggest this check a Hyrax config and be moved to Hyrax. that the FileMetadata indexing itself be deprecated there and the flag be considered temporary until the next major release.

I don't think we need tenant by tenant for disabling this, which would totally eliminate any need for Hyku to change, but if we do, there is precedent for the current account to change the Hyrax.config.

Lastly I don't see the promised `HYKU_SKIP_FILE_METADATA_SOLR_INDEXING` implemented at all in this code... but the suggestion above would skip needing to implement it since you'd get a Hyrax config one instead. So that's only a real issue if point 1 gets dismissed.

maxkadel:
Moved to samvera/hyrax#7645 as `Hyrax.config.index_file_metadata_as_document?` (defaults `true`) and dropped the per-tenant setting here entirely - an app-level flag needs no Hyku changes. It's genuinely ENV-backed now (`HYRAX_INDEX_FILE_METADATA_AS_DOCUMENT`, covered in that PR's `configuration_spec.rb`). This PR is now just the cross-tenant Solr cleanup rake task, which stays in Hyku since it has to iterate tenants/collections Hyrax doesn't know about.

maxkadel:
Closed in favor of https://github.com/samvera/hyrax/pull/7645

PR #3300, opened 2026-09-16
Prevent page bounce on work show refresh for viewer pages

AKFosterDev:
## Summary

- Prevents the visible page bounce/jump on work show page refresh by setting `history.scrollRestoration = 'manual'` on pages that have a viewer (PDF, IIIF, or video embed)
- The fix is scoped to viewer pages only — applied across all 9 show page themes (default, cultural, heritage, community, scholarly, reference, practice_research, screening_room, image)
- URL-encodes the search query parameter in PDF.js URLs to handle special characters correctly

Without this fix, refreshing a work show page with a viewer caused a visible bounce: the browser's default `scrollRestoration: 'auto'` restored the previous scroll position (often partway down the viewer), then Turbolinks booted and scrolled back to the top.

Addresses notch8/hykuup_knapsack#756

## AI Disclosure

AI was used to assist with this change. The code was reviewed and tested by a human, and the human author takes full responsibility.

## Test plan

- [x] Navigate to a work with a PDF viewer — page loads at the top with header and title visible
- [x] Refresh the page — no bounce or jump, page stays at the top
- [x] Navigate to a work with an IIIF viewer — same behavior
- [x] Navigate to a work without a viewer — page loads normally (scroll restoration unaffected)
- [x] Use browser back/forward — Turbolinks scroll restoration still works on non-viewer pages
- [x] Tested across all 9 themes: default, cultural_show, heritage_show, community_show, scholarly_show, reference_show, practice_research_show, screening_room_show, image_show

🤖 Assisted-by: Opus 4.6


ShanaLMoore (dismissed):
A few things to address:

**Still-image scroll on heritage_show** — `theme_viewer?` is broader than `viewer?` (it fires for any work with a representative, including plain thumbnails). On a still-image work, `.hrt-viewer` matches the `.hrt-viewer-still` wrapper, so the page scrolls past title and metadata to a small thumbnail. Should gate on `viewer?` or narrow `theme_viewer?`.

**Only 2 of 9 themes covered** — cultural_show, community_show, image_show, scholarly_show, reference_show, screening_room_show, and practice_research_show don't have the scroll fix. The bounce persists on those themes.

**Duplication** — The script block is near-identical between the two templates, just different CSS selectors. With 7 more themes needing it, extracting a shared partial that takes a selector local (or a standalone JS reading a data attribute) would cover all themes without 9 copies.

The `scrollRestoration` leak (never reset to `'auto'`) is minor but worth a cleanup handler in `turbolinks:before-cache`.

ShanaLMoore (approved):
Looks good. Updated PR removes the scroll-to-viewer behavior (per Nick's call in Slack), covers all 9 themes, and the PDF.js cleanup is solid with good test coverage.

Two minor notes (not blocking):

- `theme_viewer?` vs `viewer?` inconsistency across templates (heritage_show, reference_show, screening_room_show, practice_research_show use the broader guard). Harmless now since we're just disabling scroll restore, not scrolling to anything, but worth unifying later.

- `scrollRestoration = 'manual'` is never reset to `'auto'`. Low impact since Turbolinks manages its own scroll, but a `turbolinks:before-cache` cleanup would be tidy.

Confirmed working on test per the thread.

PR #3306, opened 2026-09-17
Add Tests for User Authentication Behaviors

K8Sewell:
# Summary
In an effort to close the gap between manual testing and automated testing this test has been created to address the missing tests for confirming the expected behaviors of the authentication flow for users.  There are no code changes associated with this PR.  

* n/a - only adds test for existing behaviors

@samvera/hyku-code-reviewers

Generated with assistance from: Claude Opus 5 & Claude Opus 4.8


spec/features/authentication_flows_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe 'Authentication and Account Flows', type: :feature, clean: true do
+  let(:email) { "auth-#{SecureRandom.hex(6)}@example.com" }
+  let(:password) { 'password123' }
+
+  # Authenticates through the real sign-in form (not `login_as`), so Warden test
+  # mode is not active and cannot re-inject the user on the next request. Devise
+  # is configured with `config.sign_out_via = :get`, so following the Logout link
```
maxkadel, line 12: Should we add an `after` block to ensure the user gets signed out even if this one test fails earlier than that step?

spec/features/authentication_flows_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'cgi'
+
+RSpec.describe 'Authentication and Account Flows', type: :feature, clean: true do
+  let(:email) { "auth-#{SecureRandom.hex(6)}@example.com" }
```
maxkadel, line 7: Does FactoryBot not handle creating a random email & password for us?

you could do something like 

```ruby
# either use a bang here so it's instantiated or just call it in the "it" block before running the rest of the test
let!(:our_user) { FactoryBot.create(:user) }

it '...' do
  our_user # instantiates our_user here so it already exists when we try to log in. Do this if Rubocop doesn't like the bang above
  [...]

  fill_in 'user_email', with: our_user.email
  fill_in 'user_password', with: our_user.password
end
```

maxkadel (approved):
Looks good! I have a couple non-blocking questions, but happy to see it merged either way.

spec/features/authentication_flows_spec.rb
K8Sewell: yeah, that's a good idea so it doesn't pollute the rest of the test run.  Will add.  Thank you for thinking of that!

spec/features/authentication_flows_spec.rb
K8Sewell: I was running into issues with duplicates and this was the quick fix but you're right.  Ideally, randomization would be handled at the factory bot level, and you're correct, it is.  I was trying to make as few edits to existing files as possible but now, with your input and suggestion, I think best course is to implement your change and sort out whatever was causing me to have residual users.  I'll be taking this back to draft and will re-request review once these changes are made.  Appreciate you taking a look!

